PR #6, opened 2018-06-07
GraphQL-Based API

### LINKED ISSUE #3: Add GraphQL

### LINKED ISSUE #9: Add currency to `Order`

ashkan18:
Still need to fix/add specs but this would be GraphQL based version of #1 with `CreateOrder` mutation and find order by id query for now. 
Will add spec and queries but opening PR early for comparison. 

screen shot 2018-06-07 at 6 40 51 pm

One thing to notice is GraphQL Ruby has changed it's DSL somewhat completely, this is their new recommended style which doesn't match any of our other ruby GraphQL apps :)

http://rmosolgo.github.io/blog/2018/03/25/why-a-new-schema-definition-api/

and http://rmosolgo.github.io/blog/2018/04/09/updating-github-to-graphql-1-dot-8-0/

fixes #3 

added `currency_code` to `Order` which fixes #9


app/graphql/types/base_union.rb
```diff
@@ -0,0 +1,2 @@
+class Types::BaseUnion < GraphQL::Schema::Union
+end
```
orta: What are all these for?

ashkan18: these are all _Base_ classes provided by `ruby-graphql` when you initialize the project with it.

app/controllers/api/graphql_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+        current_user: current_user
+      }
+      result = StressSchema.execute(query, variables: variables, context: context, operation_name: operation_name)
+      render json: result
+    end
+
+    private
+
+    # Handle form data, JSON body, or a blank value
+    def ensure_hash(ambiguous_param)
```
sweir27: This seems odd/ideally would be handled by a library-- why do we need this?

ashkan18:
merging this for the spike-ness sake :) opening a new one for `SubmitOrder` in few and will address any code review comments in follow up PRs :)

app/graphql/mutations/create_order.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  argument :line_items, [Types::LineItemAttributes], required: false
+
+  field :order, Types::OrderType, null: true
+  field :errors, [String], null: false
+
+  def resolve(user_id:, partner_id:, currency_code:, line_items: [])
+    validate_user!(user_id)
+    {
+      order: OrderService.create!(user_id: user_id, partner_id: partner_id, currency_code: currency_code, line_items: line_items),
```
sweir27: Is there a way to pass in all of the arguments? (i.e. `args` or something like that?)

app/graphql/mutations/create_order.rb
```diff
@@ -0,0 +1,26 @@
...
+      errors: [],
+    }
+  rescue Errors::ApplicationError => e
```
sweir27: This feels like it should/could be abstracted in a future PR.

app/graphql/mutations/create_order.rb
```diff
@@ -0,0 +1,26 @@
...
+    { order: nil, errors: [e.message] }
+  end
+
+  def validate_user!(user_id)
```
sweir27: this (eventually) can be abstracted too 😄 

I feel like graphql-ruby has changed so much that I don't know the current best-practices for this kind of thing! In convection we added the `permit` middleware, not sure if that makes sense here anymore.

spec/support/graphlient_client.rb
```diff
@@ -0,0 +1,16 @@
+require 'graphlient'
+
+RSpec.shared_context 'GraphQL Client', shared_context: :metadata do
+  let(:jwt_user_id) { 'user-id' }
+  let(:jwt_partner_ids) { ['partner-id'] }
+  let(:auth_headers) { jwt_headers(user_id: jwt_user_id, partner_ids: jwt_partner_ids) }
+  let(:client) do
+    Graphlient::Client.new('https://bearden-test.artsy.net/api/graphql', headers: auth_headers) do |client|
```
sweir27: haha `bearden-test`

app/controllers/api/graphql_controller.rb
ashkan18: this was actually auto-generated file by graphql ruby does look pretty rough :)

app/graphql/mutations/create_order.rb
ashkan18: this was also based on ruby graphql example but we could def pass the hash round and this should work since we are trying to map mutations to service methods so technically there won't' be params that we don't want 👍 will cleanup in #13

app/graphql/mutations/create_order.rb
ashkan18: yes! was planning to look into error handling in graphql ruby, the only tricky part is we still want to return `{order: nil` in the response, i think otherwise some clients would break.

app/graphql/mutations/create_order.rb
ashkan18: hehehe this is actually something nice i liked about this, in this mutation `validate_user` means different than submit order one in #13 since different user's can access this mutation.

spec/support/graphlient_client.rb
ashkan18: hahaha how does this even work!

app/graphql/mutations/create_order.rb
sweir27: Hmmmm. Are we sure that's not a default? (I think in metaphysics it works this way-- since you're requesting the `order` field in the mutation response it should return the field + nil)

app/graphql/types/order_type.rb
```diff
@@ -0,0 +1,13 @@
+class Types::OrderType < Types::BaseObject
+  description 'An Order'
+  graphql_name 'Order'
+
+  field :id, ID, null: false
+  field :code, String, null: false
+  field :user_id, String, null: false
+  field :partner_id, String, null: false
+  field :state, String, null: false
+  field :last_state_change_at, Types::DateTimeType, null: false
```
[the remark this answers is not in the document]
jonallured: Coming back to this via some discussion on https://github.com/artsy/metaphysics/issues/1118. I think a timestamp called `state_updated_at` that tracks mutations on the state column is a better fit than relying on `updated_at` on the order. For instance, what if a partner makes a goof and an admin has to change the `currency_code`? That would advance the `updated_at` timestamp, but as this is not a state change, should not advance the timestamp relative to the state. I would add this back in.

PR #13, opened 2018-06-11
Submit/Approve order

ashkan18:
# Problem
As a Collector, I want to be able to Submit my Pending order.
As a Partner user I want to be able to Approve our Submitted orders.

# Solution
- Added new `SubmitOrderMutation` which gets `credit_card_id` and optional `shipping_info` and updates order to submitted order if possible.
  - As part of validation we make sure current user is the owner of this order.
- Added new `ApproveOrderMutaiton` which gets order `id` and updates user status to `approved` if possible.
  - As part of validation we make sure current user has access to the partner of the order.

# Action Based or Generic Mutations?
Originally I tried one [`UpdateOrder` mutation](https://github.com/ashkan18/stress/commit/2baece92a5eb733d2e2ab7f7a017517f7708d11f) which could be used for different update actions possibly can happen to an `Order` but the logic in the Mutation class started getting complicated. 

The idea im trying to keep is, in the Mutation class only do basic validations (like is this person allowed to make this call) and maybe some simple param validation and then just call the service layer method for the actual business logic and more complicated validations (like if any other pending order is around). With that idea, Mutation class for a generic `UpdateOrder` had 2 issues:
- Validation logic would be different based on the new state, when validating user for a submit request we need to make sure its the same user as the order's user while when validating for _approved_ state update we need to make sure current user's partners includes this order's partner (current user has access to order's partner)
- We'd end up not using params validation provided with graphql and possibly confusing the consumers of this mutation:
  - While `shipping_info` is required for a `submit`, we don't need it for `approved` state change in the params. If we want to use same mutation for these two state changes we'd end up making `shipping_info` optional, and it's even possible for someone requesting `approved` state change to pass `shipping_info` which will be totally ignored. With separate mutations we can set 
```
shipping_info, required: true
```
for submit mutation and not have it at all for order update.

# Whats left in submit?
- [ ] We need to call gravity to verify price info and also potentially keep a copy of the artwork at the time of submission in stress.
- [ ] Shipping info? do we store the data in Gravity and pass id here or we store shipping info here?

### DIFF
--- a/app/graphql/mutations/submit_order.rb
+++ b/app/graphql/mutations/submit_order.rb
@@ -1,0 +1,25 @@
+class Mutations::SubmitOrder < Mutations::BaseMutation
+  null true
+
+  argument :id, String, required: true
+  argument :shipping_info, String, required: false
+  argument :credit_card_id, String, required: true
+
+  field :order, Types::OrderType, null: true
+  field :errors, [String], null: false
+
+  def resolve(id:, shipping_info:, credit_card_id:)
+    order = Order.find(id)
+    validate_request!(order)
+    {
+      order: OrderService.submit!(order, user_id: user_id, partner_id: partner_id, currency_code: currency_code, line_items: line_items),
+      errors: [],
+    }
+  rescue Errors::ApplicationError => e
+    { order: nil, errors: [e.message] }
+  end
+
+  def validate_user!(user_id)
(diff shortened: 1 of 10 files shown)


sweir27:
This looks great! I'm guessing we'll need a new gravity endpoint for holding the charge (I think?) so should we merge this PR in the meantime?

ashkan18:
yes, need to look into Gravity's charging side 👍 lets merge this for now

PR #14, opened 2018-06-12
Approve order

ashkan18:
Opening early, for now we should focus on #13


ashkan18:
Closing and adding this as part of #13

PR #21, opened 2018-06-12
Add sidekiq + artwork snapshot

ashkan18:
# Change
We want to store artwork data at the time of order creation on the `LineItem`.

# Solution
Added Sidekiq so we can queue job for fetching Artwork from Gravity when we create new line item.

Added `Adapters::GravityV1` endpoint (pretty much copied from [Impulse](https://github.com/artsy/impulse/blob/master/lib/adapters/gravity_v1.rb)). We most likely will need to make more V1 calls for processing charges and stuff so not using V2 at the moment.

Added `artwork_snapshot` to `LineItem` model, one thing to think of is to decide what/how we need to store on this model, current solution doesn't make much sense, I have `title`, `images`, `permalink` in their own key and `properties` being the filtered response from Gravity.
There was a discussion about this in Impulse a while ago but here we need to decide what we need to store.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,3 @@
+GRAVITY_API_ROOT=https://stagingapi.artsy.net/api
+GRAVITY_JWT=REPLACE
+STRESS_APP_ID=REPLACE
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@
 
 # Ignore master key for decrypting credentials and more.
 /config/master.key
+
+.env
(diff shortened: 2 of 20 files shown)


sweir27:
I was following some chatter in one of the platform channels the other day about how we want to start versioning artworks in gravity (I think?). Should we still merge this or is that the alternative/preferred approach?

ashkan18:
Id say lets merge since this also adds Sidekiq and we can remove that call later if needed.

PR #22, opened 2018-06-13
Add health check endpoint + some cleanup

ashkan18:
# Problem
We need health check endpoint 

Fixes: #19 

# Solution
Added `@GET api/health` which returns 
```json
{
  "healthy": true
}
```

Possibly we can do better checks by checking database connection and sidekiq queues, but for now this should be good enough.

# Other Changes
Refactored auth related code out of `ApplicationController` to new `AuthHandler`


app/controllers/application_controller.rb
```diff
@@ -1,39 +1,8 @@
 class ApplicationController < ActionController::API
   include ErrorHandler
+  include AuthHandler
```
joeyAghion: I actually think this should go into `app/controllers/concerns`.

ashkan18: also the `ErrorHandler`?

joeyAghion: Sure!

ashkan18:
going to merge 👍

Issue #16, opened 2018-06-12
Use micromachine to manage state


jonallured:
On Bearden I had a great experience using https://github.com/soveran/micromachine to manage the state in that application and I think it would be useful here in Stress. The pattern I used in Bearden was for the model to instantiate the machine and then provide methods to trigger state transitions with optional background jobs for side effects:

https://github.com/artsy/bearden/blob/master/app/models/import.rb

Then the actual machine class is where the states are setup and the valid transitions are defined:

https://github.com/artsy/bearden/blob/master/app/models/import_micro_machine.rb

ashkan18:
have you used https://github.com/aasm/aasm before? looking at it's usage and https://medium.com/geogo-in/state-machines-in-rails-5-45259a4f42da looks like it does some of the stuff with some helper methods for us already, I'm interested mostly in `may_<state>?` method and actual setting it with `<state>!`.

jonallured:
I confess that I have not used AASM before, but I have heard from other devs that it's big and unwieldy. What I liked about micromachine was that it was just one Ruby file and extremely easy to read and reason about.

PR #24, opened 2018-06-14
Introduce State Machine

ashkan18:
# Problem
Order state changes is a perfect use of state machines. Currently all states are kind of moving forward but who knows whats going to happen next.... we never know... never...

Fixes: #16 

# Solution
Followed @jonallured suggestion on using [micromachine](https://github.com/soveran/micromachine), evaluated using [aasm](https://github.com/aasm/aasm) but looks pretty heavy for what we want to do and @jonallured mentioned it's not the most loved gem out there. I'm planning to use state machine mostly for validation, and let actual setting of the state happen from the service layer. For that, `micromachine` looks like a nice little state machine handler.

Added new `OrderStates` under `Util` module which defines all of our states and possible _events(transitions)_. 

Updated `Order` to add some helper methods to provide checks if we can update to a new state, things like `order.may_submit?`, `order.may_reject?` and do validation using these.

Also provided helper methods for actually setting the state (without saving it). Im not saving it since sometimes we want to do something other than state change and then save and don't want to double save.

# Depends On
#23 
This PR was built on top of #23 thats why diff is bigger than what it should be, once thats merged this is more readable.


.env
```diff
@@ -0,0 +1,3 @@
+GRAVITY_API_ROOT=https://stagingapi.artsy.net
+GRAVITY_JWT=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJyb2xlcyI6InRydXN0ZWQsYXJ0c3kiLCJleHAiOjE4NDQ0NTE3OTMsImlhdCI6MTUyODgzMjYxNiwiYXVkIjoiNWIxZmU2MGY4MDRlNjUwMDAxYjQ4YmNlIiwiaXNzIjoiR3Jhdml0eSIsImp0aSI6IjViMjAyMjY4MjY0YzBmMDAwMTBlYjhlNiJ9.oJNfl4IfzcGoF2NO1knDMtzviP9uuH_kU9WIYZmJUow
+STRESS_APP_ID=5b1fe60f804e650001b48bce
```
starsirius: ditto. 🙈

ashkan18: pull request train got sidetracked...

app/models/util/order_states.rb
```diff
@@ -0,0 +1,47 @@
...
+    REJECTED = 'rejected',
+    FINALIZED = 'finalized'
+  ]
+
+  def self.machine
+    @machine ||= configure_machine
+  end
+
+  def self.configure_machine
+    mc = new(PENDING)
```
starsirius: This is probably fine but I wonder if the initial state of the state machine will be out-of-sync one day with the default state of the order [set on the database](https://github.com/artsy/stress/blob/8037c37d8e01a8bc0cffc92fec058adf91719de1/db/schema.rb#L41).

ashkan18: oh this class method one is only used to return `valid_states` and `events` the actual current (starting) state of order is defined/set here
https://github.com/artsy/stress/pull/24/files#diff-1c5adff4d5826555962bc3f7fcb4faf5R38

app/models/order.rb
```diff
@@ -1,33 +1,29 @@
...
     end
   end
 
+  Util::OrderStates.events.each do |event|
+    define_method "may_#{event}?" do
+      state_machine.trigger? event
+    end
+
+    define_method "#{event}" do
+      state_machine.trigger event
```
starsirius: We may consider `trigger!` and raise exceptions, instead of failing silently, when events are triggered from incompatible states.

app/services/order_service.rb
```diff
@@ -5,25 +5,50 @@ def self.create!(user_id:, partner_id:, currency_code:, line_items: [])
...
       order = Order.create!(user_id: user_id, partner_id: partner_id, currency_code: currency_code)
       line_items.each { |li| LineItemService.create!(order, li) }
+      # queue a job for few days from now to abandon the order
       order
     end
   end
 
   def self.submit!(order, credit_card_id:, shipping_info: '')
-    raise Errors::OrderError.new('Order cannot be submitted') unless order.pending?
+    raise Errors::OrderError.new('Order cannot be submitted') unless order.may_submit?
```
starsirius: The `may_event?` check is a proxy to the underlying state machine, and I wonder if it will be easier to just let `event` (like `approve`) fail and raise an exception via `trigger!`. We might need to customize the error message tho.

I am trying to think there may be domain requirements that we don't allow certain events, even it's totally valid on the state machine level...

app/models/order.rb
```diff
@@ -1,33 +1,29 @@
...
       self.state == state
     end
   end
 
+  Util::OrderStates.events.each do |event|
+    define_method "may_#{event}?" do
+      state_machine.trigger? event
+    end
+
+    define_method "#{event}" do
```
joeyAghion: No need for `"#{...}"`. You can probably just do `define_method event do...`.

app/models/order.rb
joeyAghion: Yeah this might even be a safer functionality to start with unless we'll be rigorously checking for `false` return values and rendering error responses.

app/models/util/order_states.rb
```diff
@@ -0,0 +1,47 @@
...
+    new(initial_state).tap { |machine| machine.configure(callback) }
+  end
+
+  def configure(callback = nil)
+    on(:any, &callback) if callback
+
+    self.when(:submit, PENDING => SUBMITTED)
+    self.when(:approve, SUBMITTED => APPROVED)
+    self.when(:finalize, APPROVED => FINALIZED)
+    self.when(:reject, APPROVED => REJECTED)
```
joeyAghion: Shouldn't this be `SUBMITTED => REJECTED`?

app/models/order.rb
```diff
@@ -37,4 +33,12 @@ def set_code
   def set_currency_code
     self.currency_code ||= 'usd'
   end
+
+  def state_machine
+    @state_machine ||= Util::OrderStates.start(state, method(:update_state))
```
joeyAghion: Reading through this and the micromachine docs, I think I'm in favor of much less meta-programming. I think this class could implement just this method like `@state_machine ||= build_state_machine`, where `build_state_machine` defines the states and allowed transitions all in one place. Since many of the state transitions will be driven by service methods, those can call the `.state_machine.<event>?` and `.state_machine.<event>!` directly without much additional code. We'd save the meta-programming and additional abstraction layer, and all of the calls would "standard" micromachine-style calls.

db/migrate/20180607135803_add_state_to_orders.rb
```diff
@@ -1,6 +1,6 @@
 class AddStateToOrders < ActiveRecord::Migration[5.2]
   def change
-    add_column :orders, :state, :string, default: Order::PENDING
+    add_column :orders, :state, :string, default: Util::OrderStates::PENDING
```
joeyAghion: I'd prefer _not_ setting a default here but passing `null: false`. The initial state should be specified in just one place (the service), and exceptions should be raised loudly if a row is inserted without a `state` value.

app/models/order.rb
```diff
@@ -17,16 +17,18 @@ class Order < ApplicationRecord
...
   before_save :set_currency_code
 
   scope :pending, -> { where(state: PENDING) }
 
   STATES.each do |state|
-    define_method "#{state}?" do
-      self.state == state
+    define_method "#{state}!" do
+      state_machine.trigger!(state)
+    rescue MicroMachine::InvalidState => e
```
ashkan18: not so sure about this, ideally i want service layer and graphql layer to deal with our internal errors and not with detail of what gem is used underneath.

ashkan18:
Thanks for all the feedback, kept everything in one file and separated actions and states, the only thing i don't like about it is somewhat a duplicate set of things but thats only the case because our actions map directly to states, i can see a world where they can be different. merging for now.

PR #27, opened 2018-06-15
Add spec for orders query

ashkan18:
Support query by partners based on user's jwt token and more spec.

Also as part of this, we replaces usages of `[]` for returning array of things to `connection_type` so now `orders` query returns a `connection` which means we now support all connection related pagination support and also we need to get the results using `edges` and `node`.

https://dev-blog.apollodata.com/explaining-graphql-connections-c48b7c3d6976

# Selfie

screen shot 2018-06-18 at 9 58 02 am

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -16,7 +16,14 @@ class Types::QueryType < Types::BaseObject
 
   def order(id:)
     order = Order.find(id)
-    raise GraphQL::ExecutionError, 'permission error' unless order.user_id == context[:current_user][:id]
+    raise GraphQL::ExecutionError, 'permission error' unless order.user_id == context[:current_user][:id] || context[:current_user][:partner_ids].include?(order.partner_id)
     order
+  end
+
+  def orders(params)
+    raise GraphQL::ExecutionError, 'requires one of userId or partnerId' unless params[:user_id].present? || params[:partner_id].present?
+    raise GraphQL::ExecutionError, 'Not permitted' if params[:user_id] && params[:user_id] != context[:current_user][:id]
+    raise GraphQL::ExecutionError, 'Not permitted' if params[:partner_id] && !context[:current_user][:partner_ids].include?(params[:partner_id])
+    Order.where(params)
   end
 end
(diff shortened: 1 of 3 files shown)


app/graphql/types/query_type.rb
```diff
@@ -16,7 +16,14 @@ class Types::QueryType < Types::BaseObject
 
   def order(id:)
     order = Order.find(id)
-    raise GraphQL::ExecutionError, 'permission error' unless order.user_id == context[:current_user][:id]
+    raise GraphQL::ExecutionError, 'permission error' unless order.user_id == context[:current_user][:id] || context[:current_user][:partner_ids].include?(order.partner_id)
```
williardx: What will this and other `GraphQL::ExecutionError` errors return for the client? Will it be a 500?

app/graphql/types/query_type.rb
```diff
@@ -16,7 +16,14 @@ class Types::QueryType < Types::BaseObject
...
     order
   end
+
+  def orders(params)
+    raise GraphQL::ExecutionError, 'requires one of userId or partnerId' unless params[:user_id].present? || params[:partner_id].present?
+    raise GraphQL::ExecutionError, 'Not permitted' if params[:user_id] && params[:user_id] != context[:current_user][:id]
+    raise GraphQL::ExecutionError, 'Not permitted' if params[:partner_id] && !context[:current_user][:partner_ids].include?(params[:partner_id])
+    Order.where(params)
```
williardx: I don't fully understand the role of `userId` here. I know we only want authorized users to query for a partner's order, but it looks like users can only query their own orders. As a user with partner access is it possible for me to get a permissions error when I want to view partner orders by being the "wrong" user?

app/graphql/types/query_type.rb
ashkan18: it basically returns 
```json
{
  data: null,
  errors: ["some error"]
}```

app/graphql/types/query_type.rb
ashkan18: so what happens here is, if you query using `userId` you have to be the user who owns that order.
If you are querying by `partnerId` you have to be a user who has access to that partner to be able to see orders of that partner. 

Thats what few validations there try to do, once we got more into this we may need to revisit some of these validations but i think for now this may be fine.

williardx:
LGTM, thanks @ashkan18!

PR #32, opened 2018-06-19
Add rubocop and random cleanup

jonallured:
This PR adds RuboCop and then complies with it's warnings. I did generate a todo file because I deemed things like addressing cyclomatic complexity as outside the scope of this work. I did fix some of the easy infractions on the todo file, so what's left should just be stuff that really would take refactoring time to address.


Gemfile
```diff
@@ -1,52 +1,27 @@
...
-
-# Use Rack CORS for handling Cross-Origin Resource Sharing (CORS), making cross-origin AJAX possible
-# gem 'rack-cors'
-
 group :development, :test do
-  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
-  gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
-  gem 'rspec-rails'
   gem 'graphlient'
+  gem 'rails-pry'
```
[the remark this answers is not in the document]
ashkan18: it came as part of rails default stuff, we can change it to that.

Gemfile
```diff
@@ -1,52 +1,27 @@
...
-
-# Use Capistrano for deployment
-# gem 'capistrano-rails', group: :development
-
-# Use Rack CORS for handling Cross-Origin Resource Sharing (CORS), making cross-origin AJAX possible
-# gem 'rack-cors'
-
 group :development, :test do
-  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
-  gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
```
ashkan18: why remove `byebug`?

ashkan18:
had one question about removing `byebug`. feel free to merge after 🍏

ashkan18:
fixes https://github.com/artsy/stress/issues/25

Gemfile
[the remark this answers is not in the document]
jonallured: Thanks for catching that!!

Gemfile
jonallured: The only reason I removed it was that I prefer pry - would you rather have both of them?

jonallured:
Ok, I finally got this green after banging my head against `RuboCop::Cop::Lint::ShadowedException`. With 8652df8, I've removed the redundant exception and that cop is satisfied. I left a comment to hopefully help the next dev to read that code.

Gemfile
ashkan18: yes i prefer `byebug` got used to it with `c` and `n` and other helpers :)

jonallured: Oh, I never noticed that pry didn't include that!! I added `pry-nav` with 2653613 and there's an easy way to get those shortcuts:

```ruby
# in your ~/.pryrc file
Pry.commands.alias_command 'c', 'continue'
Pry.commands.alias_command 's', 'step'
Pry.commands.alias_command 'n', 'next'
```

This way we have the best of both worlds! 😄

ashkan18: is `pry` really that much better?! these things come by default with `byebug` without adding anything custom!

jonallured: I mostly use pry to insert breakpoints and check the values of variables, something that can easily be done with byebug too, so I'd say for 90% of what I use these tools for, they are equal. That other 10% is doing things like showing the source of methods or listing the methods on an object, things that can be done other ways. I looked through the byebug docs and couldn't tell if it can do this, but either way, I'm just used to pry. I pretty much never think about stepping through code - I just insert another `binging.pry` call in the other places I want to land.

But there's no reason we can't have both! With 8921b6b I added byebug back in. You good with that?

yuki24: `pry` and `byebug` are two different implementations so they have different pros and cons. `pry` is colorful and pretty, and comes with no C extensions so it works on any Rubies including JRuby. `byebug` is a successor to MRI's debugger and can pause other threads for easier debugging. You may use the [`pry-byebug`](https://github.com/deivid-rodriguez/pry-byebug) gem so you can get most of both out of it.

ashkan18:
Thanks for bringing back `byebug`, its weird to have two gems doing the same thing but its a development only thing and not a big deal.

joeyAghion:
Total :ceiling-cat: here, but dependencies _do_ have a cost in terms of complicating upgrades via version constraints, lengthening boot time, and increasing memory footprint. (It's minor here, but in Gravity when we profiled the miserable boot time, the debugger library was significant, and we ended up using `require: false` as a result.)

At the very least, a [comment](http://joey.aghion.com/ruby-dependencies/) in the Gemfile might make clear the early rationale for including 2 similar dependencies and give a future contributor the context to evaluate whether to carry that forward.

PR #35, opened 2018-06-20
Rename to Exchange

ashkan18:
By popular demand 👍  going with Exchange :)

# Follow Up Changes
- [ ] Recreate database on rds shared staging
- [ ] Recreate k8s deployment
- [ ] Delete old DNS and create new one for staging
- [ ] do we need to rename Redis database assignment in infrastructure? https://github.com/artsy/infrastructure/blob/b75c22d067b28a6845213d21ee6336e083c0e90f/terraform/staging/consul.tf#L61 https://github.com/artsy/infrastructure/blob/b75c22d067b28a6845213d21ee6336e083c0e90f/terraform/production/consul.tf#L61 @joeyAghion ?


db/schema.rb
```diff
@@ -38,7 +38,7 @@
     t.string "partner_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.string "state", default: "pending"
+    t.string "state", null: false
```
williardx: Is this a change that you meant to make? I know that this file is autogenerated so just want to make sure.

joeyAghion: Good catch. Looks like this^ version is actually correct and the previous version was left over from a modified commit:

https://github.com/artsy/exchange/blob/3076235c73a89bf52ebc343cafe5df24271750cf/db/migrate/20180607135803_add_state_to_orders.rb#L3

williardx: Cool 👍

williardx:
Looks good to me - feel free to merge when you're ready.

db/schema.rb
ashkan18: yes, I think `schema.rb` wasn't updated in the PR that actually updated that column, most likely because I didn't updated constraint and just updated migration and it never ended up running migration locally for me, but when renamed database to `exchange` it had to do everything all over and updated `schema.rb`. Good catch 👀

PR #37, opened 2018-06-20
Force ssl on staging and prod, also add sidekiq service

ashkan18:
# Changes
- `force_ssl` on staging and production.
- Add SSL support to k8s setting.
- Add sidekiq service

https://artsyproduct.atlassian.net/browse/CP-58


joeyAghion:
👍

PR #42, opened 2018-06-21
Update README to reflect graphql schema changes

ashkan18:
# Problem
With https://github.com/artsy/metaphysics/pull/1120 in the works, we need to make sure everytime we update Exchange's GraphQL schema, we need to update the copy of it stored in Metaphysics.

# Solution
We [already added](#41 ) a danger rule to make sure it warns people if they've changed the schema, here we added some documentation about what needs to be done and updated danger file to point to this new doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,19 @@ rails db:setup
 $ rails s
 ```
 
+## Did You Change GraphQL Schema?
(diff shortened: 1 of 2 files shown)


williardx:
💯 📝

PR #45, opened 2018-06-25
Remove itemsTotalCents from schema, add as a field on order

ashkan18:
# Changes
- Remove `item_total_cents` from `Order` and instead calculate it based on the Order's line items.
- Expose `items_total_cents`, `shippint_total_cents`, `tax_total_cents`, `comission_total_cents`  and `transaction_fee_cents` in the schema.
- added spec for `order(id: ID!)` query

### DIFF
--- a/app/graphql/types/order_type.rb
+++ b/app/graphql/types/order_type.rb
@@ -8,7 +8,7 @@ class Types::OrderType < Types::BaseObject
   field :partner_id, String, null: false
   field :state, Types::OrderStateEnum, null: false
   field :currency_code, String, null: false
-  field :items_total_cents, Integer, null: true
+  field :items_total_cents, Integer, null: false
   field :shipping_total_cents, Integer, null: true
   field :tax_total_cents, Integer, null: true
   field :transaction_fee_cents, Integer, null: true
--- a/app/graphql/types/order_type.rb
+++ b/app/graphql/types/order_type.rb
@@ -8,6 +8,11 @@ class Types::OrderType < Types::BaseObject
   field :partner_id, String, null: false
   field :state, Types::OrderStateEnum, null: false
   field :currency_code, String, null: false
+  field :items_total_cents, Integer, null: true
+  field :shipping_total_cents, Integer, null: true
+  field :tax_total_cents, Integer, null: true
+  field :transaction_fee_cents, Integer, null: true
+  field :commission_fee_cents, Integer, null: true
   field :created_at, Types::DateTimeType, null: false
   field :updated_at, Types::DateTimeType, null: false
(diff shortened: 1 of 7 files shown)


app/graphql/types/order_type.rb
```diff
@@ -8,6 +8,11 @@ class Types::OrderType < Types::BaseObject
   field :partner_id, String, null: false
   field :state, Types::OrderStateEnum, null: false
   field :currency_code, String, null: false
+  field :items_total_cents, Integer, null: true
```
williardx: Is there ever a case where this would be `null`?

app/graphql/types/query_type.rb
```diff
@@ -38,7 +38,7 @@ def orders(params)
 
   private
 
-  def validate_params!(params)
+  def validate_orders_params!(params)
```
williardx: 👍

app/graphql/types/order_type.rb
ashkan18: good point! this one no (i'll update) but i was wondering for other price related fields it's possible, i feel like `shipping_total_cents` being `nil` is different than `shipping_total_cents` being 0. Nil to me means partner never set it, while 0 means partner actually is doing free shipping.

app/graphql/types/query_type.rb
ashkan18: there may be a better way to handle queries, noticed we have both queries in one file and my suggested `validate_params` is too generic and only for `orders`, ended up renaming it, maybe later we can look into how to have one query per file so we can be more focused, kind of similar to current mutations.

williardx: Do they actually need to be separate? We could fold `order` into `orders` and just add `id` as another optional parameter.

ashkan18: so the biggest difference is, `orders` returns a connection meaning:
- it can get pagination related params
- it can get sort param
- it returns an array
I think things would be lot simpler to have them as two separate queries since the ideal data structure returning is mainly different and instead of two simple queries we'd have one complicate one.
By separating them i meant, think of the case where we started exposing more query options, like query by line item or order history changes or transactions, this file can get bigger and bigger with different types of validation being all over. In case of mutation we have one `mutation` file per each mutation which makes that class really simple and easy to follow.

williardx: Cool, makes sense 👍 Agreed then that we'll want to separate out these two queries into their own files at some point.

williardx:
LGTM, feel free to merge whenever you're ready!

ashkan18:
thanks 🙌

Issue #47, opened 2018-06-27
Relax pending order creation 


ashkan18:
follow up on discussion: https://artsy.slack.com/archives/CB110C6LE/p1530118511000348

ashkan18:
# Current State
When creating an order by a user, we check if this user has any existing pending order for this artwork/editionset and if they do we raise a validation error and don't allow creating a new order for the same artwork. In current state a user can have multiple pending orders for different artworks at a time.

# Proposed Change
A user can have only one _pending_ order at a time. When user tries to create a new order, we check if they have any pending order, we abandon their existing pending order and then we create a new one.

joeyAghion:
This proposal sounds like it would support v0 well. It's worth keeping in mind how implementations may need to change to support line item quantities of >1, or in the further future when multiple items [from the same partner, probably] can be in a single order.

If clients invoke a "createOrder" mutation, that should presumably continue to create a _new_ order. If quantities >1 are to be supported, the affordance offered to buyers should maybe change to behave more like an "addToCart," where by default it starts an order with that item, but if the item is already in the pending order it increments the item's quantity. Deployed clients (like Eigen) might continue to invoke the "createOrder" operation and it could continue to work the old way.

If orders must be associated with a single partner and we switch to more of a "cart" behavior, we'll need to handle the case when a pending order exists but can't be added to, because it's associated with a different partner. At that point we either need to abandon the existing one or support multiple pending orders. We can leave that decision for later, and can depend on optional params (like `usePendingOrder=[if_possible|never]`) to ease any transition.

starsirius:
I like to think about a pending order as a shopping cart in the existing e-commerce site experience as Amazon. It makes sense to me there is no clear use case, at least for the pilot, to support multiple pending orders per user (even on different works). It's also a familiar mental model/good UX for buyers.

The only case I can think of right now is, as @joeyAghion mentioned, if we don't allow checking out the same order with multiple partners, and agreed that we can worry about that later. I believe there is room for product to develop the features and experiences. FWIW, on the actual payment side, Stripe supports (in public beta) [creating separate charges and transfers](https://stripe.com/docs/connect/charges-transfers).

PR #48, opened 2018-06-27
Relax createOrder mutation

ashkan18:
# Change
fixes https://github.com/artsy/exchange/issues/47

# What changed?
We added new `abandon` action to `Order`. Now when we create order, we first check if there are any _pending_ orders we abandon them and then we create the order.

### DIFF
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -1,8 +1,8 @@
 module OrderService
   def self.create!(user_id:, partner_id:, currency_code:, line_items: [])
     raise Errors::OrderError, 'Currency not supported' unless valid_currency_code?(currency_code)
-    raise Errors::OrderError, 'Existing pending order' if create_params_has_pending_order?(user_id, line_items)
     Order.transaction do
+      abandon_pending_orders!(user_id) if pending_order?(user_id)
       order = Order.create!(user_id: user_id, partner_id: partner_id, currency_code: currency_code, state: Order::PENDING)
       line_items.each { |li| LineItemService.create!(order, li) }
       # queue a job for few days from now to abandon the order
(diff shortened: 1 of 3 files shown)


ashkan18:
oh thats a good point about `abandoned` and `replaced`, since these two states are really similar, basically they abandon an order when they create a new one, im more towards not adding a new state for this, reporting-wise we can probably check the order creation time and compare it to when it was abandoned and find out if it was abandoned abandoned or replaced abandoned :)

app/services/order_service.rb
```diff
@@ -50,12 +50,25 @@ def self.reject!(order)
...
+  end
+
   def self.user_pending_artwork_order(user_id, artwork_id, edition_set_id = nil)
     Order.pending.joins(:line_items).find_by(user_id: user_id, line_items: { artwork_id: artwork_id, edition_set_id: edition_set_id })
   end
 
-  def self.create_params_has_pending_order?(user_id, line_items)
-    line_items.map { |li| user_pending_artwork_order(user_id, li[:artwork_id], li[:edition_set_id]) }.any?
+  def self.pending_order?(user_id)
+    Order.pending.where(user_id: user_id).exists?
```
williardx: Won't this check for any pending order? From my understanding we don't want more than one pending order for a given artwork at a time, but we could have multiple pending orders for multiple different artworks.

app/services/order_service.rb
```diff
@@ -50,12 +50,25 @@ def self.reject!(order)
...
+  end
+
+  def self.abandon_pending_orders!(user_id)
```
williardx: Similarly won't this abandon all pending orders instead of just one duplicate pending order?

ashkan18: yes, actually the latest decision (formerly known as option 4 :) ) was to have only pending order per user regardless of the artwork.

williardx: Got it! Looks good then.

williardx:
LGTM, thanks!

PR #49, opened 2018-06-29
WIP: Track order history using RecordHistoryService

williardx:
- Adds `RecordHistoryService` which creates a history object with changes for any given record configured to do so.
- Adds `history` child object to Order and creates `order_histories` table
- Tracks changes to orders following mutations in `OrderService`

Will take off WIP label once I write some tests.

### Next
- Expose history to GraphQL

### DIFF
--- a/app/models/order_history.rb
+++ b/app/models/order_history.rb
@@ -1,0 +1,3 @@
+class OrderHistory < ApplicationRecord
+  belongs_to :order
+end
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -4,47 +4,50 @@ module OrderService
     Order.transaction do
       abandon_pending_orders!(user_id) if pending_order?(user_id)
       order = Order.create!(user_id: user_id, partner_id: partner_id, currency_code: currency_code, state: Order::PENDING)
+      RecordHistoryService.create!(order, user_id, order.attributes)
       line_items.each { |li| LineItemService.create!(order, li) }
       # queue a job for few days from now to abandon the order
       order
     end
   end
 
-  # rubocop:disable Lint/UnusedMethodArgument
-  def self.submit!(order, credit_card_id:, shipping_info: '')
+  def self.submit!(order, user_id, credit_card_id:)
     Order.transaction do
       # verify price change?
       order.credit_card_id = credit_card_id
       # TODO: hold the charge for this price on credit card
       order.submit!
       order.save!
+      RecordHistoryService.create!(order, user_id, state: order.state, credit_card_id: order.credit_card_id)
     end
     order
   end
-  # rubocop:enable Lint/UnusedMethodArgument
 
-  def self.approve!(order)
+  def self.approve!(order, user_id)
     Order.transaction do
       order.approve!
       order.save!
+      RecordHistoryService.create!(order, user_id, state: order.state)
       # TODO: process the charge by calling gravity with current credit_card_id and price
     end
     order
   end
 
-  def self.finalize!(order)
+  def self.finalize!(order, user_id)
     Order.transaction do
       order.finalize!
       order.save!
+      RecordHistoryService.create!(order, user_id, state: order.state)
       # TODO: process the charge by calling gravity with current credit_card_id and price
     end
     order
   end
 
-  def self.reject!(order)
+  def self.reject!(order, user_id)
     Order.transaction do
       order.reject!
       order.save!
+      RecordHistoryService.create!(order, user_id, state: order.state)
       # TODO: release the charge
     end
     order
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -57,6 +57,7 @@ module OrderService
     Order.transaction do
       order.abandon!
       order.save!
+      RecordHistoryService.create!(order, user_id, state: order.state)
     end
   end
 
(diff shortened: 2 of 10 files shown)


joeyAghion:
Nice!

(As usual, I only have comments about naming.)

I have a mild preference for _default_ association names, like `order.order_histories` in this case. The current language actually reads quite nicely, but much later when the app is larger and there are more things to hold in one's head, having to recall what's default and what's custom or what's singular and what's plural becomes unnecessarily complex.

Regarding the `RecordHistoryService` name, my first impulse was to recommend calling it `OrderHistoryService` or just implementing its method within the `OrderService`. Looking at it again, maybe you were hoping it could be generalized beyond orders? Interesting.

app/models/order_history.rb
```diff
@@ -0,0 +1,3 @@
+class OrderHistory < ApplicationRecord
```
joeyAghion: _If_ your intent is to support histories more generally, this could be a simple `History` class with a polymorphic `parent` (or `model`) association instead of hard-coding `order`. Then again, if that's what this aims to do you might be able to incorporate some off-the-shelf gems for that purpose.

app/services/order_service.rb
```diff
@@ -4,47 +4,50 @@ def self.create!(user_id:, partner_id:, currency_code:, line_items: [])
...
       # TODO: process the charge by calling gravity with current credit_card_id and price
     end
     order
   end
 
-  def self.reject!(order)
+  def self.reject!(order, user_id)
     Order.transaction do
       order.reject!
       order.save!
```
joeyAghion: Unrelated to this PR, I'm curious why all these `save!`s are necessary immediately following a state change. Do those `...!` methods that update the state not also save? (By convention, `!`-methods do save.)

williardx:
Re: `order.order_histories`, that makes sense -- optimizing for long-term readability sounds like the right move. One nice thing about having it called `history` is that it's easier to reuse `RecordHistoryService` across multiple objects if each record has an associated object called `history` and not `classname_history`. Though I imagine there's a reasonable way of getting around that in Ruby?

And yep, the idea is to reuse this across multiple objects (e.g. offers).

app/models/order_history.rb
williardx: Oh, that's a nice idea! All record histories would then be in one single history table then?

Re: off the shelf gems, that's also reasonable. It seemed like our needs were pretty lightweight and so didn't immediately think to jump to a gem, but that might ultimately be a better solution if our needs grow at all. [Paper Trail](https://github.com/paper-trail-gem/paper_trail) seems reasonable. @ashkan18 do you have any thoughts on that?

joeyAghion: I'm actually not sure what the _current_ needs are for this. If this data will be used operationally within this system (e.g., to serve order history data via API), a dedicated `OrderHistory` implementation might make sense, since it could be customized for that use, such as by modeling state changes in a first-class way rather than bundling them into the `jsonb` field. If it's just an audit trail for internal inspection, something like papertrail might be 👍.

yuki24: @ashkan18 @williardx What's the reason why you didn't use gems like the [paper_trail](https://github.com/paper-trail-gem/paper_trail) gem?

app/services/order_service.rb
ashkan18: so here state machine actually sets the value but doesn't save it, mainly i wanted to avoid double saving, so in case of submit we want to set `credit_card_id` and also update the state, that `!` will raise exception if state is unacceptable though.
thinking more about it, it might be ok to save since most of these actions are wrapped in Transaction. Also we could set attributes before changing state and save in change state method.

app/models/order_history.rb
ashkan18: so the original use case here was to store state changes and track who made the change, then as part of that we ended up thinking about a more generic way of tracking changes. I haven't used `paper_trail` before but sounds interesting, i don't have strong opinion here, at the moment we don't have a UI as far as i know for showing state history changes but I can imagine at some point we do want to know when one order was submitted and when it was approved. 
So two options here could be:
1) add a generic track history where we track every change on the model and the modifier
2) add specific order state history table and just track state changes
Maybe we can go with 1 for now and once that UI becomes a thing we can provide 2 and track state changes as a first class thing and as for 1 if `paper_trail` is easy enough to add and not too heavy of a gem :) im 👍

app/services/order_service.rb
joeyAghion: I see. The `!` convention _is_ mostly about raising exceptions but in activerecord contexts often _also_ indicates a save. This is one of those ambiguous cases. It's fine as long as those state-changing `...!` methods are always consistently followed by a save I guess.

app/models/order_history.rb
joeyAghion: I wouldn't spend time on things that don't have specific uses yet. If we do at some point want to know when an order was submitted or approved (which are specific, critical transitions that can only happen once, as far as I understand), the simplest solution seems like it would be to track those timestamps on the top-level order, even if there's also some separate audit trail.

williardx: From looking at Paper Trail it looks like it'll do everything we need (and then some). It'll also store changes in `jsonb` which is what we were planning on doing. So sounds like the next step here is to just install Paper Trail, add tracking to the order model and revisit order state tracking once we're ready. I'll close this PR out and open a new one.

ashkan18:
thanks @williardx 👍

PR #50, opened 2018-06-29
Add Paper Trail and track changes on Order

williardx:
Following up on discussion in https://github.com/artsy/exchange/pull/49, adding Paper Trail and tracking changes in Order.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -48,5 +48,15 @@ ActiveRecord::Schema.define(version: 2018_07_02_162920) do
     t.index ["user_id"], name: "index_orders_on_user_id"
   end
 
+  create_table "versions", force: :cascade do |t|
+    t.string "item_type", null: false
+    t.integer "item_id", null: false
+    t.string "event", null: false
+    t.string "whodunnit"
+    t.jsonb "object"
+    t.datetime "created_at"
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -27,6 +27,15 @@
     t.index ["order_id"], name: "index_line_items_on_order_id"
   end
 
+  create_table "order_histories", force: :cascade do |t|
```
ashkan18: i think this is left here since your local database still has this table most likely. Not sure whats the best way to fix this but i normally drop my database and migrate everything again.
```
rails db:drop
rails db:migrate
```
Normally this should fix it.

db/schema.rb
```diff
@@ -46,5 +55,16 @@
...
+    t.integer "item_id", null: false
+    t.string "event", null: false
+    t.string "whodunnit"
+    t.jsonb "object"
+    t.datetime "created_at"
+    t.index ["item_type", "item_id"], name: "index_versions_on_item_type_and_item_id"
+  end
+
   add_foreign_key "line_items", "orders"
+  add_foreign_key "order_histories", "orders"
```
ashkan18: this is probably from before too.

db/schema.rb
williardx: Good catch! I'll fix that.

PR #51, opened 2018-07-02
Add `state_updated_at` and `state_expires_at`

ashkan18:
# Problem
In CMS UI we want to be able to show when the current state (if) expires.

# Solution
Added `state_updated_at` and `state_expires_at` fields to reflect when order state was updated and whats the current expiry, knowing that not all states may expire, this field 

Keep in mind, currently we don't know exactly which states are going to expire and what is the threshold but for now setting two states with some basic thresholds.

cc: @sepans @jonallured

### DIFF
--- a/spec/models/order_spec.rb
+++ b/spec/models/order_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe Order, type: :model do
       expect(order.state_updated_at).to eq current_timestamp
     end
 
-    it 'does not update timestamps if state changed to state with expiration' do
+    it 'updates timestamps if state changed to state with expiration' do
       current_timestamp = order.state_updated_at
       current_expiration = order.state_expires_at
       order.submit!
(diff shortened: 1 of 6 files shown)


spec/models/order_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      expect(order.state_expires_at).to eq order.state_updated_at + 2.days
+    end
+
+    it 'does not update timestamps if state did not change' do
+      current_timestamp = order.state_updated_at
+      order.update!(currency_code: 'CAD')
+      expect(order.state_updated_at).to eq current_timestamp
+    end
+
+    it 'does not update timestamps if state changed to state with expiration' do
```
williardx: I'm a little confused by this test - doesn't this check that `order.state_updated_at` and `order.state_expires_at` both change? The test description makes me think they shouldn't

ashkan18: good catch! this is me being lazy and copy/paste :)

williardx:
LGTM, thanks!

ashkan18:
Yes, once thats merged you can use it 🕺

PR #52, opened 2018-07-02
Add spec and simplify price computation

jonallured:
I happened to notice a bug here, so I wrote a failing test ("with a line item that has no price") and then fixed it. ❤️

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -41,7 +41,7 @@ class Order < ApplicationRecord
   end
 
   def items_total_cents
-    (line_items.present? && line_items.map(&:price_cents).reduce(0, :+)) || 0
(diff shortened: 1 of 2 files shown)


ashkan18:
Nice 💯

app/models/order.rb
```diff
@@ -41,7 +41,7 @@ class Order < ApplicationRecord
   end
 
   def items_total_cents
-    (line_items.present? && line_items.map(&:price_cents).reduce(0, :+)) || 0
+    line_items.pluck(:price_cents).sum
```
yuki24: I wonder if this could just be [`line_items.sum(:price_cents)`](http://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-sum).

PR #56, opened 2018-07-10
Add totalCents and subtotalCents

ashkan18:
# Problem
We were calculating total and subtotal in CMS. We want to have one source of truth for totals and subtotals and it makes sense for that to be Exchange since it's the place that will ultimately calculate these values and call Stripe with that for holding the charge.

# Solution
Add `totalCents` and `subtotalCents` to `Order` and `OrderType`.

# Follow up
Add these to MP.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -42,7 +42,15 @@ class Order < ApplicationRecord
   end
 
   def items_total_cents
-    line_items.pluck(:price_cents).sum
+    line_items.sum(:price_cents)
+  end
+
+  def subtotal_cents
+    items_total_cents + shipping_total_cents.to_i + tax_total_cents.to_i
+  end
+
+  def total_cents
+    subtotal_cents + commission_fee_cents.to_i + transaction_fee_cents.to_i
   end
 
   private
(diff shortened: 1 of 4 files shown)


app/models/order.rb
```diff
@@ -42,7 +42,15 @@ class Order < ApplicationRecord
   end
 
   def items_total_cents
-    line_items.pluck(:price_cents).sum
+    line_items.sum(:price_cents)
```
jonallured: Nice one!! 💯

app/models/order.rb
```diff
@@ -42,7 +42,15 @@ class Order < ApplicationRecord
...
+  end
+
+  def subtotal_cents
+    items_total_cents + shipping_total_cents.to_i + tax_total_cents.to_i
+  end
+
+  def total_cents
+    subtotal_cents + commission_fee_cents.to_i + transaction_fee_cents.to_i
```
jonallured: Are these stores as negative numbers? I expected to see these subtracted.

ashkan18: good catch! fixing now.

jonallured:
Thanks for this!

PR #57, opened 2018-07-10
Authorize charge on submission of order

williardx:
Addresses [CP-66](https://artsyproduct.atlassian.net/browse/CP-66)

- Adds Stripe to Exchange
- Adds new Transaction model to record all Stripe payments
- Authorizes charge on user's credit card on order submission and catches all Stripe `CardError` errors
- Modifies `submitOrder` mutation to require `destination_account_id` and adds it to Order model.

~**WIP:**~
- [x] Verifying merchant account ID following https://github.com/artsy/gravity/pull/11810
- [x] Waiting for clarification on when authorization will occur in order process
- [x] Write tests


app/graphql/types/order_type.rb
```diff
@@ -18,4 +18,5 @@ class Types::OrderType < Types::BaseObject
   field :state_updated_at, Types::DateTimeType, null: true
   field :state_expires_at, Types::DateTimeType, null: true
   field :line_items, Types::LineItemType.connection_type, null: true
+  field :destination_account_id, String, null: true
```
ashkan18: i wonder if we should expose this, id say for now lets not expose them in the schema till we find a usage for it. Looking at this purely from Exchange perspective, we are exposing partner's stripe account id to users and user's payment source id to partners.

williardx: That's fair! I'll take it out.

app/graphql/mutations/submit_order.rb
```diff
@@ -3,18 +3,19 @@ class Mutations::SubmitOrder < Mutations::BaseMutation
 
   argument :id, ID, required: true
   argument :credit_card_id, String, required: true
+  argument :destination_account_id, String, required: true
```
ashkan18: we are going to rename that in following PR to `merchant_account_id` based on our discussion on Slack.

app/services/order_service.rb
```diff
@@ -10,18 +10,21 @@ def self.create!(user_id:, partner_id:, currency_code:, line_items: [])
...
-  def self.submit!(order, credit_card_id:, shipping_info: '')
+  def self.submit!(order, credit_card_id:, destination_account_id:)
     Order.transaction do
       # verify price change?
       order.credit_card_id = credit_card_id
-      # TODO: hold the charge for this price on credit card
+      order.destination_account_id = destination_account_id
       order.submit!
+      charge = PaymentService.authorize_charge(order, order.items_total_cents)
+      TransactionService.create_success!(order, charge)
```
ashkan18: this is super minor, but i feel like it reads better if we move `order.submit!` below authorize charge. what do you think?

app/services/transaction_service.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def self.create_success!(order, charge)
+    order.transactions.create!(
+      external_id: charge.id,
+      source_id: order.credit_card_id,
+      destination_id: order.destination_account_id,
+      amount_cents: charge.amount,
+      failure_code: charge.failure_code,
+      failure_message: charge.failure_message,
+      status: 'success'
```
ashkan18: looks like this can only be `success` or `failure` lets just use `success` as a boolean here.

spec/services/order_service_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'rails_helper'
+
+describe OrderService, type: :services do
+  let(:order) { Fabricate(:order) }
+  let(:credit_card_id) { 'cc-1' }
+  let(:destination_account_id) { 'destination_account' }
+  let(:charge_success) { { id: 'ch-1' } }
+  let(:charge_failure) { { failure_message: 'some_error' } }
```
ashkan18: we should switch to use https://github.com/rebelidealist/stripe-ruby-mock for these.

PR #58, opened 2018-07-11
Shipping & Payment Mutations

ashkan18:
# Changes
We have had [some discussions](https://github.com/artsy/exchange/issues/54) about having separate mutations for setting shipping and payment info. Here we do the following:
- Added new `setShipping` mutation that gets newly added shipping related fields:
  - Added `ship_type` to `Order` which currently can be `SHIP` or `PICKUP`.
- Added new `setPayment` mutation.
- Renamed `credit_card_id` to `payment_source`. This way Exchange can create charges for both stored credit card ids or with tokenized credit card. 
- Updated `submitOrder` to only get order `id` and check if order is setup for submission (has shipping and payment info)


README.md
```diff
@@ -124,7 +124,7 @@ For input:
 {
   "input": {
     "id": "<order id>",
-    "creditCardId": "<credit card id>"
+    "paymentSource": "<credit card id>"
```
joeyAghion: As discussed elsewhere, I think Exchange should expect "core" credit card IDs here rather than a flexible set of stripe ids or tokens. There's more general value to storing payment information, so [ideally] we should simplify by supporting just 1 approach.

app/graphql/mutations/set_payment.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  argument :id, ID, required: true
+  argument :payment_source, String, required: false
+
+  field :order, Types::OrderType, null: true
+  field :errors, [String], null: false
+
+  def resolve(args)
+    order = Order.find(args[:id])
+    validate_request!(order)
```
joeyAghion: This is a nice pattern. I've seen `assert_...` used for this sort of thing that raises errors (like `assert_order_can_set_payment!` in this case), which avoids the overloaded "validation" terminology.

app/graphql/mutations/set_shipping.rb
```diff
@@ -0,0 +1,28 @@
+class Mutations::SetShipping < Mutations::BaseMutation
+  null true
+
+  argument :id, ID, required: true
+  argument :shipping_address, String, required: false
+  argument :shipping_city, String, required: false
+  argument :shipping_country, String, required: false
+  argument :shipping_postal_code, String, required: false
+  argument :shipping_type, Types::OrderShippingTypeEnum, required: false
```
joeyAghion: No state (or "region," to use the more global term)?

app/graphql/types/order_shipping_type_enum.rb
```diff
@@ -0,0 +1,4 @@
+class Types::OrderShippingTypeEnum < Types::BaseEnum
+  value 'PICKUP', 'order will be picked up by collector', value: Order::PICKUP
+  value 'SHIP', 'order will be shipped to collector', value: Order::SHIP
+end
```
joeyAghion: (Minor/language) It's a little strange to have "pickup" being a "shipping" type. Maybe they are actually "delivery" or "fulfillment" types? Or maybe there's simply a `true`/`false` `ship` flag?

app/models/order.rb
```diff
@@ -21,6 +21,11 @@ class Order < ApplicationRecord
     'submitted' => 2.days
   }.freeze
 
+  SHIPPING_TYPES = [
+    PICKUP = 'pickup'.freeze,
+    SHIP = 'ship'.freeze
+  ].freeze
```
joeyAghion: Can this reuse the enum rather than repeat?

app/models/order.rb
```diff
@@ -45,6 +50,15 @@ def items_total_cents
     line_items.pluck(:price_cents).sum
   end
 
+  def shipping_info?
+    shipping_type == PICKUP ||
+      (shipping_type == SHIP && %i[shipping_address shipping_city shipping_country shipping_postal_code].any? { |sh_field| send(sh_field).present? })
```
joeyAghion: Do you mean `all?` instead of `any?`?

app/services/order_service.rb
```diff
@@ -10,24 +10,46 @@ def self.create!(user_id:, partner_id:, currency_code:, line_items: [])
     end
   end
 
-  # rubocop:disable Lint/UnusedMethodArgument
-  def self.submit!(order, credit_card_id:, shipping_info: '')
+  def self.set_payment!(order, payment_source:)
+    raise Errors::OrderError, 'Cannot set payment info on non-pending orders' unless order.state == Order::PENDING
```
joeyAghion: It's curious how this operation can be rejected at the mutation _or_ service layer, for different reasons. Is that intentional, or should they be consolidated?

app/services/order_service.rb
```diff
@@ -10,24 +10,46 @@ def self.create!(user_id:, partner_id:, currency_code:, line_items: [])
...
+    Order.transaction do
+      order.update!(payment_source: payment_source)
+    end
```
joeyAghion: Why all these transactions? Are there callbacks you're concerned about? Similar question about the `reload`s.

app/graphql/mutations/set_shipping.rb
```diff
@@ -0,0 +1,28 @@
+class Mutations::SetShipping < Mutations::BaseMutation
+  null true
+
+  argument :id, ID, required: true
+  argument :shipping_address, String, required: false
```
joeyAghion: I might call this `..._street` to distinguish it from the whole set of fields that make up a "shipping address."

app/graphql/mutations/set_shipping.rb
ashkan18: so this is basically `delivery_type`

app/models/order.rb
ashkan18: the enum is `GraphQLEnum` i think it makes more sense to have that one use this.

app/models/order.rb
ashkan18: actually that logic is wrong, this was around the time that we said we can create shipping estimate with just having the country and i got confused here, will fix this.

app/services/order_service.rb
ashkan18: i've been trying to have more like controller level validations (check if current user has access to this order or not for example) in the mutation, but do more business level validation in the actual service. 
Basically have services be responsible for validating the business layer stuff (does order have proper set of fields to be submitted?)
And have mutation layer to be responsible for more controller level validation (can this user make this request?)

app/services/order_service.rb
ashkan18: so im planning/hoping not to use callbacks for business logic stuff here but i think `papertrail` might already be using them. I guess in case of set payment not much can go wrong since we are just setting one field but in other cases (lets say setting shipping) we may still keep the transaction since we will do more stuff in that step (estimate cost and etc).
`reload` yes! looking at it now, it seems not needed.

README.md
ashkan18: lets actually talk about it tomorrow (feels weird mentioning times in github comments!)

db/migrate/20180709195222_add_shipping_address_to_orders.rb
```diff
@@ -1,11 +1,11 @@
 class AddShippingAddressToOrders < ActiveRecord::Migration[5.2]
   def change
     change_table :orders, bulk: true do |t|
-      t.column :shipping_address, :string
+      t.column :shipping_street, :string
```
williardx: What is `shipping_street` exactly? Is that the address number + street? What if someone needs two lines for their street address?

joeyAghion: I assumed it was street address (i.e., street number, name, suffix, c/o or att:, company, etc.). This could (and maybe should) accept multi-line strings, though I've seen it modeled as `street1` and `street2` elsewhere if that's more convenient.

ashkan18: looking at [Stripe's docs](https://stripe.com/docs/api#create_charge-shipping-address-line1) maybe we can go with `shipping_address_line1` and `shipping_address_line2`? or just `shipping_line1` and `shipping_line2`.

README.md
sweir27: Curious about the outcome of this ☝️

williardx:
Looks good! Thanks @ashkan18

PR #59, opened 2018-07-16
Fetch merchant_account_id directly from Gravity

williardx:
- Removes `merchant_account_id` from `order` model and `submitOrder` mutation
- Fetches `merchant_account_id` directly from Gravity

### DIFF
--- a/app/graphql/mutations/set_payment.rb
+++ b/app/graphql/mutations/set_payment.rb
@@ -3,7 +3,6 @@ class Mutations::SetPayment < Mutations::BaseMutation
 
   argument :id, ID, required: true
   argument :credit_card_id, String, required: false
-  argument :merchant_account_id, String, required: false
 
   field :order, Types::OrderType, null: true
   field :errors, [String], null: false
@@ -20,7 +19,7 @@ class Mutations::SetPayment < Mutations::BaseMutation
   end
 
   def assert_order_can_set_payment!(order, args)
-    raise Errors::ApplicationError, 'Missing required arguments' unless args[:merchant_account_id].present? || args[:credit_card_id].present?
+    raise Errors::ApplicationError, 'Missing required arguments' if args[:credit_card_id].blank?
     raise Errors::AuthError, 'Not permitted' unless context[:current_user]['id'] == order.user_id
   end
 end
--- a/app/graphql/mutations/set_payment.rb
+++ b/app/graphql/mutations/set_payment.rb
@@ -2,7 +2,7 @@ class Mutations::SetPayment < Mutations::BaseMutation
   null true
 
   argument :id, ID, required: true
-  argument :credit_card_id, String, required: false
+  argument :credit_card_id, String, required: true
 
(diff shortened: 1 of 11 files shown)


app/graphql/mutations/set_payment.rb
```diff
@@ -20,7 +19,7 @@ def resolve(args)
   end
 
   def assert_order_can_set_payment!(order, args)
-    raise Errors::ApplicationError, 'Missing required arguments' unless args[:merchant_account_id].present? || args[:credit_card_id].present?
+    raise Errors::ApplicationError, 'Missing required arguments' if args[:credit_card_id].blank?
```
ashkan18: we used to have the check here mainly because `credit_card_id` and `merchant_account_id` where both not required but in one request we at least needed one of them. Now that `set_payment` gets only `credit_card_id` i think it's safe to remove this check and just use GraphQL's `required: true` [here](https://github.com/artsy/exchange/pull/59/files#diff-434eacd08ac7a67ad8e12a51f2f3edf5L5)

app/services/payment_service.rb
```diff
@@ -1,27 +1,29 @@
 module PaymentService
   def self.authorize_charge(order, amount)
-    raise Errors::PaymentError.new('Invalid destination account id', failure_code: 'artsy_internal', failure_message: 'Invalid destination account id') unless valid_destination_account?(order)
+    merchant_account_id = get_merchant_account_id(order.partner_id)
```
ashkan18: i think it would be cleaner here if we get full `merchant_account` and not just the id, and then from that `merchant_account` we'd use `external_id` which would be Stripe's destination. I think calling current variable `merchant_account_id` might be little misleading since this is actually `metchant_account.external_id`.

app/services/payment_service.rb
```diff
@@ -1,27 +1,29 @@
...
+    raise Errors::PaymentError.new('Partner does not have merchant account ID', failure_code: 'artsy_internal', failure_message: 'Partner does not have merchant account ID') if merchant_account_id.nil?
```
ashkan18: with ☝️ then this would be `partner does not have a merchant account`.

app/services/payment_service.rb
```diff
@@ -1,27 +1,29 @@
 module PaymentService
   def self.authorize_charge(order, amount)
```
ashkan18: what do you think about moving `get_merchant_account` method to `OrderService` or event new `MerchantAccountService` and instead pass `desination_id` to `authorize_charge` so basically `authorize_charge(order, amount, source_credit_card_id, destination_id)` this way this service doesn't have to hold logic about other order related logics and etc.

williardx: Okay! I can see the logic in that -- I think a new `MerchantAccountService` makes sense, which might evolve into a `SellerService` later on.

ashkan18:
👍 merging, we will follow up on a separate PR for service layer refactoring

PR #60, opened 2018-07-16
Create new order all the time

ashkan18:
# Problem
Current pending order validation is expensive and doesn't really provide much, After some discussions, it seems the easiest solution for now would be to always create new order regardless of existing pending orders. This is fine since all subsequent mutations work with `order` id and they should be fine. 

# Solution
Always create new orders and don't even check for existing orders.

We will be creating pending charges but they can be cleaned up in many ways by queuing jobs or daily task that cleans up old pending orders.

### DIFF
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -2,10 +2,8 @@ module OrderService
   def self.create!(user_id:, partner_id:, currency_code:, line_items: [])
     raise Errors::OrderError, 'Currency not supported' unless valid_currency_code?(currency_code)
     Order.transaction do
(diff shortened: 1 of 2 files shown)


joeyAghion:
This seems fine to me.

> We will be creating pending charges but they can be cleaned up in many ways by queuing jobs or daily task that cleans up old pending orders.

I didn't understand this ⬆️. Pending orders won't have any _charges_ associated with them, right?

ashkan18:
@joeyAghion sorry my bad, i meant pending Orders and not charges. There won't be any pending charges anywhere :)

ashkan18:
merging 👍

PR #61, opened 2018-07-16
Rename total cents methods and use buyer_total_cents to calculate charge amount on authorize charge for order submit

williardx:
- Renames `total_cents` to `seller_total_cents`
- Renames `subtotal_cents` to `buyer_total_cents`
- Uses `buyer_total_cents` to calculate amount of money to hold when authorizing charge on order submit


app/models/order.rb
```diff
@@ -51,12 +51,14 @@ def items_total_cents
     line_items.sum(:price_cents)
   end
 
-  def subtotal_cents
+  # Total amount (in cents) that the buyer will pay
+  def buyer_total_cents
```
ashkan18: i wonder if we could keep this one `subtotal_cents` but this is as descriptive to me 👍

ashkan18:
Thanks! this makes total sense, feel free to merge

cc: @sepans @jonallured we are going to update MP to catch up with this change, @jonallured you may need to update your current view to use the new field (`sellerTotalCents`).

app/models/order.rb
williardx: It feels more like a total than a subtotal to me since it's the total amount that the buyer will pay -- any reason why you want to keep it as `subtotal_cents`?

ashkan18: not really, just wasn't sure if thats the more industry standard word or not. this looks good to me.

williardx:
> Thanks! this makes total sense, feel free to merge

ba-dum tssss

PR #62, opened 2018-07-17
Add ArtworkVersionId

ashkan18:
# Change

We are adding artwork version id to Gravity's artwork, this way in other services, including Exchange, we don't need to store snapshot of the artwork at the time of Order creation aanymore.

- Added `artwork_version_id` to `LineItem`
- Removed `artwork_snapshot` from `LineItem`
- Removed `SetLineItemArtworkJob` and removed usage from `LineItemService`

Assumption here is, when creating an order, client would tell us the artwork version same way they mention artwork id, technically we could remove `artwork_id` here since version should be enough but i think it's ok to have both here, makes references from other places easier.


williardx:
@ashkan18 Could you rebase off master and re-run the database migrations so that the latest schema changes are accounted for? Otherwise looks good!

ashkan18:
closing this after some discussions, we will have to get version id by calling gravity.

PR #63, opened 2018-07-17
Refactor `OrderService.submit` into its own service

williardx:
Following up on https://github.com/artsy/exchange/pull/59:
- Refactors `OrderService.submit!` into a separate service `OrderSubmitService` since `OrderService.submit!` is getting pretty big.
- Removes `get_merchant_account` from `PaymentService` and puts adds it as a helper to `OrderSubmitService`.
- Refactors `PaymentService.authorize_charge` to accept source, destination amount and currency code as arguments instead of order; adds source and destination to error hash.
- Refactors `TransactionService` methods to only use fields from Stripe charge object or error hash.


ashkan18:
awesome! 🙌

PR #64, opened 2018-07-18
Remove required userId in createOrder

ashkan18:
# Change
Use user id in JWT for creating the order. This makes things easier on client side and follows existing patterns used for other mutations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,6 @@ For input set following variables:
 ```json
 {
   "input": {
-    "userId": "<your user id>",
     "partnerId": "<partner id>",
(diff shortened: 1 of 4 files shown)


ashkan18:
going to merge this and related MP's PR to get ready for todays walkthrough, i'll address any comments in a follow up PR :)

sweir27:
Nice! We talked about a follow-up where an admin could specify a `userId` in order to create an order on behalf of someone else, but that should probably wait anyways.

williardx:
🆒Thanks for updating docs, too!

PR #65, opened 2018-07-18
Require 'net/http' in application config

williardx:
`net/http` is available in dev but not in production, possibly due to how rails loads files in each environment. Adds `net/http` to application config to fix this difference.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -12,6 +12,7 @@ require 'action_view/railtie'
 require 'action_cable/engine'
 # require "sprockets/railtie"
 require 'rails/test_unit/railtie'
+require 'net/http'
 
 # Require the gems listed in Gemfile, including any gems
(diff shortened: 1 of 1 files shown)


ashkan18:
Looks 👍 to me, @yuki24 is this the recommended way?

ashkan18:
perfect!

config/application.rb
```diff
@@ -12,6 +12,7 @@
 require 'action_cable/engine'
 # require "sprockets/railtie"
 require 'rails/test_unit/railtie'
+require 'net/http'
```
yuki24: This is a good place to require stdlib :+1:

yuki24:
Also, a better place to require is the file that actually depends on the library. @williardx @ashkan18 do you still have the backtrace from the error you were seeing?

PR #66, opened 2018-07-18
Move the require to the file that actually depends on it

yuki24:
following up on https://github.com/artsy/exchange/pull/65

cc @alloy

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -12,7 +12,6 @@ require 'action_view/railtie'
 require 'action_cable/engine'
 # require "sprockets/railtie"
 require 'rails/test_unit/railtie'
-require 'net/http'
 
 # Require the gems listed in Gemfile, including any gems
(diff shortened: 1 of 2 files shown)


ashkan18:
we thought about adding it there actually but then i was worried if we start using `Net::HTTP` from some other place we need to remember adding it there too, right?

alloy:
Ace 👍 

@ashkan18 Yeah you would, but I would argue that that is a good thing.
* If you decide to only want to test parts of the codebase then you can load just the files you need and have a faster boot time.
* Once you remove all files that, in this case, depend on `Net::HTTP`, your app won’t be loading it any longer and thus is easier to maintain.
* Being able to see the dependencies of a file at a glance is helpful in debugging.

ashkan18:
sounds good to me, specially coming from Java and Python im used to requiring things hehehe 🙌

PR #67, opened 2018-07-18
Add shipping region

ashkan18:
This came up during mock-pilot, we were missing `shipping_region` on Order. Adding it in this PR.


williardx:
💯 Thanks!

PR #70, opened 2018-07-19
Fix approved state expiration timestamp issue

### LINKED ISSUE #69: Set state expires when approving order
See more comments here: https://artsy.slack.com/archives/CB110C6LE/p1531948876000051 ❤️

ashkan18:
# Problem
After approving an order, `state_expires_at` timestamp doesn't get updated properly and it becomes nil.
fixes #69 

# Cause
In the state expiry setting we didn't have approved since we didn't know whats the expiration rule.

# Solution
Set expiry for 5 days for _approved_.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -18,7 +18,8 @@ class Order < ApplicationRecord
 
   STATE_EXPIRATIONS = {
     'pending' => 2.days,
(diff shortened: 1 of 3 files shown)


ashkan18:
merging now on green :)

PR #72, opened 2018-07-19
Fetch Stripe credit card ID from Gravity on order submit

williardx:
**Before**: The supplied `credit_card_id` from `setPayment` mutation was passed directly to the Stripe charge object as the `source`.

**With these changes**: The supplied `credit_card_id` is used to look up the user's Stripe credit card ID in Gravity, which is then passed to the charge object as the `source`.

**What changed**:
- Fetches buyer's Stripe credit card ID from Gravity using supplied `credit_card_id` on order submit and uses in transaction
- Adds `customer_id` to Stripe charge object since we need to use it for saved credit cards
- Refactors `PaymentService.authorize_charge` to use keyword arguments as parameters

### DIFF
--- a/app/services/order_submit_service.rb
+++ b/app/services/order_submit_service.rb
@@ -4,9 +4,21 @@ module OrderSubmitService
     Order.transaction do
       # verify price change?
       order.submit!
+
       merchant_account = get_merchant_account(order)
       raise Errors::OrderError, 'Partner does not have merchant account' if merchant_account.nil?
-      charge = PaymentService.authorize_charge(order.credit_card_id, merchant_account[:external_id], order.buyer_total_cents, order.currency_code)
+
+      credit_card = get_credit_card(order)
+      raise Errors::OrderError, 'User does not have stored credit card' if credit_card.blank?
+
+      charge_params = {
+        source_id: credit_card[:external_id],
+        customer_id: credit_card[:customer_account][:external_id],
+        destination_id: merchant_account[:external_id],
+        amount: order.buyer_total_cents,
+        currency_code: order.currency_code
+      }
+      charge = PaymentService.authorize_charge(charge_params)
       TransactionService.create_success!(order, charge)
       order.save!
     end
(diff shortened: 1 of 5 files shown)


app/services/order_submit_service.rb
```diff
@@ -4,9 +4,21 @@ def self.submit!(order)
     Order.transaction do
       # verify price change?
       order.submit!
+
       merchant_account = get_merchant_account(order)
       raise Errors::OrderError, 'Partner does not have merchant account' if merchant_account.nil?
-      charge = PaymentService.authorize_charge(order.credit_card_id, merchant_account[:external_id], order.buyer_total_cents, order.currency_code)
+
+      credit_card = get_credit_card(order)
```
ashkan18: what do you think about moving this _preparation code_ (fetching credit card / merchant account) out of transaction block? I think in the end it doesn't really impact the logic in any way but i wonder if it makes things more obvious if we just keep internal transactional changes within the transaction block.

app/services/order_submit_service.rb
```diff
@@ -4,9 +4,21 @@ def self.submit!(order)
...
+      raise Errors::OrderError, 'User does not have stored credit card' if credit_card.blank?
```
ashkan18: do you think we should track this case as a failed attempted `transaction`? or some other way since it does sounds like a weird case. Maybe we should also log it or something.

app/services/order_submit_service.rb
```diff
@@ -4,9 +4,21 @@ def self.submit!(order)
     Order.transaction do
       # verify price change?
       order.submit!
+
       merchant_account = get_merchant_account(order)
       raise Errors::OrderError, 'Partner does not have merchant account' if merchant_account.nil?
```
ashkan18: i know this is not part of this PR but maybe we should also log this in a more significant way since this may need some internal actions on our end. Like notifying the partner, liaison or something.

app/services/order_submit_service.rb
ashkan18: totally agree, a Sentry log should be perfect for this case.

app/services/order_submit_service.rb
williardx: Good idea! I'll clean that up.

PR #73, opened 2018-07-19
Add `external_charge_id` to Order model and save on order submit

williardx:
### Before
We would authorize a charge and save the transaction on order submit but not store the charge id on the order model itself. Since an order may have multiple charges, we would have to decide which transaction ought to be captured in order to complete the order.

### With these changes
The order model now contains an `external_charge_id` field that stores the Stripe charge id of the charge created on submitting an order. This allows us to easily retrieve it later when we want to capture the charge.

### What changed
- Adds `external_charge_id` string field to `orders` table
- Saves Stripe charge id to `external_charge_id` on successful submission of an order.

### DIFF
--- a/app/services/order_submit_service.rb
+++ b/app/services/order_submit_service.rb
@@ -20,6 +20,7 @@ module OrderSubmitService
     Order.transaction do
       order.submit!
       charge = PaymentService.authorize_charge(charge_params)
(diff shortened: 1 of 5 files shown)


ashkan18:
Nice 🙌

PR #74, opened 2018-07-19
Update order code to match expected format

ashkan18:
# Change
https://www.notion.so/artsy/Define-order-number-structure-65e2d9186f214b1184113a9d5a010b3b

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -74,7 +74,7 @@ class Order < ApplicationRecord
   private
 
   def set_code
-    update!(code: "B%06d" % self.id)
+    update!(code: format('B%06d', id))
   end
 
   def update_state_timestamps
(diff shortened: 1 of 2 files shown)


ashkan18:
yes, `id` is just `order`'s sequential auto generated id and we just pad it with zeroes and `B` at the beginning.

williardx:
Looking at the discussion in Notion we talked about human-readable numbers being randomized. I don't know that a decision was ever finalized on that, but I would definitely lobby for randomized vs. sequential order numbers.

ashkan18:
oh somehow i totally missed that chat icon for that item and all discussion we had, i'd say if we are going with random numbers we might be better with random characters, padded random numbers feel little strange.

williardx:
Do they need to be padded? I.e. could we just generate a random 8-digit number in a given range?

ashkan18:
we'd get more chances and less collisions, right? 6 digit number would ignore 0000001 as an option, i just commented on that notion chat section, im leaning more towards `B<6 random non-digits>` all caps.

sepans:
Looks great 🥇 merging.

PR #76, opened 2018-07-20
Capture charge on order approval

williardx:
- Adds `PaymentService.capture_charge`, which retrieves a Stripe charge by id and captures it.
- Captures charge on order approval and records the successful (or failed) transaction.
- Adds Boolean `captured` column to `orders` to differentiate between captured and uncaptured charges.
- Adds `stripe-ruby-mock` to dependencies and refactors some tests to use mocked Stripe calls.

### Code Review Changes

- [x] Change `captured` Boolean to `type` to describe what kind of transaction it is
- [x] Create shared context for Stripe tests
- [x] Add logging statements for failed charges

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ group :development, :test do
   gem 'graphlient'
   gem 'rspec-rails'
   gem 'rubocop'
+  gem 'stripe-ruby-mock', '~> 2.5.4', :require => 'stripe_mock'
 end
 
 group :development do
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,6 @@ group :development, :test do
   gem 'graphlient'
   gem 'rspec-rails'
   gem 'rubocop'
-  gem 'stripe-ruby-mock', '~> 2.5.4', require: 'stripe_mock'
 end
 
 group :development do
@@ -28,4 +27,5 @@ end
 
 group :test do
   gem 'fabrication'
+  gem 'stripe-ruby-mock', '~> 2.5.4', require: 'stripe_mock'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,7 @@ group :development, :test do
   gem 'graphlient'
   gem 'rspec-rails'
   gem 'rubocop'
-  gem 'stripe-ruby-mock', '~> 2.5.4', :require => 'stripe_mock'
+  gem 'stripe-ruby-mock', '~> 2.5.4', require: 'stripe_mock'
 end
 
 group :development do
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -35,6 +35,7 @@ module OrderService
     order
   rescue Errors::PaymentError => e
     TransactionService.create_failure!(order, e.body)
+    Rails.logger.error("Could not approve order #{order.id}: #{e.message}")
     raise e
   end
 
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -28,17 +28,20 @@ module OrderService
   def self.approve!(order)
     Order.transaction do
       order.approve!
+      charge = PaymentService.capture_charge(order.external_charge_id)
+      TransactionService.create_success!(order, charge)
       order.save!
-      # TODO: process the charge by calling gravity with current credit_card_id and price
     end
     order
+  rescue Errors::PaymentError => e
+    TransactionService.create_failure!(order, e.body)
+    raise e
   end
 
   def self.finalize!(order)
     Order.transaction do
       order.finalize!
       order.save!
-      # TODO: process the charge by calling gravity with current credit_card_id and price
     end
     order
   end
--- a/app/services/order_submit_service.rb
+++ b/app/services/order_submit_service.rb
@@ -27,6 +27,7 @@ module OrderSubmitService
     order
   rescue Errors::PaymentError => e
     TransactionService.create_failure!(order, e.body)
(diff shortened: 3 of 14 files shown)


app/services/payment_service.rb
```diff
@@ -21,4 +21,17 @@ def self.authorize_charge(source_id:, customer_id:, destination_id:, amount:, cu
...
+    charge = Stripe::Charge.retrieve(charge_id)
+    charge.capture
+  rescue Stripe::StripeError => e
+    body = e.json_body[:error]
+    failed_charge = {
+      id: charge_id,
+      failure_code: body[:code],
+      failure_message: body[:message]
+    }
+    raise Errors::PaymentError.new(e.message, failed_charge)
```
ashkan18: i like our pattern here 👍 of raising our own errors

app/services/transaction_service.rb
```diff
@@ -19,6 +20,7 @@ def self.create_success!(order, charge)
       amount_cents: charge.amount,
       failure_code: charge.failure_code,
       failure_message: charge.failure_message,
+      captured: charge.captured,
```
ashkan18: do you think we need a transaction type or something? with current `captured` flag we can tell if this was a successful `hold` or `capture` but when it comes to failure it's hard to find _failed_ `capture` transactions, right? 

i wonder if it makes sense to add `type` or `action` to transaction and then have `capture` and `hold` being different types. This way a failed capture would have type `capture` and `success` being `false`. With current schema we could do the same thing too though if we set `capture` to `true` here https://github.com/artsy/exchange/pull/76/files#diff-455653e32502e66fef7ea14f1f772dd0R10 for capture transactions.
What do you think?

spec/services/payment_service_spec.rb
```diff
@@ -1,27 +1,86 @@
 require 'rails_helper'
 require 'support/gravity_helper'
+require 'stripe_mock'
```
ashkan18: 👍

spec/services/payment_service_spec.rb
```diff
@@ -1,27 +1,86 @@
...
 
 describe PaymentService, type: :services do
-  let(:source_id) { 'cc-1' }
+  let(:stripe_helper) { StripeMock.create_test_helper }
+  before { StripeMock.start }
+  after { StripeMock.stop }
```
ashkan18: there are two alternatives i can think of for these, one is to add them as a shared context and whenever we need stripe setup we could do
```ruby
include_context 'with Stripe mock'
```

A more involved alternative is to also use types https://relishapp.com/rspec/rspec-core/docs/hooks/filters and basically say if type is `stripe` or something start the server.

maybe shared context works better in our case, what do you think?

app/services/order_service.rb
```diff
@@ -37,17 +37,20 @@ def self.set_shipping!(order, attributes)
   def self.approve!(order)
     Order.transaction do
       order.approve!
+      charge = PaymentService.capture_charge(order.external_charge_id)
+      TransactionService.create_success!(order, charge)
       order.save!
-      # TODO: process the charge by calling gravity with current credit_card_id and price
     end
     order
+  rescue Errors::PaymentError => e
```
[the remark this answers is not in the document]
williardx: Agreed - if there's a simple/easy way to log it right now I'm down to add that in.

app/services/transaction_service.rb
williardx: Oh yeah, I like that! That's really what I'm trying to distinguish between, here, but you're right that having every failed transaction classified as non-captured would make it hard to find failed captures. I'll update that.

app/services/order_service.rb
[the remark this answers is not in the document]
ashkan18: i think this is good for now, eventually i was thinking of posting a rabbitMQ event and maybe post slack message to some channel based on this, we are kind of doing that currently for some credit failures in invoicing project, look at #fraud slack channel.

Gemfile
```diff
@@ -19,6 +19,7 @@ group :development, :test do
   gem 'graphlient'
   gem 'rspec-rails'
   gem 'rubocop'
+  gem 'stripe-ruby-mock', '~> 2.5.4', require: 'stripe_mock'
```
ashkan18: sorry for catching this late, it may make more sense to move this under `group :test:` only

app/services/order_submit_service.rb
```diff
@@ -27,6 +27,7 @@ def self.submit!(order)
     order
   rescue Errors::PaymentError => e
     TransactionService.create_failure!(order, e.body)
+    Rails.logger.error("Could not submit order #{order.id}: #{e.message}")
```
ashkan18: 👍

app/services/payment_service.rb
```diff
@@ -9,6 +9,8 @@ def self.authorize_charge(source_id:, customer_id:, destination_id:, amount:, cu
       destination: destination_id,
       capture: false
     )
+    charge.transaction_type = 'hold'
```
ashkan18: we may want to define possible types in `Transaction` model, something like
```
TYPES = [
  HOLD = 'hold',
  CAPTURE = 'capture'
]
```
and here say 
```
charge.transaction_type = Transaction::HOLD
```

Gemfile
williardx: Good catch!

app/services/payment_service.rb
williardx: Oh yeah, that's much better

ashkan18:
🕺

PR #77, opened 2018-07-20
Create order with artwork

### LINKED ISSUE #75: CreateOrder weirdness
# Current state
Current `createOrder` mutation requires `partnerId`, a list of line items in which we require an artwork.

# Weirdness points
We set `partnerId` on an `Order` but it's actually related directly to the artworks in the order line items. With current mutation it's technically possible for someone to call this mutation, set `partnerId` to partner A but add artworks of partner B to the line items.
We talked about not requiring `partnerId` in the mutation and from exchange we fetch the partner Id , this is also as weird as the current solution since exchange basically has to find the first artwork in the line items (ignoring all other artworks) and from that artwork, set `Orders`'s `partnerId` which seems super weird.

This is all because of the fact that an order can have multiple line items (potentially multiple artworks) but belongs to specific partner. So we need to make sure all line items of an order belong to the partner of that order.

# Possible solution A

### Provide `createOrderFromArtwork` mutation
```graphql
mutation createOrderFromArtwork(input: {artworkId: String!, editionsetId: String, quantity: Int})
```
This way we start from line item and we make sure this order can only have one line item (matching our current front end rules.

Later when we started supporting multiple artwork purchases from a partner we can provide 
```graphql
mutation createOrder(input: {partnerId: String!})
```
and 
```graphql
mutation addLineItem(input: {artworkId: String!, editionsetId: String})
```
and in ☝️we make sure artworkId being added to the order belongs to the order's partner.

# Possible Solution B
Enforce 1-1 relation between order and line items, basically an order can have only one line item. This would solve our problem of assuming partner id from line items but the issue here is this modeling won't be realistic over time.

# Proposed solution
Go with option **A** as more flexible and closer to reality option.

ashkan18:
# Problem
Current `createOrder` mutation expects clients to send `partnerId`, `priceCents`, `currencyCode` as part of line items array. These are values that Exchange could figure out based on `artworkId`. 

Fixes #75 

# Solution
Make create order easier by removing old `createOrder` and providing `createOrderWithArtwork` which gets
- `artworkId` required
- `editionSetId` optional

We then fetch artwork data from Gravity and based on that we create `order` with proper `partnerId`, if `editionSetId` was set in mutation we get price based on that edition set.


app/graphql/mutations/create_order_with_artwork.rb
```diff
@@ -0,0 +1,17 @@
+class Mutations::CreateOrderWithArtwork < Mutations::BaseMutation
+  argument :artwork_id, String, 'Artwork Id', required: true
+  argument :edition_set_id, String, 'EditionSet Id', required: false
+  argument :quantity, Integer, 'Number of items in the line item', required: false
+
+  field :order, Types::OrderType, null: true
+  field :errors, [String], null: false
+
+  def resolve(artwork_id:, edition_set_id: nil, quantity: 1)
```
williardx: I'm a little fuzzy on our use of `quantity` - do we expect to have orders with more than one item? Would buying two prints from an editioned set required two separate orders?

app/services/line_item_service.rb
```diff
@@ -1,19 +0,0 @@
-module LineItemService
```
williardx: What made you decide to get rid of this and call `order.line_items.create!` directly?

app/graphql/mutations/create_order_with_artwork.rb
ashkan18: yes, this is just to allow for buying multiples, apparently that has happened in legacy ecommerce

app/services/line_item_service.rb
ashkan18: ended up creating a separate `CreateOrderService` which has some of the methods used to be here, also this module used to copy a snapshot of the artwork and with artwork version we don't need that anymore, we can add it back if there was a need for more complicated logic for adding line items.

this reminds me though that i need to drop `artwork_snapshot` from `LineItem`. Going to do that in this PR

ashkan18:
nice @orta ☝️

app/graphql/mutations/create_order_with_artwork.rb
```diff
@@ -0,0 +1,17 @@
+class Mutations::CreateOrderWithArtwork < Mutations::BaseMutation
```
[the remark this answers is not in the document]
ashkan18: yes i can't decide between `from` and `with`, if `with` is not confusing i'd say lets go _with_ it :)

ashkan18:
thanks all, merging 👍

PR #78, opened 2018-07-23
allow trusted users to perform graphql queries

mbilokonsky:
short and sweet, with @ashkan18's help

### DIFF
--- a/app/controllers/concerns/auth_handler.rb
+++ b/app/controllers/concerns/auth_handler.rb
@@ -4,7 +4,7 @@ module AuthHandler
       protected
 
       def authenticate_request!
-        unless user_id_in_token?
+        unless user_id_in_token? || !auth_token.key?('sub')
           render json: { errors: ['Not Authenticated'] }, status: :unauthorized
           return
         end
--- a/app/controllers/concerns/auth_handler.rb
+++ b/app/controllers/concerns/auth_handler.rb
@@ -4,10 +4,6 @@ module AuthHandler
       protected
 
       def authenticate_request!
-        unless user_id_in_token? || !auth_token.key?('sub')
-          render json: { errors: ['Not Authenticated'] }, status: :unauthorized
-          return
-        end
         @current_user = auth_token.with_indifferent_access.merge(id: auth_token[:sub])
       rescue JWT::DecodeError # includes verification error too
         render json: { errors: ['Not Authenticated'] }, status: :unauthorized
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -17,7 +17,7 @@ class Types::QueryType < Types::BaseObject
 
   def order(id:)
     order = Order.find(id)
-    raise GraphQL::ExecutionError, 'Not permitted' unless trusted? || user_permitted?(order.user_id) || partner_permitted?(order.partner_id)
+    raise GraphQL::ExecutionError, 'Not permitted' unless user_permitted?(order.user_id) || partner_permitted?(order.partner_id)
     order
   end
 
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -40,7 +40,7 @@ class Types::QueryType < Types::BaseObject
 
   def validate_orders_params!(params)
     raise GraphQL::ExecutionError, 'requires one of userId or partnerId' unless params[:user_id].present? || params[:partner_id].present?
-    raise GraphQL::ExecutionError, 'Not permitted' if params[:user_id] && params[:user_id] != context[:current_user][:id]
+    raise GraphQL::ExecutionError, 'Not permitted' if params[:user_id] && params[:user_id] != context[:current_user][:id] || !context[:current_user][:roles].include?('trusted')
     raise GraphQL::ExecutionError, 'Not permitted' if params[:partner_id] && !context[:current_user][:partner_ids].include?(params[:partner_id])
   end
 end
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -38,9 +38,25 @@ class Types::QueryType < Types::BaseObject
 
   private
 
+  def user_logged_in?(user_id)
+    user_id == context[:current_user][:id]
+  end
+
+  def user_trusted?
+    context[:current_user][:roles].include?('trusted')
+  end
+
+  def user_permitted?(id) 
+    user_trusted? || user_logged_in?(id)
+  end
+
+  def partner_permitted?(id)
+    context[:current_user][:partner_ids].include?(id)
+  end
+
   def validate_orders_params!(params)
     raise GraphQL::ExecutionError, 'requires one of userId or partnerId' unless params[:user_id].present? || params[:partner_id].present?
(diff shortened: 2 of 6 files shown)


ashkan18:
👍 quick and easy, my only comment is lets add some spec in https://github.com/artsy/exchange/blob/master/spec/controllers/api/requests/orders_query_request_spec.rb

app/graphql/types/query_type.rb
```diff
@@ -38,9 +38,25 @@ def orders(params)
...
+
+  def partner_permitted?(id)
+    context[:current_user][:partner_ids].include?(id)
+  end
+
   def validate_orders_params!(params)
     raise GraphQL::ExecutionError, 'requires one of userId or partnerId' unless params[:user_id].present? || params[:partner_id].present?
-    raise GraphQL::ExecutionError, 'Not permitted' if params[:user_id] && params[:user_id] != context[:current_user][:id]
-    raise GraphQL::ExecutionError, 'Not permitted' if params[:partner_id] && !context[:current_user][:partner_ids].include?(params[:partner_id])
+    raise GraphQL::ExecutionError, 'Not permitted' if params[:user_id] && !user_permitted?(params[:user_id])
```
ashkan18: to better reflect reality i _think_ its better to have something like
```ruby
raise GraphQL::ExecutionError, 'requires one of userId or partnerId' unless params[:user_id].present? || params[:partner_id].present?
return if trusted?(context[:current_user])
raise GraphQL::ExecutionError, 'Not permitted' if params[:user_id] && !user_permitted?(params[:user_id])
raise GraphQL::ExecutionError, 'Not permitted' if params[:partner_id] && !partner_permitted?(params[:partner_id])
```
and then `user_permitted?` ends up being basically
```ruby
user_id == context[:current_user][:id]
```
this way we can also remove `user_logged_in?` method, in this case if a user gets to this validation, they are technically _logged in_ but maybe not authorized to make this request. in case of a trusted app, they are authorized to fetch orders by user and also partners.

spec/controllers/api/requests/order_query_request_spec.rb
```diff
@@ -54,6 +54,17 @@
       end
     end
 
+    context 'trusted account accessing order' do
+      let(:jwt_roles) { 'trusted' }
```
ashkan18: this test may not be accurate since with this setup JWT user is also correct user and also its trusted app, i _think_ this test would pass even if you remove the role. We can do few things here, either set `jwt_user_id` to `nil` (this will be more realistic since trusted app tokens don't have a user id. Or we can update the `id` we are trying to fetch to someone else's order.

It may be useful to also make sure we got the correct order in the response along with making sure no errors are raised.

I'd also suggest adding spec to make sure not trusted apps can't access this order.

spec/controllers/api/requests/order_query_request_spec.rb
```diff
@@ -54,6 +54,17 @@
...
+      let(:order_to_use) { Fabricate(:order, partner_id: second_partner_id, user_id: user_id) }
```
ashkan18: do we need this order or we could use existing orders defined https://github.com/artsy/exchange/pull/78/files#diff-7263081cce0e1c7060fdca29ad4d9ad8R10

spec/controllers/api/requests/orders_query_request_spec.rb
```diff
@@ -80,5 +74,22 @@
         expect(results.data.orders.edges.map(&:node).map(&:id).map(&:to_i)).to eq([user1_order2.id, user1_order1.id])
       end
     end
+
+    context 'user query where user is not in jwt' do
+      it 'usually returns permission error' do
```
ashkan18: if we are expecting to raise error i would change `it` description to `it 'raises permission error'`

app/graphql/types/query_type.rb
```diff
@@ -38,9 +38,33 @@ def orders(params)
 
   private
 
+  def trusted?
+    context[:current_user][:roles].include?('trusted')
+  end
+
+  def user_permitted?(id)
+    trusted? || id == context[:current_user][:id]
```
ashkan18: do we still need `trusted?` check here since we are doing it as separate check in line 20? also this is super super minor, but might read better if we have `user_permitted?` to get an Order instead of `id` so the caller would do something like `user_permitted?(order)`. what do you think?

app/graphql/types/query_type.rb
```diff
@@ -38,9 +38,33 @@ def orders(params)
...
+  end
+
+  def partner_permitted?(id)
+    trusted? || context[:current_user][:partner_ids].include?(id)
```
ashkan18: do we need `trusted?` here?

app/graphql/types/query_type.rb
mbilokonsky: Ah, I think the confusion is because I left `trusted? ||` in line 20. I moved the `trusted?` check down into those methods as part of a cleanup of the orders validation logic, but the single-order validation logic is its own path.

Maybe the real problem is that we should unify validation into a single pipeline?

app/graphql/types/query_type.rb
mbilokonsky: We do from the standpoint of the refactor of `validate_orders_params!`, because it no longer does the check up front. (because we were getting rubocop complaining about cyclomatic complexity).

app/graphql/types/query_type.rb
```diff
@@ -38,9 +38,33 @@ def orders(params)
...
+
+  def validate_partner(id)
+    raise GraphQL::ExecutionError, 'Not permitted' unless partner_permitted?(id)
+  end
+
   def validate_orders_params!(params)
-    raise GraphQL::ExecutionError, 'requires one of userId or partnerId' unless params[:user_id].present? || params[:partner_id].present?
-    raise GraphQL::ExecutionError, 'Not permitted' if params[:user_id] && params[:user_id] != context[:current_user][:id]
-    raise GraphQL::ExecutionError, 'Not permitted' if params[:partner_id] && !context[:current_user][:partner_ids].include?(params[:partner_id])
+    if params[:user_id].present?
```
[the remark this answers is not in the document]
mbilokonsky: I considered that but if there is no user or partner present then the query itself is bogus, right? We still want to ensure that either user or partner exist, so putting the `trusted?` check before we do that feels wrong to me.

So the idea is that we fork once here, based on whether it's a user request or a partner request, and then each enters into its own validation pathway. I'm okay with repeating `trusted?` at the top of each path.

app/controllers/concerns/auth_handler.rb
```diff
@@ -4,7 +4,7 @@ def self.included(clazz)
       protected
 
       def authenticate_request!
-        unless user_id_in_token?
+        unless user_id_in_token? || !auth_token.key?('sub')
```
ashkan18: haha this is really hard to read! finally i managed to read it and basically it seems we can remove this check altogether, basically here we are eventually saying if there is user id (`sub`) in the token of there is no `sub` in the token. I think we can just remove this.

mbilokonsky:
pinging @ashkan18 for final review and merge

ashkan18:
Nice! thanks for your patience this looks great 🙌

PR #79, opened 2018-07-23
Add shipping calculation

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PLATFORM-665
We want to calculate shipping when we set shipping on an order.

# Changes
- Added `ShippingService` which provides `calculate_shipping` for a line item
- Updated `OrderService` to call ☝️ for every line item in the order and calculate total shipping and set it on the order
- For now added `DISABLE_SHIPPING_CALCULATION` to avoid dependency to https://github.com/artsy/gravity/pull/11826 and have this be shippable
- For now calculating tax there with hardcoded `100_00` 💵 
- Used `Rails.cache` to cache artwork results for 10 mins, im not much sure about 10 mins part but it may be ok if we force refresh at submission time.

# Depends kind of on
https://github.com/artsy/gravity/pull/11826
this can be merged anytime before ☝️ but we need to set `DISABLE_SHIPPING_CALCULATION` to `true` in ENV. Once ☝️ is merged we can remove the env setting and also remove the feature in `dev_features.yml`.

# Migration 
- [ ] setup memcached and set `MEMCACHE_HOST` env variable


app/services/artwork_service.rb
```diff
@@ -1,8 +1,16 @@
...
+  ARTWORK_CACHE_EXPIRATION_IN_MINUTES = 10
   def self.find(artwork_id)
-    Adapters::GravityV1.request("/artwork/#{artwork_id}?include_deleted=true")
+    find!(artwork_id)
+  rescue Adapters::GravityError
+    nil
+  end
+
+  def self.find!(artwork_id)
+    Rails.cache.fetch("gravity_#{artwork_id}", expires_in: ARTWORK_CACHE_EXPIRATION_IN_MINUTES.minutes) do
```
williardx: Will this eventually point at an artwork version?

.rubocop.yml
```diff
@@ -32,3 +32,6 @@ Style/FrozenStringLiteralComment:
 
 Rails/NotNullColumn:
   Enabled: false
+
+Style/AsciiComments:
+  Enabled: false
```
williardx: 💯 💯

app/services/artwork_service.rb
ashkan18: yes, although we should think about how we want to handle caching, ideally we always point to latest version.

starsirius: Do we want the key to be scoped under artwork and differentiate it from other types of object?

app/services/order_service.rb
```diff
@@ -9,16 +9,23 @@ def self.set_payment!(order, credit_card_id:)
...
-        attributes.slice(
-          :shipping_address_line1,
-          :shipping_address_line2,
-          :shipping_city,
-          :shipping_region,
-          :shipping_country,
-          :shipping_postal_code,
-          :fulfillment_type
+        attrs.merge(
+          attributes.slice(
```
starsirius: #minor. The merge order here implies we will be able to override `shipping_total_cents` and `tax_total_cents` with `attributes` passed in, regardless of the calculation, right? I wonder if we want to disallow that and always respect the calculation. If we want to manually set them, we can still update the order directly.

app/services/shipping_service.rb
```diff
@@ -0,0 +1,28 @@
+module ShippingService
+  def self.calculate_shipping(line_item, fulfillment_type:, shipping_country:)
+    # TODO: 🚨 remove this feature flag, only needed during development 🚨
+    return 0 if Rails.application.config_for(:dev_features)['disable_shipping_calculation']
+    artwork = ArtworkService.find(line_item.artwork_id)
+    raise Errors::OrderError, 'Cannot caclulate shipping, unknown artwork' unless artwork
```
starsirius: "calculate"

starsirius: In case of exception cases like 404, is [this](https://github.com/artsy/exchange/blob/09ca94e6e3927f87ca53031037f9afea311b5d5f/lib/adapters/gravity_v1.rb#L14) going to return the error message or raise an error?

app/services/shipping_service.rb
```diff
@@ -0,0 +1,28 @@
...
+      calculate_international(artwork)
+    end
+  end
+
+  def self.domestic?(artwork_location, shipping_country)
+    artwork_location[:country].casecmp(shipping_country).zero?
+  end
+
+  def self.calculate_domestic(artwork)
+    artwork[:domestic_shipping_fee_cents]
```
starsirius: In https://github.com/artsy/gravity/pull/11826, I'm going to set the default value of `domestic_shipping_fee_cents` and `international_shipping_fee_cents` to `nil`. It should be an edge case, but we may want to be defensive here.

app/services/artwork_service.rb
ashkan18: good point, yes, will add artwork to the key im sure later we need to cache partners.

app/services/order_service.rb
ashkan18: good point, i think this is fine though since we `slice` the `attributes` so if they send `shipping_total_cents` and `tax_total_cents` we'd still be using out own calculated values since we don't get them from attributes, right?

app/services/shipping_service.rb
ashkan18: i have a spec for it here 
https://github.com/artsy/exchange/pull/79/files#diff-38235e15f6a0663d04941504b54ba1baR43

since we are using `find` and not `find!` we'll get `nil` in those cases and we'll raise `OrderError` and eventually graphql error. does it make sense?

app/services/shipping_service.rb
ashkan18: good point! adding a spec to confirm it.

app/services/order_service.rb
starsirius: Yeah you are right. I missed that! 👍

app/services/shipping_service.rb
starsirius: Yeah, my question was basically if `Adapters::GravityV1.request` would raise exceptions for cases like 404 or we have to inspect the [response status](https://github.com/artsy/exchange/blob/09ca94e6e3927f87ca53031037f9afea311b5d5f/lib/adapters/gravity_v1.rb#L13).

ashkan18: oh sorry misread that, i was reading about `Net::HTTP` and i think it raises all sort of exception for each code, some people recommended a catch all with `StandardError` which we do here as scary as it sounds 
https://github.com/artsy/exchange/blob/master/lib/adapters/gravity_v1.rb#L15
so 404 ends up being `GravityError`

we do this in Impuse
https://github.com/artsy/impulse/blob/master/lib/adapters/radiation.rb#L8

i'll try locally tomorrow

starsirius: Yeah I wasn't sure and the doc wasn't clear to me. Looks like Impulse did [check the response](https://github.com/artsy/impulse/blob/71da5ac1a3a02569480762da5592fd93074a15c3/lib/adapters/radiation.rb#L35) and raise properly exceptions ourselves tho, but I could be wrong.

app/services/artwork_service.rb
```diff
@@ -1,8 +1,16 @@
 module ArtworkService
   PRIVATE_ARTWORK_FIELDS = %i[price_changed_by flags inventory publish_change_requested_by location deleted_by gene_count partner_gene_count dates external_id display_price_range confidential_notes created_at inventory_id published_changed_at publish_change_requested_at secondary_market deleted dimensions_string].freeze
-
+  ARTWORK_CACHE_EXPIRATION_IN_MINUTES = 10
```
izakp: something for `_config.rb`?  I.e. ENV override?

app/services/artwork_service.rb
```diff
@@ -1,8 +1,16 @@
...
   def self.find(artwork_id)
-    Adapters::GravityV1.request("/artwork/#{artwork_id}?include_deleted=true")
+    find!(artwork_id)
+  rescue Adapters::GravityError
+    nil
+  end
+
+  def self.find!(artwork_id)
```
izakp: #minor Is it idiomatic to use a `!`-version of a method name when there's not an in-place update being performed on some object?  I think these two methods could be collapsed (if not used in specs somewhere)

app/services/order_service.rb
```diff
@@ -9,16 +9,23 @@ def self.set_payment!(order, credit_card_id:)
 
   def self.set_shipping!(order, attributes)
     raise Errors::OrderError, 'Cannot set shipping info on non-pending orders' unless order.state == Order::PENDING
+
     Order.transaction do
+      attrs = {
+        shipping_total_cents: order.line_items.map { |li| ShippingService.calculate_shipping(li, attributes.slice(:shipping_country, :fulfillment_type)) }.sum,
+        tax_total_cents: 100_00 # TODO: 🚨 replace this with real tax calculation 🚨
```
izakp: `grep -r 🚨.` 😂

app/services/shipping_service.rb
```diff
@@ -0,0 +1,28 @@
...
+      calculate_international(artwork)
+    end
+  end
+
+  def self.domestic?(artwork_location, shipping_country)
+    artwork_location[:country].casecmp(shipping_country).zero?
+  end
+
+  def self.calculate_domestic(artwork)
+    artwork[:domestic_shipping_fee_cents] || 0
```
izakp: Free shipping if it returns `0`?

spec/services/shipping_service_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      expect(ShippingService.domestic?(artwork_location, 'IR')).to be false
+    end
+  end
+
+  describe '#calculate_domestic' do
+    it 'returns domestic shipping cost' do
+      expect(ShippingService.calculate_domestic(artwork)).to eq 100_00
+    end
+
+    context 'with nil domestic shipping' do
```
izakp: As above - curious what this edge case means for the end user

spec/lib/adapters/gravity_v1_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+require 'webmock/rspec'
+
+describe Adapters::GravityV1 do
+  it 'raises Gravity Error when get 404' do
+    stub_request(:get, /artwork/).to_return(status: 404)
```
starsirius: I think this raises an exception because there is no body returned and [`JSON.parse`](https://github.com/artsy/exchange/blob/09ca94e6e3927f87ca53031037f9afea311b5d5f/lib/adapters/gravity_v1.rb#L14) raised an exception trying to parse it. I wonder if we simulate Gravity response like below, it will still raise exceptions:

```ruby
stub_request(:get, /artwork/).to_return(status: 404, body: {error: 'Artwork Not Found'}.to_json)
```

app/services/artwork_service.rb
ashkan18: i _think_ `!` also means it raises error, im going to do a follow up PR to clean this up along with adding cache flag you mentioned in above comment 👍

app/services/artwork_service.rb
ashkan18: good point! will add this in a follow up PR need to merge this to unblock other PRs.

spec/services/shipping_service_spec.rb
ashkan18: i think that means they haven't set the shipping yet good point though, i wonder if we should error on that case @starsirius ?

going to merge this PR and will address changes in follow up PR.

spec/lib/adapters/gravity_v1_spec.rb
ashkan18: good catch! switched to Faraday 🕺

PR #80, opened 2018-07-24
Post Events

ashkan18:
# Problem
Post events to RabbitMQ for some of orders actions, currently `submit` and `approve`.

# Migration
Setup RabbitMQ connection on staging and prod.
- [x] staging
- [x] production


app/events/order_event.rb
```diff
@@ -0,0 +1,36 @@
+class OrderEvent < Events::BaseEvent
+  TOPIC = 'BNMO'.freeze
+
+  def self.post(order, action, user_id)
+    event = OrderEvent.new(user: user_id, action: action, model: order)
```
williardx: Why does this create a new `OrderEvent` from within the class instead of using `self`?

spec/jobs/set_line_item_artwork_job_spec.rb
```diff
@@ -1,5 +1,5 @@
 require 'rails_helper'
```
williardx: Filename should be for `PostNotificationJob`

app/jobs/post_notification_job.rb
```diff
@@ -0,0 +1,8 @@
+class PostNotificationJob < ApplicationJob
```
bhoggard: Is this our standard pattern now? A delayed post to RabbitMQ? Is that because we think RabbitMQ is less reliable than Redis?

ashkan18: yes, i think we do this everywhere, there are 2 benefits:
- Work is happened in background. I can imagine this is not too big of a difference since posting to Redis and RabbitMQ are the same thing in a way, but imagine something in creating event going wrong, we still want to return success, request-wise.
-  We get retries in case of failures. While it sounds odd, this has happened few times already specially when we upgraded bunny and there were connection issues and etc.

I agree from higher level this looks strange.

spec/jobs/set_line_item_artwork_job_spec.rb
ashkan18: good catch!

PR #82, opened 2018-07-25
Fix price listed

ashkan18:
# Problem
We used `price` fields coming from Gravity to store artwork/editionset price information. During test we noticed price coming in is:
- string
- in dollars
- with actual currency symbol `$500`
We expected it to be integer and in cents.

# Solution
in https://github.com/artsy/gravity/pull/11826 we started exposing `price_listed` field which will be the dollar amount, in this PR we switched to use that.

# Possible follow up
ideally we expose the price in cents from Gravity and we don't have to [duplicate](https://github.com/artsy/gravity/blob/65e398e3648d61175e7a8f4403a2d379b5aa2107/app/models/util/for_sale.rb#L221) the logic of converting dollars to cents here in Exchange.


app/services/create_order_service.rb
```diff
@@ -26,13 +26,21 @@ def self.artwork_price(external_artwork, edition_set_id: nil)
...
 
   def self.find_edition_set(external_artwork, edition_set_id)
     external_artwork[:edition_sets].find { |e| e[:id] == edition_set_id }
   end
+
+  # TODO: 🚨 update gravity to expose amount in cents and remove this duplicate logic
+  # https://github.com/artsy/gravity/blob/65e398e3648d61175e7a8f4403a2d379b5aa2107/app/models/util/for_sale.rb#L221
+  def self.price_in_cents(price_in_dollars, currency)
+    raise Errors::OrderError 'No price found' unless price_in_dollars&.positive?
+    raise Errors::OrderError 'Invalid currency' if currency.present? && currency != 'USD' # TODO: handle other currencies
```
williardx: Would be good to do conversions to uppercase or lowercase strings, here, unless we can always expect Gravity to give back uppercase.

app/services/create_order_service.rb
```diff
@@ -26,13 +26,21 @@ def self.artwork_price(external_artwork, edition_set_id: nil)
     if edition_set_id
       edition_set = find_edition_set(external_artwork, edition_set_id)
       raise Errors::OrderError, 'Unknown edition set.' unless edition_set
-      edition_set[:price]
+      price_in_cents(edition_set[:price_listed], external_artwork[:currency])
```
starsirius: This is probably fine for the pilot, but each edition set has its own currency, and currently CMS supports supplying it.

screen shot 2018-07-26 at 11 55 23 am

starsirius: Also, is it `:currency` or [`:price_currency`](https://github.com/artsy/gravity/blob/dac17c695d10096c47073ccaf97b5596357ea0ac/app/models/domain/artwork.rb#L974)?

app/services/create_order_service.rb
starsirius: Do we also want to raise errors when currency is not present?

williardx: `'usd'` is also a good candidate for a global constant that we could use across Exchange

app/services/create_order_service.rb
ashkan18: good point and good point 👍

app/services/create_order_service.rb
ashkan18: good point! going to add `SUPPORTED_CURRENCIES` somewhere

ashkan18:
Addressed code review comments, going to merge to try it on staging.

app/models/order.rb
```diff
@@ -1,5 +1,6 @@
 class Order < ApplicationRecord
   has_paper_trail
+  SUPPORTED_CURRENCIES = %w[usd].freeze
```
starsirius: @ashkan18 Gravity uses [all caps](https://github.com/artsy/gravity/blob/685b7abe04ce01ea7b33792c8bc599174d23abca/app/models/util/money_extensions.rb), such as "USD", and I think it's the ISO code format, referenced in places like [`MoneyExtension`](https://github.com/artsy/gravity/blob/685b7abe04ce01ea7b33792c8bc599174d23abca/app/models/util/money_extensions.rb#L15). We may consider keeping them consistent.

PR #83, opened 2018-07-25
Add more careful error handling for fetching credit card and merchant account

williardx:
- Modifies `get_merchant_account` and `get_credit_card` to rescue `GravityError`s. If the error is a 404 raises a more informative error message.
- Checks that credit cards are not deactivated
- Moves merchant account / credit card data validation to `get_merchant_account` and `validate_credit_card` methods
- Adds `status` to `GravityError`

### DIFF
--- a/lib/adapters/gravity_v1.rb
+++ b/lib/adapters/gravity_v1.rb
@@ -1,18 +1,14 @@
 module Adapters
-  class GravityError < StandardError
-    attr_reader :message, :status
-    def initialize(message, status = nil)
-      @message = message
-      @status = status
-    end
-  end
+  class GravityError < StandardError; end
+  class GravityNotFoundError < GravityError; end
   class GravityV1
     def self.request(url)
       url = "#{Rails.application.config_for(:gravity)['api_v1_root']}#{url}"
       response = Faraday.get(url, {}, headers)
-      raise GravityError.new("couldn't perform request! Response was #{response.status}.", response.status) unless response.success?
+      raise GravityNotFoundError if response.status == 404
+      raise GravityError, "couldn't perform request! Response was #{response.status}." unless response.success?
       JSON.parse(response.body, symbolize_names: true)
-    rescue GravityError => e
+    rescue GravityNotFoundError => e
       raise e
     rescue StandardError => e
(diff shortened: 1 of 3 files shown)


lib/adapters/gravity_v1.rb
```diff
@@ -1,12 +1,19 @@
...
+    end
   end
   class GravityV1
     def self.request(url)
       url = "#{Rails.application.config_for(:gravity)['api_v1_root']}#{url}"
       response = Faraday.get(url, {}, headers)
-      raise GravityError, "couldn't perform request! Response was #{response.status}." unless response.success?
+      raise GravityError.new("couldn't perform request! Response was #{response.status}.", response.status) unless response.success?
       JSON.parse(response.body, symbolize_names: true)
+    rescue GravityError => e
```
williardx: @ashkan18 The code here raises a `GravityError` on an unsuccessful request which then gets rescued because it's a `StandardError`. Raising a `StandardError`, though, in turn wipes out the status that gets set. I don't want to assume that every `StandardError` will have a status, so I want to let the `GravityError` that gets raised from an unsuccessful request go forward without being rescued. Do you have any better ideas of how to do this? It feels a little silly to me to raise a `GravityError` and then immediately rescue it, only to raise it again unmodified...

lib/adapters/gravity_v1.rb
```diff
@@ -1,12 +1,19 @@
 module Adapters
   class GravityError < StandardError
+    attr_reader :message, :status
+    def initialize(message, status = nil)
+      @message = message
+      @status = status
```
ashkan18: interesting, one thing about this is, the caller of this method are now have to deal with http status codes (currently only 404). If you think from caller's perspective they care if card wasn't found or if there was timeout or other errors, I'd suggest we could add a separate exception only for not found so basically
```ruby
class GravityError < StandardError
end
class GravityNotFound < StandardError
end
```
then internally in this class we could do
```ruby
raise GravityNotFound if response.status == 404
raise GravityError, e.message
```
What do you think?

williardx: That makes sense! I'll change that.

PR #85, opened 2018-07-26
Calculate commission at submission time

ashkan18:
# Problem
We want to be able to calculate commission fees at order's submission.

# Solution
We already have `commission_rate` defined on `Partner` mode in Gravity, here we first fetch the partner from Gravity and we fetch `commission_rate` from there.

# Background vs. Non-Backgournd
I debated if i should calculate commission fee in the background (knowing it involves another call to Gravity), technically it's possible since Collector's submit the order and they don't care about the commission, it's important when we show it to partners which technically happens after submission. The only tricky part here is if for whatever reason the queue was backed up we won't be able to show commission to partners, so for now im actually calculating it during submission process.

# Depends On
https://github.com/artsy/gravity/pull/11831


config/gravity.yml
```diff
@@ -1,7 +1,8 @@
 development: &default
   api_v1_root: <%= "#{ENV['GRAVITY_API_ROOT'] || 'https://stagingapi.artsy.net'}/api/v1" %>
   xapp_token:  <%= ENV['GRAVITY_JWT'] || 'https://media.giphy.com/media/yow6i0Zmp7G24/giphy.gif' %>
-  artwork_cache_in_minutes: <%= (ENV['GRAVITY_JWT'] || '10').to_i %>
+  artwork_cache_in_minutes: <%= (ENV['ARTWORK_CACHE_IN_MINUTES'] || '10').to_i %>
+  partner_cache_in_minutes: <%= (ENV['PARTNER_CACHE_IN_MINUTES'] || '10').to_i %>
```
joeyAghion: Prefer seconds for this sort of thing...

spec/controllers/api/requests/submit_order_mutation_request_spec.rb
```diff
@@ -25,7 +26,7 @@
       )
     end
     let(:line_item) do
-      Fabricate(:line_item, order: order)
+      Fabricate(:line_item, order: order, price_cents: 100_000)
```
joeyAghion: `1000_00` would be more appropriate here, given cents. (If rubocop has a problem with that I'd disable the corresponding rule.)

joeyAghion:
I agree that calling Gravity to get the partner data in-request is reasonable here, especially with a cache.

The cache may be _more_ of a problem for artworks, which can change `availability` in critical ways as a result of orders.

spec/controllers/api/requests/submit_order_mutation_request_spec.rb
ashkan18: yes rubocop complained about it, I'll do it in separate PR since now there are parallel PRs that add price and will lead to some failures. Originally i had them cent-friendly and rubocop complained :(

ashkan18:
going to merge so i can try on staging.

PR #86, opened 2018-07-26
WIP Support for tracking fulfillments

### LINKED ISSUE #71: Add shipping fields to finalizeOrder mutation
Now that I've moved to the second page in CMS for the order detail page, I'm looking at the `finalizeOrder` mutation and so I think it's time to discuss how this should work. Here's the comp:

screen shot 2018-07-19 at 11 01 04 am

So what I want to send along in the mutation is:

* order id
* courier (fedex, ups, dhl, other)
* tracking id
* estimated delivery date

Note, that if you pick other for courier, you are meant to specify a name, but it appears that the field is not required, so I think both "other" and "local speedy delivery co" would be valid values. The former would mean that a partner picked other, but did not specify a name and the latter would mean the partner picked other and then specified a name.

ashkan18:
# Problem
We want to be able to set fulfillment tracking information when finalizing an order.

# Solution
We've added `Fulfillment` model (not 100% sure about the name), an `Order` can have many fulfillments, but for now `finalizeOrder` mutation gets one and adds it to the `Order` this is so we can handle possibilities later where we have multiple tracking and etc.

Updated `finalizeOrder` so the new input format is:
```json
{
  "input": {
    "id": "<order id>",
    "fulfillment": {
      "courier": "<some courier>",
      "trackingId": "track0123",
      "estimatedDelivery": "YYYY-MM-DD"
    }
  }
}
```

closes #71


db/migrate/20180726143043_create_fulfillments.rb
```diff
@@ -0,0 +1,12 @@
+class CreateFulfillments < ActiveRecord::Migration[5.2]
+  def change
+    create_table :fulfillments do |t|
+      t.references :order, foreign_key: true
+      t.string :courier
+      t.string :tracking_id
+      t.date :estimated_delivery
```
joeyAghion: I feel like fulfillments relate to 1 or more order line items, not just the order itself. We might not need this level of detail but it falls into the category of something that would be hard to retrofit.

ashkan18: true, but at the same time, i don't know if current UI allows selecting line items. I can imagine this might be something later on we can figure out with migration, right?

app/graphql/mutations/finalize_order.rb
```diff
@@ -2,15 +2,16 @@ class Mutations::FinalizeOrder < Mutations::BaseMutation
   null true
 
   argument :id, ID, required: true
+  argument :fulfillment, Types::FulfillmentAttributes, required: true
 
   field :order, Types::OrderType, null: true
   field :errors, [String], null: false
 
-  def resolve(id:)
+  def resolve(id:, fulfillment:)
```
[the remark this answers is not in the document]
ashkan18: we had similar situation with `Order`s having multiple `LineItems` but in our current scenarios and order can only have one line item, so we ended up providing a specific mutation `createOrderWithArtwork` that kind of assumes one line item per order. I tried to follow the same pattern here but now that im thinking about it, we may want to rename this `finalize` to something like `finalizeWithFulfillment` or something.

ashkan18:
after our chat this morning i think it's easier to close this PR and start from `master` 🙌

PR #87, opened 2018-07-27
Limit access to fields

ashkan18:
# Problem
Current schema allows for clients to request all available fields, this is not the desired behavior since we don't want buyer to see seller specific fields like `transaction_fee_cents` and etc.

# Solution
Used GraphQL Ruby's [Authorization](http://graphql-ruby.org/authorization/authorization) approach. We added `BaseField` in which:
- We added `buyer_only` and `seller_only` attributes which can be set on field definition
```ruby
field :seller_total_cents, Integer, null: true, seller_only: true
field :buyer_total_cents, Integer, null: true, buyer_only: true
```
- We added `authorized?` method to `BaseField` which is called by graphql ruby on every field being returned, current code only supports `buyer_only` and `seller_only` on `Order` type, but if needed we can add it to other Types too.
- In `authorized?` we check if Object passed in is an `Order` and if any of `buyer_only` or `seller_only` was defined on the field, based on the setting we check if current user is buyer or seller. This logic works fine on checking buyer since we check `order`'s `user_id` with current sessions and if they match current session is buyer. On the seller side we check if `order`'s `partner_id` was in the list of `partner_ids` in the session we allow accessing the field.

**Note** we still allow accessing all fields, we just return `nil` if user is not authorized.

# Ignore in this PR
ignore what I asked rubocop to ignore :) basically all price cents values no longer need `_` on every 3 digits since it was annoying on cents related fields.

### DIFF
--- a/app/graphql/types/base_field.rb
+++ b/app/graphql/types/base_field.rb
@@ -1,0 +1,27 @@
+class Types::BaseField < GraphQL::Schema::Field
+  def initialize(buyer_only: false, seller_only: false, **kwargs, &block)
+    @buyer_only = buyer_only
+    @seller_only = seller_only
+    super(**kwargs, &block)
+  end
+
+  def authorized?(obj, ctx)
+    case obj
+    when Order then order_field_authorized?(obj, ctx[:current_user])
+    else
+      super
+    end
+  end
+
+  private
+
+  def order_field_authorized?(obj, current_user)
+    if @buyer_only && current_user['id'] != obj.user_id
+      false
+    elsif @seller_only && !current_user[:partner_ids].include?(obj.partner_id)
+      false
+    else
+      true
+    end
+  end
+end
--- a/app/graphql/types/base_field.rb
+++ b/app/graphql/types/base_field.rb
@@ -16,12 +16,8 @@ class Types::BaseField < GraphQL::Schema::Field
   private
 
   def order_field_authorized?(obj, current_user)
-    if @buyer_only && current_user['id'] != obj.user_id
-      false
-    elsif @seller_only && !current_user[:partner_ids].include?(obj.partner_id)
-      false
-    else
-      true
-    end
+    return current_user[:id] == obj.user_id if @buyer_only
+    return current_user[:partner_ids].include?(obj.partner_id) if @seller_only
+    true
   end
 end
--- a/app/graphql/types/order_type.rb
+++ b/app/graphql/types/order_type.rb
@@ -19,10 +19,10 @@ class Types::OrderType < Types::BaseObject
   field :items_total_cents, Integer, null: false
   field :shipping_total_cents, Integer, null: true
   field :tax_total_cents, Integer, null: true
-  field :transaction_fee_cents, Integer, null: true
-  field :commission_fee_cents, Integer, null: true
-  field :buyer_total_cents, Integer, null: true
-  field :seller_total_cents, Integer, null: true
+  field :transaction_fee_cents, Integer, null: true, seller_only: true
+  field :commission_fee_cents, Integer, null: true, seller_only: true
+  field :seller_total_cents, Integer, null: true, seller_only: true
+  field :buyer_total_cents, Integer, null: true, buyer_only: true
   field :created_at, Types::DateTimeType, null: false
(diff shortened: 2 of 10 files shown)


spec/controllers/api/requests/order_field_permissions_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+            order(id: $id) {
+              id
+              buyerTotalCents
+              commissionFeeCents
+              sellerTotalCents
+            }
+          }
+        GRAPHQL
+      end
+      it 'returns nil for seller_only fields' do
```
williardx: `seller_only` should be `buyer_only` here

spec/controllers/api/requests/order_query_request_spec.rb
```diff
@@ -73,8 +73,14 @@
           expect(result.data.order.currency_code).to eq 'usd'
           expect(result.data.order.state).to eq 'PENDING'
           expect(result.data.order.items_total_cents).to eq 0
-          expect(result.data.order.seller_total_cents).to eq user2_order1.seller_total_cents
-          expect(result.data.order.buyer_total_cents).to eq user2_order1.buyer_total_cents
+        end
+
+        it 'cannot access seller_only buyer_only fields' do
+          # TODO: we may want to change this logic later but for now not allowing
```
williardx: How come trusted apps are excluded here? Wouldn't that affect Metaphysics?

ashkan18: it won't impact Metaphysics. MP uses user's token to access Exchange so we should be good here. 
Right now the only _trusted_ user i can think of is Gravity when trying to handle GDPR stuff and all that needs is to see an order with state, so we might be good here for now but we can def add trusted app logic later. what do you think?

app/graphql/types/order_type.rb
```diff
@@ -19,10 +19,10 @@ class Types::OrderType < Types::BaseObject
...
   field :shipping_total_cents, Integer, null: true
   field :tax_total_cents, Integer, null: true
-  field :transaction_fee_cents, Integer, null: true
-  field :commission_fee_cents, Integer, null: true
-  field :buyer_total_cents, Integer, null: true
-  field :seller_total_cents, Integer, null: true
+  field :transaction_fee_cents, Integer, null: true, seller_only: true
+  field :commission_fee_cents, Integer, null: true, seller_only: true
+  field :seller_total_cents, Integer, null: true, seller_only: true
+  field :buyer_total_cents, Integer, null: true, buyer_only: true
```
joeyAghion: I think sellers can be authorized to see this field. Accurately representing the total before fees may be useful for UIs.

spec/controllers/api/requests/create_order_with_artwork_mutation_request_spec.rb
```diff
@@ -67,7 +67,7 @@
               expect(order.user_id).to eq jwt_user_id
               expect(order.partner_id).to eq partner_id
               expect(order.line_items.count).to eq 1
-              expect(order.line_items.first.price_cents).to eq 540_012
+              expect(order.line_items.first.price_cents).to eq 5400_12
```
joeyAghion: 🙏

app/graphql/types/base_field.rb
```diff
@@ -0,0 +1,27 @@
...
+  private
+
+  def order_field_authorized?(obj, current_user)
+    if @buyer_only && current_user['id'] != obj.user_id
+      false
+    elsif @seller_only && !current_user[:partner_ids].include?(obj.partner_id)
+      false
+    else
+      true
+    end
```
joeyAghion: Minor but I can imagine some other structures for this like:

    return current_user['id'] == obj.user_id if @buyer_only
    return current_user[:partner_ids].include?(obj.partner_id) if @seller_only
    true

Also, why both the symbol and string methods of accessing fields on `current_user`?

ashkan18: we use `with_indifferent_access` when parsing JWT, so either works but going with symbol everywhere 👍

ashkan18:
going to merge, this will verify slack integration too for this repo :)

PR #89, opened 2018-07-30
Add new mutation for finalizing orders with one fulfillment

### LINKED ISSUE #71: Add shipping fields to finalizeOrder mutation

ashkan18:
# Problem
We started to introduce `Fulfillment` in #86  for tracking fulfillments per order. After some discussion we decided that the modeling could be better where we track fulfillment per `lineItem` instead of per order. 

Basically each line item can have many fulfillments (Imagine a big sculpture, or a line item with quantity more than one) and also each fulfillment could possibly handle multiple line items.

fixes #71 

# Solution
Added `Fulfillment` and `LineItemFulfillment` which is the connection model between `LineItem` and `Fulfillment`.
A `LineItem` can have many `fulfillments` and a `Fulfillment` can have many `line_items`.

We talked about having two separate mutations for this possibly, one for adding fulfillments and one for finalizing order. While this seems to be the proper way overall, in order to handle our current use case (one fulfillment per order) we only now provide one _helper_ mutation that gets order id and a fulfillment. For each line item in that order we set their fulfillments to this current one and then we finalize the order.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,14 +209,29 @@ For input:
 
 #### Finalize an Order
 ```graphql
-# finalize order
-mutation($input: FinalizeOrderInput!) {
-  finalizeOrder(input: $input) {
+# finalize order with one fulfillment
+mutation($input: FinalizeWithOneFulfillmentInput!) {
+  finalizeWithOneFulfillment(input: $input) {
     order {
       id
       userId
       partnerId
       state
+      lineItems{
+        edges{
+          node{
+            fulfillments{
+              edges{
+                node{
+                  courier
+                  trackingId
+                  estimatedDelivery
+                }
+              }
+            }
+          }
+        }
+      }
     }
     errors
   }
@@ -226,7 +241,11 @@ For input:
 ```json
 {
   "input": {
-    "id": "<order id>"
+    "id": "<order id>",
+    "fulfillment": {
+      "courier": "FedEx",
+      "trackingId": "track-431"
+    }
   }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ In order to talk to Exchange GraphQL endpoint:
 
 ## Order Lifecycle
 
-Users create and submit orders. Partners then approve and finalize or reject the order.
+Users create and submit orders. Partners then approve and fulfill or reject the order.
 
 ### As a User
 
@@ -207,11 +207,11 @@ For input:
 ```
 
 
-#### Finalize an Order
+#### Fulfill an Order
 ```graphql
-# finalize order with one fulfillment
-mutation($input: FinalizeWithOneFulfillmentInput!) {
-  finalizeWithOneFulfillment(input: $input) {
+# fulfill order with one fulfillment
+mutation($input: FulfillWithOneFulfillmentInput!) {
+  fulfillWithOneFulfillment(input: $input) {
     order {
       id
       userId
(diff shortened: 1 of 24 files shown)


README.md
```diff
@@ -209,14 +209,29 @@ For input:
 
 #### Finalize an Order
 ```graphql
-# finalize order
-mutation($input: FinalizeOrderInput!) {
-  finalizeOrder(input: $input) {
+# finalize order with one fulfillment
+mutation($input: FinalizeWithOneFulfillmentInput!) {
+  finalizeWithOneFulfillment(input: $input) {
```
joeyAghion: As discussed I'm not a big fan of the "finalize" language because there will be further states after fulfillment.

app/graphql/types/fulfillment_attributes.rb
```diff
@@ -0,0 +1,8 @@
+class Types::FulfillmentAttributes < Types::BaseInputObject
+  description 'Attributes of a Fulfillment'
+
+  argument :courier, String, required: true
+  argument :tracking_id, String, required: false
+  argument :estimated_delivery, Types::DateType, required: false
+  argument :notes, String, required: false
```
joeyAghion: How will in-person pick-ups be represented in the data? Will they have 0 fulfillments, or fulfillments that are distinguished in some way? It might be worth modeling them as a kind of fulfillment, because that would support orders that are partially picked-up and partially shipped.

ashkan18: so that `courier` field is just string with not much validation, i was wondering that will be `pickup` or something and since `tracking_id` is optional i thought we are good, but we can add separate field for this special case too.
Although i don't know how to parse things if order fulfillment type is `PICKUP` but fulfillment data does not match.

README.md
ashkan18: going to talk to @joanna20 about it since we had a list of states in a doc somewhere.

ashkan18: we are good with calling the state `fulfilled` => https://www.notion.so/artsy/37c311363ef046c3aa546047e60cc58a?v=de68d5bbc30748f88b0d92a059bc0ba8

app/graphql/mutations/fulfill_with_one_fulfillment.rb
```diff
@@ -1,16 +1,18 @@
-class Mutations::FinalizeOrder < Mutations::BaseMutation
+class Mutations::FulfillWithOneFulfillment < Mutations::BaseMutation
```
joeyAghion: This PR looks good with the most recent updates. My only 🤔is about the awkward `FulfillWithOneFulfillment` name. Not sure what's better but maybe `fulfillAtOnce` or `addSoleFulfillment` instead.

ashkan18:
🙌

PR #91, opened 2018-07-30
Estimate sales tax on orders

williardx:
This is a first pass at implementing sales tax estimations for sellers.

## What changed
- Calculates sales tax for each line item in an order using TaxJar's API
- Adds `fetch_partner_location` to `GravityService`, which retrieves a partner's billing location based on their partner id.
- Adds `taxjar_helper` which provides helper methods to stub TaxJar API calls and get back a realistic response. 
  - Parts of this were pulled from [taxjar-ruby](https://github.com/taxjar/taxjar-ruby), e.g. the fixture for the response to the `/taxes` call. They use the same license as Exchange (MIT), so I don't believe there are any problems there but I'm also not a licensing expert.
- Adds `unit_converter` helper that converts between cents and dollars since TaxJar takes dollars and we use cents.

## What this does _not_ do
- Estimate sales tax for situations where Artsy is expected to be the merchant of record
- Post completed transactions with sales tax estimations to TaxJar

## Migrations
- ~This work depends on to-be-completed work in Gravity to expose a partner's billing location to trusted apps.~
- ~Add `TAXJAR_API_KEY` to Exchange staging and production~


app/services/order_service.rb
```diff
@@ -10,22 +10,23 @@ def self.set_payment!(order, credit_card_id:)
...
-          attributes.slice(
-            :shipping_address_line1,
-            :shipping_address_line2,
-            :shipping_city,
-            :shipping_region,
-            :shipping_country,
-            :shipping_postal_code,
-            :fulfillment_type
-          )
+          **shipping_address,
```
ashkan18: 💯

app/services/order_service.rb
```diff
@@ -10,22 +10,23 @@ def self.set_payment!(order, credit_card_id:)
...
+      :shipping_city,
+      :shipping_region,
+      :shipping_country,
+      :shipping_postal_code
+    )
     Order.transaction do
       attrs = {
         shipping_total_cents: order.line_items.map { |li| ShippingService.calculate_shipping(li, attributes.slice(:shipping_country, :fulfillment_type)) }.sum,
-        tax_total_cents: 100_00 # TODO: 🚨 replace this with real tax calculation 🚨
+        tax_total_cents: SalesTaxService.calculate_total_sales_tax(order, attributes[:fulfillment_type], shipping_address)
```
ashkan18: this is super super minor but if you want, you could change `calculate_total_sales_tax` to receive `fulfillment_type` also as keyword argument and basically add `fulfillment_type` to `attributes.slice` above. But your current code also is clean and easy to read so whichever you prefer.

app/services/sales_tax_service.rb
```diff
@@ -0,0 +1,26 @@
...
+    raise Errors::ApplicationError, e.message
+  end
+
+  def self.fetch_sales_tax(_price_cents, _origin_address, _destination_address)
+    # TODO: 🚨 Add in sales tax API call 🚨
+    100
+  end
+
+  def self.artsy_should_remit_taxes?(destination_address)
+    remitting_states = %w[
```
ashkan18: this might worth making a const in the same file so something like this
```ruby
REMITTING_STATES = %w[....].freeze
```
If they way we create charge is going to be different for these states this const will be used from other places too.

williardx: Good idea! I'll change that.

williardx:
@ashkan18 This is ready to review now! It's different from the placeholder code that was here before.

app/services/sales_tax_service.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def self.calculate_total_sales_tax(order, shipping)
+    shipping_address = {
+      country: shipping[:shipping_country],
+      postal_code: shipping[:shipping_postal_code],
+      state: shipping[:shipping_region],
+      city: shipping[:shipping_city],
+      address: shipping[:shipping_address_line_1]
+    }
+    seller_address = GravityService.fetch_partner_location(order.partner_id)
```
ashkan18: #minor suggestion, but we could avoid this call for `PICKUP` orders, right? We could have a check in this method that if Order is `PICKUP` then call `calculate_pickup_sales_tax` and otherwise `calculate_shipping_sales_tax` and in both of them maybe use your existing `calculate_line_item_sales_tax` which now only gets a `destination_address`. This might also make things little more obvious in terms of how pickup and ship are different tax-wise.

spec/controllers/api/requests/set_shipping_mutation_request_spec.rb
```diff
@@ -64,7 +76,11 @@
         end
       end
 
-      it 'sets shipping info on the order' do
+      it 'sets shipping info and sales tax on the order' do
+        allow(Adapters::GravityV1).to receive(:request).twice.with('/artwork/a-1').and_return(artwork1)
```
ashkan18: hmmm... why `twice`? is there something we could improve here?

app/services/sales_tax_service.rb
williardx: I think it's safer to assume little to no knowledge about how taxes work and pass TaxJar all potentially relevant information, which would include the seller's primary business address.

spec/controllers/api/requests/set_shipping_mutation_request_spec.rb
williardx: The methods for setting shipping and sales tax both call `GravityService.get_artwork`, so that's where our twice comes from. We were previously caching this in Exchange but that was giving us some grief. Is this a problem if Gravity is caching the artwork?

app/services/sales_tax_service.rb
ashkan18: oh ignore my comment, i see that we still send `seller_address` to Taxjar, i thought we only send `seller` or `artwork`, makes total sense 👍

spec/support/taxjar_helper.rb
```diff
@@ -0,0 +1,76 @@
+require 'json'
+require 'taxjar'
+require 'webmock/rspec'
+
+def stub_tax_for_order
+  stub_request(:post, Taxjar::API::Request::DEFAULT_API_URL + '/v2/taxes').to_return(body: sales_tax_fixture.to_json, headers: { content_type: 'application/json; charset=utf-8' })
+end
+
+def sales_tax_fixture
```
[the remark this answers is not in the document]
williardx: Nice! That's cool - will keep that in mind for next time

williardx:
Going to go ahead and merge to unblock myself - we can address the question of multiple calls to the same Gravity endpoint in another PR if need to be

app/services/sales_tax_service.rb
```diff
@@ -0,0 +1,63 @@
...
+      to_street: destination_address[:address],
+      nexus_addresses: [
+        {
+          country: seller_address[:country],
+          zip: seller_address[:postal_code],
+          state: seller_address[:state],
+          city: seller_address[:city],
+          street: seller_address[:address]
+        }
+      ],
```
joeyAghion: The fact that they expect an array and we're only considering the one billing location causes me some concern but 🤷‍♂️.

config/taxjar.yml
```diff
@@ -0,0 +1,6 @@
+development: &default
+  taxjar_api_key: <%= ENV['TAXJAR_API_KEY'] || '' %>
```
joeyAghion: This could just be `api_key` since it's already namespaced to `taxjar`.

app/services/sales_tax_service.rb
```diff
@@ -0,0 +1,63 @@
+require 'taxjar'
+require 'unit_converter'
```
joeyAghion: `require 'taxjar'` shouldn't be necessary given the Gemfile.

`require 'unit_converter'` _might_ be necessary at the moment, but I usually advise adding `lib` to the eager-loaded paths ([link](https://gist.github.com/maxivak/381f1e964923f1d469c8d39da8e2522f)) so it becomes unnecessary.

app/services/sales_tax_service.rb
```diff
@@ -0,0 +1,63 @@
...
+        }
+      ],
+      shipping: UnitConverter.convert_cents_to_dollars(shipping_total_cents)
+    )
+  end
+
+  def self.artsy_should_remit_taxes?(destination_address)
+    REMITTING_STATES.include? destination_address[:shipping_region].downcase
+  end
+end
```
joeyAghion: This is great encapsulation of the sales tax concern. A design change to consider:

Rather than the service being a module purely for file and code organization, there are advantages to defining it as an actual class.
* You could instantiate one with the order model (and whatever other inputs), so as to avoid having to pass around the context via arguments.
* You could clearly distinguish the public methods (interface) from the private/helper methods.
* You could load and memo-ize any external resources just once for the life of the service instance.
* And you could even accept optional dependencies (like Gravity and Taxjar adapters) to ease testing. E.g.:

```ruby
class SalesTaxService
  def initialize(order, fulfillment_type, shipping, shipping_total_cents, gravity = GravityService, tax = Taxjar::Client.new(...))
    @order = order
    @gravity = gravity
    ...
  end

  def total_sales_tax
    @total_sales_tax ||= begin
      @order.line_items.map { |li| line_item_sales_tax(li) }.sum
    end
  end

  private

  def seller_address
    @seller_address ||= gravity.fetch_partner_location(@order.partner_id)
  end

  def artwork(id)
    @artworks ||= {}
    @artworks[id] ||= @gravity.get_artwork(id)
  end

  ...
end
```

Apologies for suggesting a refactor so early, but I've found that this kind of design can make a lot of things easier down the road.

spec/controllers/api/requests/set_shipping_mutation_request_spec.rb
```diff
@@ -1,14 +1,21 @@
 require 'rails_helper'
 require 'support/gravity_helper'
+require 'support/taxjar_helper'
```
joeyAghion: Once these are in more than a few spec files, I usually just add them once to spec_helper.rb (or rails_helper.rb these days).

spec/controllers/api/requests/set_shipping_mutation_request_spec.rb
joeyAghion: I think loading the artwork twice _is_ a problem when it happens in-request (like here). The memo-ization I suggested above is helpful for this kind of thing.

app/services/sales_tax_service.rb
williardx: Do you think we should be putting in more locations here? I'm only aware of the billing location that might be an additional relevant tax nexus.

config/taxjar.yml
williardx: 👍

app/services/sales_tax_service.rb
williardx: Looks like we already added `lib` to eager-loaded paths so I'll take that one out, too.

app/services/sales_tax_service.rb
joeyAghion: I think this correctly implements the business rules _we discussed_. My understanding of nexus ([source](https://blog.taxjar.com/sales-tax-nexus-definition/)) is that it would consider at least all physical gallery locations, and maybe others. (This isn't meant to distract you and is mostly a product question /cc @madeleineb.)

williardx: Good point! We should figure out if we need _all_ artwork locations in the tax nexus field -- based on what you linked it does sound like we do.

PR #92, opened 2018-07-31
Transaction fee cents

ashkan18:
# Problem
We want to calculate Transaction fee when an order is submitted.

# Solution
We now calculate and set `transaction_fee_cents` on an order when order is submitted. This fee currently matches Stripe's US fee which would work for us as long as Partner's merchant account is setup in a US bank.

# Some Refactor
@williardx based on what we talked about before, i moved `get_merchant_account` and `get_credit_card` methods to `GravityService`, let me know what you think 👍

### DIFF
--- a/app/services/gravity_service.rb
+++ b/app/services/gravity_service.rb
@@ -4,4 +4,22 @@ module GravityService
       Adapters::GravityV1.request("/partner/#{partner_id}/all")
     end
   end
+
+  def self.get_merchant_account(partner_id)
+    merchant_account = Adapters::GravityV1.request("/merchant_accounts?partner_id=#{partner_id}").first
+    raise Errors::OrderError, 'Partner does not have merchant account' if merchant_account.nil?
+    merchant_account
+  rescue Adapters::GravityNotFoundError
+    raise Errors::OrderError, 'Unable to find partner or merchant account'
+  rescue Adapters::GravityError => e
+    raise Errors::OrderError, e.message
+  end
+
+  def self.get_credit_card(credit_card_id)
+    Adapters::GravityV1.request("/credit_card/#{credit_card_id}")
+  rescue Adapters::GravityNotFoundError
+    raise Errors::OrderError, 'Credit card not found'
+  rescue Adapters::GravityError => e
+    raise Errors::OrderError, e.message
+  end
 end
--- a/app/services/order_submit_service.rb
+++ b/app/services/order_submit_service.rb
@@ -3,8 +3,9 @@ module OrderSubmitService
     # verify price change?
     raise Errors::OrderError, "Missing info for submitting order(#{order.id})" unless can_submit?(order)
 
-    merchant_account = get_merchant_account(order)
-    credit_card = get_credit_card(order)
+    merchant_account = GravityService.get_merchant_account(order.partner_id)
+    credit_card = GravityService.get_credit_card(order.credit_card_id)
+    validate_credit_card!(credit_card)
 
     charge_params = {
       source_id: credit_card[:external_id],
@@ -20,6 +21,7 @@ module OrderSubmitService
       order.external_charge_id = charge[:id]
       TransactionService.create_success!(order, charge)
       order.commission_fee_cents = calculate_commission(order)
+      order.transaction_fee_cents = calculate_transaction_fee(order)
       order.save!
       PostNotificationJob.perform_later(order.id, Order::SUBMITTED, by)
     end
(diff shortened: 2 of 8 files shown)


app/services/gravity_service.rb
```diff
@@ -4,4 +4,22 @@ def self.fetch_partner(partner_id)
       Adapters::GravityV1.request("/partner/#{partner_id}/all")
     end
   end
+
+  def self.get_merchant_account(partner_id)
+    merchant_account = Adapters::GravityV1.request("/merchant_accounts?partner_id=#{partner_id}").first
+    raise Errors::OrderError, 'Partner does not have merchant account' if merchant_account.nil?
+    merchant_account
+  rescue Adapters::GravityNotFoundError
+    raise Errors::OrderError, 'Unable to find partner or merchant account'
```
williardx: Should these still raise `OrderError` now that they're not tied to `OrderService`? I almost want there to be a `GravityServiceError` but am not sure how that would be different from `GravityError`

app/services/order_submit_service.rb
```diff
@@ -3,8 +3,9 @@ def self.submit!(order, by: nil)
     # verify price change?
     raise Errors::OrderError, "Missing info for submitting order(#{order.id})" unless can_submit?(order)
 
-    merchant_account = get_merchant_account(order)
-    credit_card = get_credit_card(order)
+    merchant_account = GravityService.get_merchant_account(order.partner_id)
```
williardx: 👏

app/services/order_submit_service.rb
```diff
@@ -67,4 +49,9 @@ def self.calculate_commission(order)
     Rails.logger.error("Could not fetch partner for order #{order.id}: #{e.message}")
     raise Errors::OrderError, 'Cannot fetch partner'
   end
+
+  def self.calculate_transaction_fee(order)
+    # This is based on Stripe US fee, it will be different for other countries
+    (Money.new(order.buyer_total_cents * 2.9 / 100, 'USD') + Money.new(30, 'USD')).cents
```
williardx: Would it make sense to the 2.9% and $0.30 as environment variables? That way we could change it easily without having to commit and deploy, and also gives us a place to find it.

app/services/order_submit_service.rb
```diff
@@ -30,27 +32,7 @@ def self.submit!(order, by: nil)
...
-    validate_credit_card(credit_card)
-    credit_card
-  rescue Adapters::GravityNotFoundError
-    raise Errors::OrderError, 'Credit card not found'
-  rescue Adapters::GravityError => e
-    raise Errors::OrderError, e.message
-  end
-
-  def self.validate_credit_card(credit_card)
+  def self.validate_credit_card!(credit_card)
```
williardx: How come you added an `!` onto the method name? My understanding is that `!` means the method mutates the object it's acting on or is otherwise dangerous, but neither of those seem to fit here

ashkan18: im still little fuzzy about `!` in general but the way i've heard in some places it's being used is it's considered as _this is dangerous, be careful_ in this case im thinking `!` here means it raises error be careful :) Kind like how `find_by` and `find_by!` are defined in ActiveRecord https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by-21

app/services/order_submit_service.rb
ashkan18: i feel like having it in `ENV` might make it too convenient for us to modify this, it might be ok thing to make this change kind of a big deal, like going through code review and etc. what do you think?

williardx: Okay that makes sense to me -- I was thinking about the hypothetical situation of needing to change it in a pinch, but I don't _think_ that should happen since we should know well in advance when Stripe is changing their fees. I think it'd be good to have in a constants file, though, or at least as named variables to make it clearer what those numbers are.

app/services/order_submit_service.rb
williardx: Okay cool, sounds good to me 👍

app/services/gravity_service.rb
ashkan18: not sure how github lost my comments here! this is a good point, i was thinking about removing `GravityError` from Adapter and let discuss raise that if needed. what do you think?

ashkan18: looking deeper at it, Im actually not too against what we currently have, it's nice to have `Adapter` to raise `GravityError` and in our _service_ layer we return more business related errors, the only thing is `OrderError` little too specific, maybe later we can rename it to `GenericError`/`ValidationError` or something. We can do that in a follow up PR though.

ashkan18:
merging this and we can address exception rename/cleanup in a separate PR 🙌

PR #93, opened 2018-07-31
Move to the new mutation style

orta:
I won't get time to finish this today it looks like


orta:
This *will* be a breaking change in  Metaphysics schema, which will be a breaking change in Reaction's source code. So, let's get those up and ready too.

app/graphql/mutations/create_order_with_artwork.rb
```diff
@@ -1,17 +1,17 @@
 class Mutations::CreateOrderWithArtwork < Mutations::BaseMutation
+  null true
+
   argument :artwork_id, String, 'Artwork Id', required: true
   argument :edition_set_id, String, 'EditionSet Id', required: false
   argument :quantity, Integer, 'Number of items in the line item', required: false
 
-  field :order, Types::OrderType, null: true
-  field :errors, [String], null: false
+  field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure', null: false
```
orta: Converts from two nullable objects, to a union where none of the items are null. It can either _definitely_ be an order, or it can definitely be an error.

app/graphql/mutations/create_order_with_artwork.rb
```diff
@@ -1,17 +1,17 @@
...
   def resolve(artwork_id:, edition_set_id: nil, quantity: 1)
     {
-      order: CreateOrderService.with_artwork!(user_id: context[:current_user][:id], artwork_id: artwork_id, edition_set_id: edition_set_id, quantity: quantity),
-      errors: []
+      order_or_error: { order: CreateOrderService.with_artwork!(user_id: context[:current_user][:id], artwork_id: artwork_id, edition_set_id: edition_set_id, quantity: quantity) }
     }
   rescue Errors::ApplicationError => e
-    { order: nil, errors: [e.message] }
+    { order_or_error: { error: Types::MutationErrorType.from_application(e) } }
   end
```
orta: The hash we return will always contain `order_or_error ` but inside that it's either an `error`, or an `order`.

app/graphql/mutations/order_or_failure_union_type.rb
```diff
@@ -0,0 +1,24 @@
...
+  def self.resolve_type(object, _context)
+    if object.key?(:order)
+      Mutations::OrderWithMutationSucesss
+    elsif object.key?(:error)
+      Mutations::OrderWithMutationFailure
+    else
+      raise "Unexpected Return value: #{object.inspect}"
+    end
+  end
+end
```
orta: These are generic and available to use on any mutation that touches an order

app/graphql/types/mutation_error_type.rb
```diff
@@ -0,0 +1,15 @@
...
+  field :failure_reason, String, null: true, description: 'What caused the error'
+  field :recovery_suggestion, String, null: true, description: 'What can the user do to fix the error'
+  field :code, Integer, null: true, description: 'Optional error code, for example when coming from a REST request'
+
+  def self.from_application(err)
+    {
+      description: err.message
+    }
+  end
+end
```
orta: This can be expanded, but it's simple enough for now.

spec/controllers/api/requests/create_order_with_artwork_mutation_request_spec.rb
```diff
@@ -46,8 +55,10 @@
         it 'does not create order and returns proper error' do
           expect do
             response = client.execute(mutation, input: mutation_input)
-            expect(response.data.create_order_with_artwork.order).to be_nil
-            expect(response.data.create_order_with_artwork.errors).to match ['Unknown artwork artwork-id']
+            expect(response.data.create_order_with_artwork.order_or_error).not_to respond_to(:order)
+            expect(response.data.create_order_with_artwork.order_or_error.error).not_to be_nil
```
orta: Note how we can't do `expect(x).to be_nil` anymore, it has to be `expect(x).not_to respond_to(:order)` because the types now state that it's one of the other.

dangerfile.ts
```diff
@@ -1,7 +1,15 @@
...
   warn("If you want these changes to be reflected in Metaphysics, you will need to [update the stored schema](https://github.com/artsy/exchange#did-you-change-graphql-schema).")
 }
+
+if (danger.git.created_files.includes("schema.graphql")) {
+  fail("Please remove the schema from your PR `rm schema.graphql` - it's meant to go to metaphysics.")
+}
+
+if (danger.git.created_files.includes("exchange.graphql")) {
+  fail("Please remove the schema from your PR `rm exchange.graphql` - it's meant to go to metaphysics.")
+}
```
orta: I ended up shipping both of these to the PR at one point, so better to let that not happen to someone else.

app/graphql/mutations/create_order_with_artwork.rb
alloy: Can we consider using camel-case for these like we’ve been trying to do for new MP fields? Especially when we’d stitch this into MP’s schema, it’s not going to be very intuitive that some are called `fooOrError` and others are called `foo_or_error`.

orta: It *always* makes a schema with `camelCase`, and does all the conversion under the hood - this lost me three hours of debugging with @yuki24  - [here](https://github.com/artsy/metaphysics/pull/1231/files#diff-e153ed58c608d37675000c96fe9993e3R37) it is in the schema

ashkan18: yes, this is confusing/little annoying/still useful in some ways :)

alloy: Wow, that’s unexpected!

But cool, that’s solved then 👍

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "ruby.intellisense": "rubyLocate"
```
ashkan18: 🔍 🔎

app/graphql/mutations/plain_mutation.rb
```diff
@@ -0,0 +1,2 @@
+class Mutations::PlainMutation < GraphQL::Schema::Mutation
```
ashkan18: is this used anywhere?

app/graphql/types/mutation_error_type.rb
```diff
@@ -0,0 +1,15 @@
+class Types::MutationErrorType < Types::BaseObject
+  description 'An generic error type for mutations'
+  graphql_name 'MutationError'
+
+  field :description, String, null: false, description: 'What error has occured'
+  field :failure_reason, String, null: true, description: 'What caused the error'
+  field :recovery_suggestion, String, null: true, description: 'What can the user do to fix the error'
+  field :code, Integer, null: true, description: 'Optional error code, for example when coming from a REST request'
```
ashkan18: should we set these fields? im also down for removing them for now since we only have description.

.vscode/settings.json
orta: Now Ruby does jump to symbol etc in vscode

app/graphql/mutations/plain_mutation.rb
orta: Yeah, the new mutation is a subclass of `Mutations::PlainMutation` - it's recommended in the docs.

app/graphql/types/mutation_error_type.rb
orta: I was figuring we'd make ^ the interface that Metaphysics uses for all errors, so then it'd be consistent in here

dangerfile.ts
ds300: would it make sense to add them to .gitignore?

ashkan18:
lets do this! https://artsy.slack.com/archives/CB110C6LE/p1535034832000100

ds300:
I think this part of the README should have been updated in this PR?

https://github.com/artsy/exchange/blob/master/README.md#create-an-order-with-an-artwork

PR #94, opened 2018-07-31
Add active admin

ashkan18:
# Feature
Added a simple active admin with `artsy-auth` in front of it that only allows `sales_admin` to see this dashboard. 

This PR includes:
- setting up `active_admin`
- setting up `artsy-auth` 
- allow only `sales_admin` users to access this dashboard
- update active admin to use `artsy-auth` this basically means:
  - make app not be `api_only`
  - update `application.rb` to require `require 'sprockets/railtie'`
  - Create `BaseApiController` for Api controllers and update `ApplicationController` to include admin related auth checks.
  - added only `order` page.

# Migrations
Set following env variables:
```
EXCHANGE_APP_ID
EXCHANGE_APP_SECRET
```
- [x] Staging
- [x] Production

# Selfie

screen shot 2018-07-31 at 4 12 32 pm

screen shot 2018-07-31 at 4 12 42 pm

# Follow up
- [x] Clean up fields we show
- [x] Limit actions (we don't want to delete any record for example)


bhoggard:
There is 1 rubocop error, otherwise LGTM.

ashkan18:
thanks! going to merge to try on Staging (and prod :) ) I'll address any code review comments in follow up PR

PR #97, opened 2018-08-06
Fix csrf issue

sweir27:
I actually am not sure why this was working correctly before, but this appears to fix the exchange issues we've been seeing.


jonallured:
Perfecto! Merge on green please.

PR #98, opened 2018-08-07
Raise error when calculating shipping for artwork without location

starsirius:
https://artsy.slack.com/archives/CB110C6LE/p1533670813000231

We saw a `Cannot match against 'undefined' or 'null'.` error when using MP to set shipping for artworks without location. This makes it raise proper error with messages when location is missing.

Another case not addressed here is a location without country will be treated as international, given current implementation. Not sure if this will happen in Gravity but want to raise it here. 

screen shot 2018-08-07 at 5 51 36 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ For input:
   "input": {
     "id": "<order id>",
     "fulfillmentType": "SHIP/PICKUP",
-    "shippingAddresLine1": "",
+    "shippingAddressLine1": "",
     "shippingAddressLine2": "",
     "shippingCity": "",
     "shippingRegion": "",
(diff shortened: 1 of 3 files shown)


sepans:
🙏 @starsirius

starsirius:
@williardx I expect most of the current artworks don't have an associated location because there was no use case for most partners. The `location` field was introduced for a special partner type: private collectors, to specify where a work is located, and I believe we only have a few of them.

In BNMO, we planned to reuse this field and made it required for works opted in BNMO, so I imagine it'll be an edge case missing it. I'm not sure defaulting it to billing location will make sense, say, for calculating shipping, tax, etc., but we will learn more down the line.

PR #99, opened 2018-08-09
Better error handling for Gravity requests

williardx:
### Problem
During BNMO demo we found our error handling for 404's from Gravity was not very robust or informative. This was partially because we raise a `GravityNotFoundError` in `GravityV1.request` and then end up rescuing it and re-raising a `GravityError`, leading to less helpful error handling.

### Solution
Allow `GravityV1.request` to raise whatever errors it wants and let the methods that wrap it rescue errors with the correct error messages.

### What changed
- Removes rescuing from `GravityV1.request`
- Adds rescues for `StandardError` in addition to `GravityError` for `GravityService` methods
- Adds error handling for `GravityService.fetch_partner`
- Adds tests to ensure that `GravityService` methods properly catch 404's
- Opportunistically adds a column that apparently wasn't previously committed to `schema.rb`.

### DIFF
--- a/app/services/create_order_service.rb
+++ b/app/services/create_order_service.rb
@@ -17,7 +17,7 @@ module CreateOrderService
 
   def self.get_artwork(artwork_id)
     Adapters::GravityV1.request("/artwork/#{artwork_id}")
-  rescue Adapters::GravityError => e
+  rescue Adapters::GravityError, StandardError => e
     Rails.logger.warn("Could not fetch artwork #{artwork_id} from gravity: #{e.message}")
     nil
   end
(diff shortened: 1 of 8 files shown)


app/services/create_order_service.rb
```diff
@@ -17,7 +17,7 @@ def self.with_artwork!(user_id:, artwork_id:, edition_set_id: nil, quantity:)
 
   def self.get_artwork(artwork_id)
```
sweir27: +1 for consolidating logic/error-handling and for the added tests!

Highlighted a question here though-- why is this under the `create_order_service` instead of `gravity_service`? Might be nice to keep all of the explicit gravity handling in one place.

williardx: Yep it should be there! I had left that change out originally since I was expecting to do more work on this and didn't want to expand the scope. I'll put it into `gravity_service`.

PR #100, opened 2018-08-10
add order data to rabbitmq event

erikdstock:
This PR: 
- updates `OrderEvent` to use the more general 'ecommerce' `TOPIC` name I've been using in pulse
- adds a collected `shipping_info` object to the properties payload
- adds a `#to_s` method to prevent sending an ugly ruby object through amqp.

Tracked in https://artsyproduct.atlassian.net/browse/PURCHASE-369

### DIFF
--- a/app/events/order_event.rb
+++ b/app/events/order_event.rb
@@ -1,5 +1,5 @@
 class OrderEvent < Events::BaseEvent
-  TOPIC = 'BNMO'.freeze
+  TOPIC = 'ecommerce'.freeze
 
   def self.post(order, action, user_id)
     event = new(user: user_id, action: action, model: order)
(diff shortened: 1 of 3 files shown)


app/events/order_event.rb
```diff
@@ -1,5 +1,5 @@
 class OrderEvent < Events::BaseEvent
-  TOPIC = 'BNMO'.freeze
+  TOPIC = 'ecommerce'.freeze
```
joeyAghion: I definitely prefer this over the temporary/marketing-specific labeling. I might generalize it even further to just "commerce" since the e- doesn't really distinguish anything.

app/events/order_event.rb
```diff
@@ -15,13 +15,23 @@ def subject
...
-      partner_id: @object.partner_id,
+      created_at: @object.created_at,
       currency_code: @object.currency_code,
-      state: @object.state,
       items_total_cents: @object.items_total_cents,
       line_items: @object.line_items.map { |li| line_item_detail(li) },
-      updated_at: @object.updated_at,
-      created_at: @object.created_at
+      partner_id: @object.partner_id,
+      shipping_info: shipping_info(@object),
```
joeyAghion: This introduces a different structure among what are otherwise standard, already-namespaced order fields. (Anyone referencing these will have to know that, in this context, the fields must be accessed like `data.shipping_info.shipping_...` instead of the way they're referenced elsewhere.) If there isn't a good downstream reason for this, I'd prefer just returning them in the usual flat structure.

app/events/order_event.rb
erikdstock: heh- We started with commerce but I convinced barry to add the `e` since i had named things in pulse `ecommerce`. That can be fixed later though- i agree this is better.

williardx:
👍 👍

PR #101, opened 2018-08-13
Fetch credit card and store credit card details in setPayment mutation

williardx:
### Problem
See https://github.com/artsy/exchange/issues/54#issuecomment-411809399.

### Solution
- Moves fetching the user's credit card from Gravity from `OrderSubmitService` to `OrderService.set_payment!`
- Adds `external_credit_card_id` and `external_customer_id` columns to `orders`
- Saves Stripe credit card and customer ID on order model

### DIFF
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -2,7 +2,13 @@ module OrderService
   def self.set_payment!(order, credit_card_id:)
     raise Errors::OrderError, 'Cannot set payment info on non-pending orders' unless order.state == Order::PENDING
     Order.transaction do
-      order.update!(credit_card_id: credit_card_id)
+      credit_card = GravityService.get_credit_card(credit_card_id)
+      validate_credit_card!(credit_card)
+      order.update!(
+        credit_card_id: credit_card_id,
+        external_credit_card_id: credit_card[:external_id],
+        external_customer_id: credit_card[:customer_account][:external_id]
+      )
     end
     order
   end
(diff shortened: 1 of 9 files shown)


app/services/order_submit_service.rb
```diff
@@ -2,14 +2,10 @@ module OrderSubmitService
...
     raise Errors::OrderError, "Missing info for submitting order(#{order.id})" unless can_submit?(order)
-
     merchant_account = GravityService.get_merchant_account(order.partner_id)
-    credit_card = GravityService.get_credit_card(order.credit_card_id)
-    validate_credit_card!(credit_card)
-
     charge_params = {
-      source_id: credit_card[:external_id],
-      customer_id: credit_card[:customer_account][:external_id],
+      source_id: order.external_credit_card_id,
```
sweir27: This worries me a little bit-- since we're splitting up the `setPayment` and `submit` actions, a user may alter their saved credit card in-between hitting those two steps. If we don't re-fetch for the credit card (or listen on updates here and update saved information as-needed), we risk charging someone who had intentionally deleted their gravity record.

It actually doesn't bother me that we'd fetch for the credit card again, as this may happen sometime later.

app/services/order_service.rb
```diff
@@ -2,7 +2,13 @@ module OrderService
   def self.set_payment!(order, credit_card_id:)
     raise Errors::OrderError, 'Cannot set payment info on non-pending orders' unless order.state == Order::PENDING
     Order.transaction do
-      order.update!(credit_card_id: credit_card_id)
+      credit_card = GravityService.get_credit_card(credit_card_id)
+      validate_credit_card!(credit_card)
```
sweir27: This is probably overkill in the current use-case (since we'll be calling `setPayment` in the same step as creating the credit card in gravity), but will likely be more relevant if we ever ask people to use a saved card.

williardx:
Closing this out in favor of the status quo due to ☝️ ! (i.e. https://github.com/artsy/exchange/pull/101/files/991dd857e437b7a29f997fce9fc61ef5b6fd6c66#r209756942)

PR #104, opened 2018-08-15
Refund charges for rejected orders

williardx:
### Problem
We need to refund a charge when an order is rejected.

### Solution
On order rejection, create a new `Stripe::Refund` object with the order's charge ID and record the transaction.

### What changed
- Adds `PaymentService.refund_charge` method, which takes in a Stripe charge ID and refunds it.
- Adds `PaymentService.refund_charge` to `OrderService.reject!` as well as a record of the refund transaction. The refund transaction stores the charge ID and the amount.
- Removes `TransactionService.create_success!` and `TransactionService.create_failure!` in favor of a single method `TransactionService.create!` that accepts a hash of the fields used in a transaction. Updates associated tests and code.

### DIFF
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -86,6 +86,7 @@ module OrderService
     rescue Errors::PaymentError => e
       TransactionService.create!(order, e.body)
       Rails.logger.error("Could not reject order #{order.id}: #{e.message}")
+      raise e
     end
     order
   end
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -73,16 +73,18 @@ module OrderService
 
   def self.reject!(order)
     Order.transaction do
-      order.reject!
-      refund = PaymentService.refund_charge(order.external_charge_id)
-      transaction = {
-        external_id: order.external_charge_id,
-        amount_cents: refund.amount,
-        transaction_type: Transaction::REFUND,
-        status: Transaction::SUCCESS
-      }
-      TransactionService.create!(order, transaction)
-      order.save!
+      order.with_lock do
+        order.reject!
+        refund = PaymentService.refund_charge(order.external_charge_id)
+        transaction = {
+          external_id: order.external_charge_id,
+          amount_cents: refund.amount,
+          transaction_type: Transaction::REFUND,
+          status: Transaction::SUCCESS
+        }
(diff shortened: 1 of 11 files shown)


bhoggard:
I would like to wait to review this after the new locking code is in place.

williardx:
https://github.com/artsy/exchange/pull/105 adds locking to `submit` and `approve`, and I just added locking for `reject`. It also fixes some Rubocop errors that are happening here, so once we merge in that PR I'll rebase.

app/services/transaction_service.rb
```diff
@@ -1,27 +1,16 @@
...
-    order.transactions.create!(
-      external_id: charge.id,
-      source_id: charge.source,
-      destination_id: charge.destination,
-      amount_cents: charge.amount,
-      failure_code: charge.failure_code,
-      failure_message: charge.failure_message,
-      transaction_type: charge.transaction_type,
-      status: 'success'
+      transaction.slice(
```
bhoggard: Nice

app/services/order_service.rb
```diff
@@ -62,12 +72,23 @@ def self.fulfill_at_once!(order, fulfillment, by)
   end
 
   def self.reject!(order)
-    Order.transaction do
+    order.with_lock do
       order.reject!
+      refund = PaymentService.refund_charge(order.external_charge_id)
+      transaction = {
+        external_id: order.external_charge_id,
```
ashkan18: i think this should be `refund.id` Stripe gives us a refund id back here https://stripe.com/docs/api#refund_object

williardx: Yeah, I was a little up in the air on this one -- agreed `refund.id` is something we should store but there's otherwise no obvious place to store which charge it's related to in the current transaction schema.

ashkan18: so an `order` has many `transactions` and when we add a transaction we already assign it to the order here https://github.com/artsy/exchange/blob/master/app/services/transaction_service.rb#L16. And order `external_charge_id` should be same as this `external_id` right? I _think_ we should be good in setting it `refund.id` here.

ashkan18: awesome! 🙌

jonallured:
LGTM, merging!

PR #105, opened 2018-08-15
Lock order before submitting or approving

williardx:
### Problem
It's currently possible to enter a race condition and submit or approve an order twice, resulting in duplicate transactions and Stripe calls.

### Solution
Lock the order when submitting or approving at the database level.

### What changed
- Wrapped order submit and approve code in `order.with_lock`
- Created a method for constructing charge parameters  in `OrderSubmitService.submit!`, mostly to satisfy Rubocop.

### DIFF
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -33,14 +33,12 @@ module OrderService
   end
 
   def self.approve!(order, by: nil)
-    Order.transaction do
-      order.with_lock do
-        order.approve!
-        charge = PaymentService.capture_charge(order.external_charge_id)
-        TransactionService.create_success!(order, charge)
-        order.save!
-        PostNotificationJob.perform_later(order.id, Order::APPROVED, by)
-      end
+    order.with_lock do
+      order.approve!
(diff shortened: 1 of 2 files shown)


app/services/order_service.rb
```diff
@@ -34,11 +34,13 @@ def self.set_shipping!(order, attributes)
 
   def self.approve!(order, by: nil)
     Order.transaction do
-      order.approve!
-      charge = PaymentService.capture_charge(order.external_charge_id)
-      TransactionService.create_success!(order, charge)
-      order.save!
-      PostNotificationJob.perform_later(order.id, Order::APPROVED, by)
+      order.with_lock do
```
jonallured: Turns out `with_lock` includes a transaction:

https://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html#method-i-with_lock

So, you can bump this back out to replace that `Order.transaction` line. 🎉

williardx: Oh nice! Thanks, I'll change that.

williardx:
Cool! Going to go ahead and merge

PR #106, opened 2018-08-15
Adds shipping name to db, GraphQL schema

ashfurrow:
This PR addresses [PURCHASE-377](https://artsyproduct.atlassian.net/browse/PURCHASE-377), adding `shipping_name` to the Order db table and the various parts of the GraphQL schema.

~This PR is a work-in-progress because I have not yet tested it locally (but unit tests pass!).~

Here's it working in action:

screen shot 2018-08-16 at 10 01 55 am

Once this is merged, I will send a PR to metaphysics to update its copy of the Exchange schema.

### DIFF
--- a/app/graphql/types/order_type.rb
+++ b/app/graphql/types/order_type.rb
@@ -9,6 +9,7 @@ class Types::OrderType < Types::BaseObject
   field :credit_card_id, String, null: true
   field :state, Types::OrderStateEnum, null: false
   field :currency_code, String, null: false
+  field :shipping_name, String, null: false
   field :shipping_address_line1, String, null: true
   field :shipping_address_line2, String, null: true
   field :shipping_city, String, null: true
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -66,7 +66,7 @@ class Order < ApplicationRecord
 
   def shipping_info?
     fulfillment_type == PICKUP ||
-      (fulfillment_type == SHIP && %i[shipping_address_line1 shipping_city shipping_country shipping_postal_code].all? { |sh_field| send(sh_field).present? })
+      (fulfillment_type == SHIP && complete_shipping_details?)
   end
 
   def payment_info?
@@ -107,5 +107,9 @@ class Order < ApplicationRecord
       self.state = machine.state
     end
     machine
+  end
+
+  def complete_shipping_details?
+    [shipping_name, shipping_address_line1, shipping_city, shipping_country, shipping_postal_code].all?(&:present?)
   end
 end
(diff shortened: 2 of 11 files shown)


app/models/order.rb
```diff
@@ -66,7 +66,7 @@ def seller_total_cents
 
   def shipping_info?
     fulfillment_type == PICKUP ||
-      (fulfillment_type == SHIP && %i[shipping_address_line1 shipping_city shipping_country shipping_postal_code].all? { |sh_field| send(sh_field).present? })
+      (fulfillment_type == SHIP && %i[shipping_name shipping_address_line1 shipping_city shipping_country shipping_postal_code].all? { |sh_field| send(sh_field).present? })
```
jonallured: There's a refactoring opportunity here, I think. I wish this looked more like this:

```ruby
class Order
  # ...
  def shipping_info?
    fulfillment_type == PICKUP || (fulfillment_type == SHIP && complete_shipping_details?)
  end

  private

    %i[shipping_name shipping_address_line1 shipping_city shipping_country shipping_postal_code].all? { |sh_field| send(sh_field).present? }
  end
end
```

What do you think?

ashfurrow: I think that makes sense, I'll make that change 👍

ashfurrow: Dunno! Git history doesn't show any obvious reason. I'll include that in the refactor 👍

ashfurrow: Hmm! There must be a Ruby thing I haven't learned yet (or learned and forgot) because without `send` the tests fail. Looking into it more...

ashfurrow: I paired with Jon on this briefly and my code was failing locally due to the following difference:

```diff
-  %i[shipping_name shipping_address_line1 shipping_city...
+  [shipping_name, shipping_address_line1, shipping_city...
```

ashkan18:
Nice! 🏅

ashfurrow:
Thanks! Someone want to smash that green button? 💚

ashkan18:
i'd be honored 👍

app/graphql/types/order_type.rb
```diff
@@ -9,6 +9,7 @@ class Types::OrderType < Types::BaseObject
   field :credit_card_id, String, null: true
   field :state, Types::OrderStateEnum, null: false
   field :currency_code, String, null: false
+  field :shipping_name, String, null: false
```
sweir27: I believe this should be `null: true` since an order won't always have to have a `shipping_name` (i.e. when the order is created it won't have this set).

jonallured: Yeah, nice catch 👍

jonallured: This will also be null for pickup orders, no?

sweir27: Oh, yes! In that case null forever

PR #107, opened 2018-08-16
Updates setup instructions

ashfurrow:
I ran into a few issues setting up Exchange locally, so this PR updates the readme and example `.env` file to something that would've helped me at the time. 

Most problematic was the `GRAVITY_API_ROOT` env var, which would lead to 404's on the following endpoint when creating an order:

```
https://stagingapi.artsy.net/api/api/v1/artwork/581b69c5a09a6724180021f3
```

Double `/api` path – classic!


williardx:
💯Thanks @ashfurrow!

PR #108, opened 2018-08-16
Use ISO 8601 for DateTimeType

starsirius:
In https://github.com/artsy/volt/issues/3117, we noticed that exchange doesn't return timestamps in ISO 8601, and that caused metaphysics to parse the time incorrectly and present wrong timestamps to clients.

This make `Types::DateTimeType` consistent to return timestamps in ISO 8601, as in Gravity.

### DIFF
--- a/app/graphql/types/date_time_type.rb
+++ b/app/graphql/types/date_time_type.rb
@@ -1,13 +1,3 @@
-class Types::DateTimeType < Types::BaseScalar
-  description 'DateTime'
-
-  def self.coerce_input(input_value, _context)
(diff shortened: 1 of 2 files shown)


starsirius:
The spec [failed](https://circleci.com/gh/artsy/exchange/457?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) as expected.

screen shot 2018-08-16 at 11 01 14 am

jonallured:
Thank you so much for both figuring this out and fixing it!! ❤️ ❤️ ❤️

PR #109, opened 2018-08-16
remove Events:: namespace from OrderEvent call

erikdstock:
`PostNotificationJob#perform` calls `Events::OrderEvent#post`, but there is no `Events` namespace. As a result AMQP jobs are quietly retrying.

https://github.com/artsy/exchange/blob/4a38e435751b25553c5a5784a8e81ef0c48ea7ee/app/events/order_event.rb#L1

### DIFF
--- a/app/jobs/post_notification_job.rb
+++ b/app/jobs/post_notification_job.rb
@@ -3,6 +3,6 @@ class PostNotificationJob < ApplicationJob
 
   def perform(order_id, action, user_id)
     order = Order.find(order_id)
-    Events::OrderEvent.post(order, action, user_id)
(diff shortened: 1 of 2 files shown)


joeyAghion:
Not really trivial! Could use a spec, since apparently this code isn't traversed by existing tests.

erikdstock:
👍 - will add test.

jonallured:
Awesome, and thanks for adding the spec!! 🏆 

merge on green

williardx:
Thanks @erikdstock!

jonallured:
Oh hrm, I wonder why merge on green didn't work. 🤔

PR #110, opened 2018-08-16
Don't cache artwork responses and delete `ArtworkService`

williardx:
### Problem
Whenever we set an artwork location in CMS we have to wait ~10 minutes for Exchange's cache to expire before we can view the updated artwork.

### Solution
Having multiple caches that have different versions of the same data leads to confusion. I don't know of a compelling reason to cache artwork responses in Exchange right now, so we can just remove the caching there.

### What changed
- Uses uncached `GravityService.get_artwork` instead of cached `ArtworkService.find`
- Deletes now-deprecated `ArtworkService` module


app/services/artwork_service.rb
```diff
@@ -1,20 +0,0 @@
...
-                              dates external_id display_price_range confidential_notes created_at inventory_id published_changed_at publish_change_requested_at secondary_market deleted dimensions_string].freeze
-  def self.find(artwork_id)
-    find!(artwork_id)
-  rescue Adapters::GravityError
-    nil
-  end
-
-  def self.find!(artwork_id)
-    Rails.cache.fetch("gravity_artwork_#{artwork_id}", expires_in: Rails.application.config_for(:gravity)['artwork_cache_in_seconds']) do
-      Adapters::GravityV1.request("/artwork/#{artwork_id}?include_deleted=true")
```
williardx: One thing I'm not clear on here is why we have `include_deleted=true` in this request to get an artwork but we don't have it in other requests. I'm not sure if we need it -- the only place we were using this call was in `shipping_service.rb`, so it doesn't seem important. It might also be a moot point with upcoming artwork versions work. I'm keeping it in just in case for now, but if someone has more clarity here I'd be glad to remove it.

joeyAghion: Looks like it was added early on (https://github.com/artsy/exchange/pull/21) and maybe was just trying to be as general-purpose as possible. There are legitimate cases for internal, trusted systems to load data about deleted artworks, but I don't think calculating shipping is one of them unless it might recur later on in the life of an order.

app/services/gravity_service.rb
```diff
@@ -27,8 +27,8 @@ def self.get_credit_card(credit_card_id)
     raise Errors::OrderError, e.message
   end
 
-  def self.get_artwork(artwork_id)
-    Adapters::GravityV1.request("/artwork/#{artwork_id}")
+  def self.get_artwork(artwork_id, include_deleted: false)
+    Adapters::GravityV1.request("/artwork/#{artwork_id}?include_deleted=#{include_deleted}")
```
joeyAghion: We should avoid manually URL-encoding parameters and building querystrings in multiple locations. You can actually call `.to_query` on any hash to build a properly encoded querystring, but even better would be to delegate this to Faraday's `get` method, which already accepts a hash of arguments that we're currently leaving empty:

https://github.com/artsy/exchange/blob/1500fe48ef990c0082c88351c0c3875bf4cf5231/lib/adapters/gravity_v1.rb#L7

joeyAghion:
👍 Lesson learned!

app/services/gravity_service.rb
williardx: Nice! Noted for next time.

app/services/artwork_service.rb
williardx: Cool, thanks -- in theory the artwork could be deleted between creating an order and setting shipping but I'd question whether or not the right the correct behavior here would be to allow the user to continue on with that deleted artwork. AFAIK we don't have plans for setting shipping after an order has been submitted, so I'll take this out for now.

jonallured:
LGTM!

PR #112, opened 2018-08-17
Refactor state methods to get a block and also save the order

ashkan18:
# Problem
With current code, our service layer code is very fragile in terms of we need to first call state change method first, then do some stuff and then save.
```ruby
order.with_lock do
  order.approve!
  charge = PaymentService.capture_charge(order.external_charge_id)
  TransactionService.create_success!(order, charge)
  order.save!
end
```

# Possible Better Solution
Update state change methods to get a block, and also save the change after running the block if given. With this change the code above will look like
```ruby
order.approve! do
  charge = PaymentService.capture_charge(order.external_charge_id)
  TransactionService.create_success!(order, charge)
end
```

# Change
Updated state changes method to get an optional block and also save the order changes afterwards.

# Few Things To Consider
1) Also moved the order locking logic in the state change methods, im not 100% about this but might be ok.
2) There is a potential issue here when posting events (and generally doing things in the background within the block) in 👇 case
```ruby
order.approve! do
  charge = PaymentService.capture_charge(order.external_charge_id)
  TransactionService.create_success!(order, charge)
  PostNotificationJob.perform_later(order.id, Order::APPROVED, by)
end
```
order is going to be saved after the blocked we passed, but we queue a job to post events, if this job gets picked up before order being saved, it may pick the old order. Once solution could be moving post event outside of the block potentially.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -43,8 +43,13 @@ class Order < ApplicationRecord
   scope :active, -> { where(state: [SUBMITTED, APPROVED]) }
 
   ACTIONS.each do |action|
-    define_method "#{action}!" do
-      state_machine.trigger!(action)
+    define_method "#{action}!" do |&block|
+      with_lock do
+        state_machine.trigger!(action)
+        block.call if block.present?
+        save!
+        self
+      end
     rescue MicroMachine::InvalidState
       raise Errors::OrderError, "Invalid action on this #{state} order"
     end
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -34,14 +34,11 @@ module OrderService
   end
 
   def self.approve!(order, by: nil)
-    order.with_lock do
-      order.approve!
+    order.approve! do
       charge = PaymentService.capture_charge(order.external_charge_id)
       TransactionService.create_success!(order, charge)
-      order.save!
       PostNotificationJob.perform_later(order.id, Order::APPROVED, by)
     end
-    order
   rescue Errors::PaymentError => e
     TransactionService.create_failure!(order, e.body)
     Rails.logger.error("Could not approve order #{order.id}: #{e.message}")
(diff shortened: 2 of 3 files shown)


app/services/order_service.rb
```diff
@@ -34,14 +34,11 @@ def self.set_shipping!(order, attributes)
   end
 
   def self.approve!(order, by: nil)
-    order.with_lock do
-      order.approve!
+    order.approve! do
       charge = PaymentService.capture_charge(order.external_charge_id)
       TransactionService.create_success!(order, charge)
-      order.save!
       PostNotificationJob.perform_later(order.id, Order::APPROVED, by)
```
starsirius: Moving the async notification out of the block makes sense to me and should address the race condition you mentioned. I'm not sure exactly the benefit of wrapping it in the transaction, given it's not really rollback-able.

app/models/order.rb
```diff
@@ -43,8 +43,13 @@ class Order < ApplicationRecord
   scope :active, -> { where(state: [SUBMITTED, APPROVED]) }
 
   ACTIONS.each do |action|
-    define_method "#{action}!" do
-      state_machine.trigger!(action)
+    define_method "#{action}!" do |&block|
+      with_lock do
+        state_machine.trigger!(action)
+        block.call if block.present?
+        save!
```
starsirius: I like the direction of avoiding us from making mistakes! 👍 

I wonder if the separation of in-memory manipulation and db persistence will lead to unintentional behaviors:

```ruby
with_lock do
  state_machine.trigger!(action)   # => in-memory order state has changed, but hasn't been persisted to db
  block.call if block.present?     # => do whatever we want, but in a situation that state in memory and db are out-of-sync
  save!                            # => persist state change in db now
end
```

app/services/order_service.rb
williardx: This makes me think that once we're done with the Stripe call we should consider the order transaction complete regardless of `TransactionService`, `PostNotificationJob` or other future post-transaction tasks we need to do. It wouldn't be good for Stripe and our database to have conflicting information about the state of an order because `PostNotificationJob` failed and caused everything to rollback.

ashkan18: both good points, i'll move it to outside of the block.

ashkan18:
addressed all comments, should be 👍

williardx:
Thanks @ashkan18!

PR #113, opened 2018-08-17
Expose state_expires_at field for OrderEvent

bhoggard:
https://github.com/artsy/pulse/pull/229 depends on this.

### DIFF
--- a/app/events/order_event.rb
+++ b/app/events/order_event.rb
@@ -13,32 +13,30 @@ class OrderEvent < Events::BaseEvent
   end
 
   def properties
-    Time.use_zone('Eastern Time (US & Canada)') do
-      {
-        code: @object.code,
-        created_at: @object.created_at,
-        currency_code: @object.currency_code,
-        items_total_cents: @object.items_total_cents,
-        shipping_total_cents: @object.shipping_total_cents,
-        tax_total_cents: @object.tax_total_cents,
-        buyer_total_cents: @object.buyer_total_cents,
-        commission_fee_cents: @object.commission_fee_cents,
-        transaction_fee_cents: @object.transaction_fee_cents,
-        seller_total_cents: @object.seller_total_cents,
-        line_items: @object.line_items.map { |li| line_item_detail(li) },
-        partner_id: @object.partner_id,
-        shipping_name: @object.shipping_name,
-        shipping_address_line1: @object.shipping_address_line1,
-        shipping_address_line2: @object.shipping_address_line2,
-        shipping_city: @object.shipping_city,
-        shipping_country: @object.shipping_country,
-        shipping_postal_code: @object.shipping_postal_code,
-        shipping_region: @object.shipping_region,
-        state: @object.state,
-        updated_at: @object.updated_at,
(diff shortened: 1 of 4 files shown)


app/events/order_event.rb
```diff
@@ -12,30 +12,37 @@ def subject
     }
   end
 
+  def confirmation_deadline(order)
+    (order.updated_at + 48.hours).iso8601
```
erikdstock: Since this could apply to any order event I wonder if we should use a field separate from `updated_at`. We could try to be clever about it, like only including a value if the event was Order::SUBMITTED- not sure if that covers all cases either, though. I wonder if it would be best to ultimately have a record of when the purchase was submitted, confirmed, etc.

williardx: What confirmation deadline is this? Does it coincide with `Order::SUBMITTED` expiring?

bhoggard: Yes.

bhoggard: Maybe we should set a specific `submitted_date` on the order?

bhoggard: I didn't know about that. Great!

app/events/order_event.rb
```diff
@@ -13,29 +13,32 @@ def subject
...
-      shipping_address_line1: @object.shipping_address_line1,
-      shipping_address_line2: @object.shipping_address_line2,
-      shipping_city: @object.shipping_city,
-      shipping_country: @object.shipping_country,
-      shipping_postal_code: @object.shipping_postal_code,
-      shipping_region: @object.shipping_region,
-      state: @object.state,
-      updated_at: @object.updated_at
-    }
+    Time.use_zone('Eastern Time (US & Canada)') do
```
starsirius: I'm not sure if Exchange, as an API, should worry about time zone. I feel always returning UTC is more predictable from a client's perspective, and clients, e.g. Pulse, can do any desired formatting.

spec/events/order_event_spec.rb
```diff
@@ -29,7 +31,7 @@
       Fabricate(:line_item, price_cents: 100, order: order)
     ]
   end
-  let(:event) { OrderEvent.new(user: user_id, action: Order::SUBMITTED, model: order) }
+  let(:event) { OrderEvent.new(user: user_id, action: Order::SUBMITTED, model: order.reload) }
```
starsirius: Was the `reload` necessary because of the `order.save!` below?

app/events/order_event.rb
ashkan18: good catch! missed this one. yes i agree, makes sense to just return the timestamp and not deal with timezone here.

bhoggard:
I added a commit to remove the time zone. As far as I can tell, we don't need to modify https://github.com/artsy/pulse/pull/229 because it was already transforming the field to Eastern time.

spec/events/order_event_spec.rb
bhoggard: I thought it was, but it's working OK without it. I'll remove it.

starsirius:
Merging! Thanks, @bhoggard!! 🥇

PR #114, opened 2018-08-17
Raise error when missing shipping info

ashkan18:
# Problem
When `domestic_shipping_fee_cents` and `international_shipping_fee_cents` are not set we were assuming that means they are 0 but it actually means partner never confirmed/set the shipping cost on this artwork and we should raise error.

### DIFF
--- a/app/services/shipping_service.rb
+++ b/app/services/shipping_service.rb
@@ -20,10 +20,10 @@ module ShippingService
   end
 
   def self.calculate_domestic(artwork)
-    artwork[:domestic_shipping_fee_cents] || 0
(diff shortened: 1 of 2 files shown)


app/services/shipping_service.rb
```diff
@@ -20,10 +20,10 @@ def self.domestic?(artwork_location, shipping_country)
   end
 
   def self.calculate_domestic(artwork)
-    artwork[:domestic_shipping_fee_cents] || 0
+    artwork[:domestic_shipping_fee_cents] || raise(Errors::OrderError, 'Artwork is missing shipping fee.')
```
starsirius: Not sure if we want to be more specific about domestic vs. international in the message, but can follow up if needed.

PR #115, opened 2018-08-17
Schedule job to expire orders on order creation, submission and approval

williardx:
### Problem
Orders in pending, submitted and approved states should automatically expire if they remain in those states for too long. Pending orders should be abandoned and submitted and approved orders should be rejected.

### Solution
Schedule a job to run at the order expiration time given by `order.state_expires_at` that abandons pending orders and rejects submitted / approved orders. The job short circuits if the order is in a different state or if it's before the expiration time.

### What changed
- Adds `ExpireOrderJob` which abandons pending orders and rejects submitted / approved orders if they're in those states after their expiration time.
- Schedules running an `ExpireOrderJob` job at the expiration time specified by `order.state_expires_at` on order creation, submission and approval.
- Adds Timecop for tests


app/jobs/expire_order_job.rb
```diff
@@ -0,0 +1,10 @@
+class ExpireOrderJob < ApplicationJob
+  queue_as :default
+
+  def perform(order_id, state)
+    order = Order.find(order_id)
+    return unless order.state == state && Time.new > order.state_expires_at
```
joeyAghion: This could use `Time.now` to be more explicit and should probably be `>=` (sidekiq is _fast_).

spec/controllers/api/requests/approve_order_mutation_request_spec.rb
```diff
@@ -34,6 +34,11 @@
         }
       }
     end
+
+    before do
+      ActiveJob::Base.queue_adapter = :test
```
joeyAghion: This looks like it might also affect any specs run _after_ this context. It might need to be unset in an `around` or `after` block, or just be set to the default in a global `before`.

williardx:
This is failing due to some Rubocop errors that will clear up once https://github.com/artsy/exchange/pull/104 is merged in and I rebase.

app/jobs/expire_order_job.rb
```diff
@@ -0,0 +1,10 @@
...
+    OrderService.abandon!(order) if order.state == Order::PENDING
+    OrderService.reject!(order) if order.state == Order::SUBMITTED || order.state == Order::APPROVED
```
joeyAghion: I wonder about some of this complexity, and whether it might be simpler to just queue separate, simple `reject_submitted_order`, `reject_approved_order`, and `abandon_pending_order` jobs. That would be less meta but there'd be some repetition among the implementations.

This is pretty simple for now, so we could just reevaluate if the behaviors for the different conditions start to diverge.

app/jobs/expire_order_job.rb
williardx: Sounds good!

spec/controllers/api/requests/approve_order_mutation_request_spec.rb
williardx: Is there any case in which we'd want this to be something other than `:test` when doing tests?

app/jobs/expire_order_job.rb
williardx: Agreed, let's revisit if things get more complicated.

spec/controllers/api/requests/approve_order_mutation_request_spec.rb
joeyAghion: Yes. It can be convenient for background jobs to be handled "inline" during testing. So, e.g., you can simulate an API request and then make assertions about the state of its work. I believe this is the default, though I'm not too familiar with the sidekiq/activejob interface.

joeyAghion: Actually, it looks like you might be able to use ActiveJob's built-in testing helpers to make the assertion below without needing to override the `queue_adapter`: https://edgeguides.rubyonrails.org/testing.html#custom-assertions-and-testing-jobs-inside-other-components

app/jobs/expire_order_job.rb
ashkan18: super #minor but ruby case statement might read better here:
```ruby
case order.state
when Order::PENDING
  OrderService.abandon!(order)
when Order::SUBMITTED, Order::APPROVED
  OrderService.reject!(order)
end
```
on a separate note, now that i think about it more, i feel like rejecting an Approved order might be a more significant action since it includes actual money transfer of refund and we may need to notify everyone about it. Not related to this PR but in general it scares me a little 😱

app/jobs/expire_order_job.rb
```diff
@@ -0,0 +1,10 @@
+class ExpireOrderJob < ApplicationJob
+  queue_as :default
+
+  def perform(order_id, state)
+    order = Order.find(order_id)
```
jonallured: Something I always think about in these jobs is what to do if I get an `order_id` that's invalid. The call to `find` here will raise `ActiveRecord::RecordNotFound` if there is no Order by that id. I usually use `find_by` and the safe navigator:

```ruby
def perform(order_id, state)
  order = Order.find_by(id: order_id)
  return unless order&.state == state && Time.now >= order&.state_expires_at
  # ...
end
```

Another approach would be to rescue that exception or simply allow the retry behavior. Just wanted to point it out so that we're being intentional here. ❤️

spec/controllers/api/requests/approve_order_mutation_request_spec.rb
jonallured: This is how I dealt with features wanting inline jobs on Bearden:

https://github.com/artsy/bearden/blob/master/spec/rails_helper.rb#L38-L41

app/jobs/expire_order_job.rb
ashkan18: since we don't delete Orders anywhere (yet!) i think a retry in this case might be actually useful. It will make it show up in Sidekiq dashboard and etc. 

Alternative would be using `discard_on` https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html
and in the block do some sort of notification, something like newrelic error or rails error log or something.

app/jobs/expire_order_job.rb
williardx: I agree that we'll want some more nuance around the failure states of our orders but I'm not exactly sure what that nuance is yet. We should get more clarity with product / ops around what they expect to do in these cases and what information they'll need.

And I like the case statement here, I'll update it!

app/jobs/expire_order_job.rb
```diff
@@ -0,0 +1,14 @@
...
+    return unless order.state == state && Time.now >= order.state_expires_at
+    case order.state
+    when Order::PENDING
+      OrderService.abandon!(order)
+    when Order::SUBMITTED, Order::APPROVED
+      OrderService.reject!(order)
```
ashkan18: looking at https://www.notion.so/artsy/37c311363ef046c3aa546047e60cc58a?v=de68d5bbc30748f88b0d92a059bc0ba8 i think we may want to have separate methods for `submitted` and `approved` cases and add `USER_LAPSE``PARTNER_LAPSE` to our states. if you prefer we can do that in a separate PR too if other PRs are waiting for this one to get merged.

spec/services/order_submit_service_spec.rb
```diff
@@ -56,6 +55,12 @@
           expect(PostNotificationJob).to have_been_enqueued
         end
 
+        it 'queues a job for rejecting the order when it expires' do
+          job = ActiveJob::Base.queue_adapter.enqueued_jobs.detect { |j| j[:job] == ExpireOrderJob }
+          expect(job).to_not be_nil
+          expect(job[:at].to_i).to eq order.reload.state_expires_at.to_i
```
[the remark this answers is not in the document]
williardx: Ah yep, that's a good idea

app/jobs/expire_order_job.rb
williardx: Are `USER_LAPSE` and `PARTNER_LAPSE` really the states we want though? They sound more like context around an order being expired rather than a state in their own right. I could see "expired" being a separate state that we might want to differentiate from rejected, though.

joeyAghion: I agree that we might be conflating states with _transitions_. "Lapsed" or "expired" feels different from rejected, but whether it's user- or partner-lapsed sort of depends on the previous order data and state.

williardx: Since there's more to discuss here let's leave this change for a separate PR.

PR #116, opened 2018-08-20
@ashkan18 => allow shipping_name to be null in graphql schema

sweir27:
#minor but when orders are created, they won't have this field set (so we are getting errors on the initial mutation for a null field not being nullable).


ashkan18:
good catch 🙌

ashkan18:
we may need to do a quick https://github.com/artsy/exchange#did-you-change-graphql-schema

ashfurrow:
Yeah, thanks Sarah! Sorry about that.

PR #117, opened 2018-08-21
switch from bitnami/ruby:2.5.1 to official ruby:2.5.1 for performance

bhoggard:
After seeing @izakp's comments, I thought I should update Exchange too.


ashkan18:
👍 
cc: @williardx since we were trying to do this yesterday too :)

izakp:
Thanks for following up on this!

PR #118, opened 2018-08-21
Adding new state for ignored submitted orders

ashkan18:
# Problem
This is mainly follow up on comments in #115 where we decided to have a separate state for submitted orders that get expired.

I also talked to @joanna20 about not automatically refund already approved orders (orders that were approved by partners but haven't been fulfilled yet). We decided that we are not going to automatically refund them but we may want to notify internal admins to contact partners and follow up (or eventually send follow up emails).

# Changes
I added new `SELLER_LAPSED` state (which im still struggling with the name a little) based on https://www.notion.so/artsy/37c311363ef046c3aa546047e60cc58a?v=de68d5bbc30748f88b0d92a059bc0ba8. We talked a bit about having `expired` state for all possible reasons an order goes to expired, but Im not opposed to have more specific state in this case mainly knowing we are going to post events and for the consumer of these events they may want to defer between an order expired because of seller never responded or order expired because never fulfilled or other possible expired cases.

I also renamed `ExpireOrderJob` to `OrderFollowUpJob` since in Approve case we are actually not expiring anything and we just want to post an event. Let me know what you think.


joanna20:
@ashkan18 does this state apply to either step - the order confirm or the order fulfill step? Can an order get into this state if the seller doesn't confirm within 2 days, and also if they don't ship within 7?

ashkan18:
@joanna20 no actually, sorry for my bad description, it only applies to submitted orders that are not approved/rejected by partners within 2 days. 
For the case of approved orders that are not fulfilled state will stay approved but im going to post an event so we can nudge admins or possibly partners about it.

ashkan18:
cc: @madeleineb

app/services/order_service.rb
```diff
@@ -69,29 +69,42 @@ def self.fulfill_at_once!(order, fulfillment, by)
...
+
   def self.abandon!(order)
     order.abandon!
   end
 
   def self.valid_currency_code?(currency_code)
     Order::SUPPORTED_CURRENCIES.include?(currency_code.downcase)
   end
+
+  def self.refund(order)
```
williardx: We should be rescuing `PaymentError` here and logging an unsuccessful refund if it happens

williardx: ah! I see, you're still catching it in `reject!`

spec/jobs/order_follow_up_job_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'rails_helper'
+require 'timecop'
+
+describe OrderFollowUpJob, type: :job do
+  let(:state) { Order::PENDING }
+  let(:order) { Fabricate(:order, state: state) }
+  describe '#perform' do
+    context 'with an order in the same state after its expiration tim' do
```
williardx: Typo - should be "time"

ashkan18:
As part of this also updated shipping expiration to 7 days, cc: @madeleineb @joanna20 

@joeyAghion going to merge this, I'll address comments in separate PR.

PR #120, opened 2018-08-22
Allow sales_admin to access all orders

bhoggard:
Slack discussion is [here](https://artsy.slack.com/archives/C9ZJYFDNF/p1534970696000100).

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -47,7 +47,7 @@ class Types::QueryType < Types::BaseObject
   end
 
   def partner_permitted?(id)
-    context[:current_user][:partner_ids].include?(id)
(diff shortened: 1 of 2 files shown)


ashkan18:
makes sense! can you add a spec for it to confirm?

app/graphql/types/query_type.rb
```diff
@@ -47,7 +47,7 @@ def user_permitted?(id)
   end
 
   def partner_permitted?(id)
-    context[:current_user][:partner_ids].include?(id)
+    context[:current_user][:partner_ids].include?(id) || context[:current_user][:roles].include?('sales_admin')
```
ashkan18: this may change later in https://github.com/artsy/exchange/pull/121, we may want to add a separate method for checking `admin_permitted?` but this is 👍  for now.

PR #123, opened 2018-08-23
Fix null fulfillment_type issue

ashkan18:
Followup on #122  which was followup on #119 
Basically here make sure we handle `null` for `fulfillment_type`.


orta:
This is because when you first create an order, you haven't made a choice yet - making that choice sets the type 👍

PR #124, opened 2018-08-23
Record timestamp when order enters a new state

williardx:
This PR addresses [PLATFORM-773](https://artsyproduct.atlassian.net/browse/PLATFORM-773) and updates a timestamp when an order becomes abandoned, submitted, approved, rejected or fulfilled.

Metaphysics PR forthcoming!

### What Changed
- Adds 5 new datetime columns to the orders table: `abandoned_at`, `submitted_at`, `approved_at`, `rejected_at` and `fulfilled_at`.
- Sets `order.$STATE_at` for each state except pending whenever the state changes.
- Adds `submitted_at`, `approved_at`, `rejected_at` and `fulfilled_at` to the GraphQL schema.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -90,7 +90,7 @@ class Order < ApplicationRecord
 
   def update_state_timestamps
     self.state_updated_at = Time.now.utc
-    self.send("#{self.state}_at=", Time.now.utc) unless self.state == Order::PENDING
+    send("#{state}_at=", Time.now.utc) unless state == Order::PENDING
     self.state_expires_at = STATE_EXPIRATIONS.key?(state) ? state_updated_at + STATE_EXPIRATIONS[state] : nil
   end
 
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -90,6 +90,7 @@ class Order < ApplicationRecord
 
   def update_state_timestamps
     self.state_updated_at = Time.now.utc
+    self.send("#{self.state}_at=", Time.now.utc) unless self.state == Order::PENDING
     self.state_expires_at = STATE_EXPIRATIONS.key?(state) ? state_updated_at + STATE_EXPIRATIONS[state] : nil
   end
 
--- a/spec/controllers/api/requests/approve_order_mutation_request_spec.rb
+++ b/spec/controllers/api/requests/approve_order_mutation_request_spec.rb
@@ -69,6 +69,7 @@ describe Api::GraphqlController, type: :request do
           expect(order.reload.state).to eq Order::APPROVED
           expect(order.reload.transactions.last.external_id).to eq uncaptured_charge.id
           expect(order.reload.transactions.last.transaction_type).to eq Transaction::CAPTURE
+          expect(order.reload.approved_at).not_to be_nil
         end.to change(order, :state_expires_at)
       end
 
--- a/app/graphql/types/order_type.rb
+++ b/app/graphql/types/order_type.rb
@@ -22,6 +22,10 @@ class Types::OrderType < Types::BaseObject
   field :state_updated_at, Types::DateTimeType, null: true
   field :state_expires_at, Types::DateTimeType, null: true
   field :line_items, Types::LineItemType.connection_type, null: true
+  field :submitted_at, Types::DateTimeType, null: true
+  field :approved_at, Types::DateTimeType, null: true
+  field :rejected_at, Types::DateTimeType, null: true
+  field :fulfilled_at, Types::DateTimeType, null: true
 
   def fulfillment
     # fulfillment is not a field on order so we have to resolve it here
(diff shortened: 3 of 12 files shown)


app/models/order.rb
```diff
@@ -90,6 +90,7 @@ def set_code
 
   def update_state_timestamps
     self.state_updated_at = Time.now.utc
+    send("#{state}_at=", Time.now.utc) unless state == Order::PENDING
```
joeyAghion: Just noting that there may be states that can be entered more than once (especially when offers happen, but maybe now also?), and the timestamp will be overwritten upon each transition.

spec/controllers/api/requests/approve_order_mutation_request_spec.rb
```diff
@@ -69,6 +69,7 @@
           expect(order.reload.state).to eq Order::APPROVED
           expect(order.reload.transactions.last.external_id).to eq uncaptured_charge.id
           expect(order.reload.transactions.last.transaction_type).to eq Transaction::CAPTURE
+          expect(order.reload.approved_at).not_to be_nil
```
joeyAghion: You can use `.to be_present` if you dislike `not_to` like me (purely FYI).

jonallured:
Another idea here is to use PaperTrail. @joeyAghion has a good point that when we hit use cases that allow an order to go back and forth between states, the timestamp fields added here won't tell the whole story about the states of an order. I took a quick look on staging to see what I could come up with, check this out:

```ruby
> order = Order.find 180
> order.state
=> "fulfilled"
> submitted_versions = order.versions.select { |v| v.event == 'update' && v.object['state'] == 'submitted' }
> submitted_versions.map { |v| v.object['state_updated_at'] }
=> ["2018-08-22T16:43:50.062Z", "2018-08-22T16:43:50.062Z"]
```

PaperTrail stores the before as documented in the table here:

https://github.com/paper-trail-gem/paper_trail#1c-basic-usage

So that means the versions matching my `select` above are the before states, not the changes - not ideal, but...I noticed that PaperTrail has an optional feature for storing the diff:

https://github.com/paper-trail-gem/paper_trail#3c-diffing-versions

I think it would probably be a good idea to add the `object_changes` column as they suggest.

Anyway, I guess why I'm interested in this approach is that we're already storing this info, so let's use it! If not for this, then why are we using PaperTrail on Exchange?!

ashkan18:
We had somewhat similar discussion about this when we were thinking about re-using orders in Gravity. I think we should handle the cases going back and forth between these states and store all of those state transitions. Knowing that there are already consumers for this data and we need to expose to outside world in GraphQL, im more for adding a separate model for capturing state changes. 

Im thinking of Papertrail as more of an audit thing and thinking this new `OrderStateHistory` first class model as something more proper to expose to outside world. 

If we found a use for knowing `latest_<state>_at` we can always denormalize it later.

williardx:
@ashkan18 is there a difference between this new `OrderStateHistory` model and having the `<state>_at` attributes on the order model point to a query executed through PaperTrail? I like @jonallured's point that PaperTrail is already doing most of the work we'd need for that.

FWIW I think we could also construct our order states so that they don't cycle, though maybe that's a topic better left to the offer modeling session.

jonallured:
> I think we should handle the cases going back and forth between these states and store all of those state transitions.

@ashkan18 maybe I'm missing something, but isn't this what PaperTrail does? Can you help me understand what your hesitation is for using PaperTrail?

jonallured:
Ok, gotcha, so you're thinking about how we'll expose the history of an order at the API-level. I think that we might be saying different things. I agree that I wouldn't want to see the PT versions exposed, I think there are better ways for consumers of the API to get at an order's history. So that topic is worth a separate conversation and needs some planning for sure.

What I'm trying to advocate for is that we don't store these state transitions twice in the database. If PT is already doing it, then let's use that to our advantage and fulfill our API contract using this data. Volt doesn't care about how Exchange comes up with the history of an order, only that it is exposed at the API level. PT or another db table are implementation details, ya know?

ashkan18:
Looking at original ticket https://artsyproduct.atlassian.net/browse/PLATFORM-773, it seems we are exposing this to show them in some UI somewhere. I was ultimately talking about how we expose them in API and yes PT doesn't sound like something we want to expose externally.

As for duplicate storing of the same data, im not totally against it if there is a good reason for it. Im thinking of PT as something that tracks all important models for our internal audit which can be possibly archived on S3 or something in the future if needed and potentially even purged from our databased for records older than something. 

We had some discussions around this before https://github.com/artsy/exchange/pull/49#discussion-diff-199203182R1.

jonallured:
Ok, I guess I've derailed this PR long enough! 😝 

I still don't fully understand why we'd start by duplicating data rather than reaching for that solution only if/when it's necessary, but it sounds like @ashkan18 feels strongly about this, so I'll back off. Thanks for hearing me out!! ❤️

williardx:
Both solutions here come with tradeoffs. If we use PaperTrail then we're tying our data and implementation to this gem in a way that's not strictly necessary, and if we don't then we're duplicating data and prematurely doing work / increasing complexity.

If PaperTrail is taken to purely be an audit log, though, then it does feel weird to reach in and use it like it's any other product data. Given that, having our own `StateHistory` model seems like it would ultimately be more flexible for the future, so I'm okay with doing that.

williardx:
Closing this PR out in favor of a new one

PR #125, opened 2018-08-23
Revert "Fix null fulfillment_type issue"

ashkan18:
Reverts artsy/exchange#123


orta:
If it was green before, it should be green after.

PR #127, opened 2018-08-23
Revert query back to previous shipping changes because of MP weirdness

ashkan18:
# Problem
We tried https://github.com/ashkan18/metaphysics/tree/update-shipping-fields to match the latest schema changes introduced in #119 but ran into weird issues when trying to get a connection of orders.

# solution
This PR reverts the query part of things while for mutation we are still keeping new style (see new style mutation in #119)


ashkan18:
sorry was in a rush to revert to get staging in good state again, basically in MP we added this union https://github.com/ashkan18/metaphysics/blob/update-shipping-fields/src/schema/ecommerce/types/requested_fulfillment.ts to match exchange when getting order requested fulfillment. Things were working when we were trying to get one specific order in https://github.com/ashkan18/metaphysics/blob/update-shipping-fields/src/schema/ecommerce/order.js but when trying to get a list (connection) of orders using this https://github.com/ashkan18/metaphysics/blob/update-shipping-fields/src/schema/ecommerce/orders.js query, lets say we had 20 orders and one of them had shipping info, the whole thing would silently fail and return empty array of `edges` for `OrderConnection`. 
We verified that:
- Exchange was returning correct response
- This was only happening when there was at least one order with non-null shipping information
- Removing `requestedFulfillment` from https://github.com/ashkan18/metaphysics/blob/update-shipping-fields/src/schema/ecommerce/orders.js#L46 would fix the issue but then we wouldn't get fulfillment info

As far as i can tell, the type and union and all that works fine since everything worked perfectly when getting one order, so my guess is something is failing when using this union type in a connection.

PR #129, opened 2018-08-24
[COORDINATED MERGE] Bring back requestedFulfillmentUnion

ashkan18:
# Problem
Because of MP's issues in using `requestedFulfilmentUnion` when getting a connection of `Order`s we reverted the changes we made in #119 and few followups.
We now have  a solution for that issue and can bring this back.

** COORDINATED MERGE ** going to open a MP PR to match these changes, lets merge this with that one around the same time so staging MP -> Exchange stays in sync.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,9 @@ mutation($input: SetShippingInput!) {
           addressLine1
           country
         }
+        ... on Pickup {
+          fulfillmentType
+        }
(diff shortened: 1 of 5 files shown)


ashkan18:
MP change https://github.com/artsy/metaphysics/pull/1241

PR #130, opened 2018-08-26
Update inventory

ashkan18:
# Problem
We want to deduct inventory when a user _submits_ an order. We need to un-deduct inventory in reject , seller-lapse and other edge cases.

# Solution
We first deduct all the inventory, then we call Stripe to hold the charge. We need to handle following cases.

## Submit order
<table>
<tr>
<th> Scenario </th>
<th> What happens </th>
</tr>

<tr>
<td>Fails to deduct one of the line items</td>
<td>Keep track of all deducted line items and when one fails make sure we un-deduct only the already deducted ones</td>
</tr>

<tr>
<td>Stripe charge fails</td>
<td>We've already deducted inventory for all line items, we need to un-deduct all of them</td>
</tr>
</table> 

## Reject/Seller Lapse
Un-deduct all line items.

# Other Changes
Ended up making `OrderSubmitServer` actual class with instance methods which we can instantiate with an `order` and call `process!` on it. Provided a `call` class method that actually does that.

# Note
We could make Gravity calls less complicated if there was an endpoint in Gravity that could get list of inventory deductions and do all in one call (wonder if GraphQL could help here). This logic will be as complicated on Gravity's end (maybe more since there are no transactions there) but wanted to mention as an alternative.

### DIFF
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -69,7 +69,7 @@ module OrderService
 
   def self.reject!(order, by)
     order.reject! do
-      release_inventory_refund(order)
+      refund(order)
     end
     PostNotificationJob.perform_later(order.id, Order::REJECTED, by)
     order
@@ -81,7 +81,7 @@ module OrderService
 
   def self.seller_lapse!(order)
     order.seller_lapse! do
-      release_inventory_refund(order)
+      refund(order)
     end
     PostNotificationJob.perform_later(order.id, Order::SELLER_LAPSED)
     order
@@ -95,7 +95,7 @@ module OrderService
     Order::SUPPORTED_CURRENCIES.include?(currency_code.downcase)
   end
 
-  def self.release_inventory_refund(order)
+  def self.refund(order)
     refund = PaymentService.refund_charge(order.external_charge_id)
     transaction = {
       external_id: refund.id,
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -69,7 +69,7 @@ module OrderService
 
   def self.reject!(order, by)
     order.reject! do
-      refund(order)
+      release_inventory_refund(order)
     end
     PostNotificationJob.perform_later(order.id, Order::REJECTED, by)
     order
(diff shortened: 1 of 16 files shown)


orta:
Sorry, been doing some work on the [Peril dockerfile](https://hub.docker.com/r/dangersystems/peril/builds/) and it's temporarily breaking staging.

app/services/order_submit_service.rb
```diff
@@ -1,30 +1,62 @@
...
+class OrderSubmitService
+  def self.call!(order, by: nil)
+    new(order, by).process!
+  end
 
-    merchant_account = GravityService.get_merchant_account(order.partner_id)
-    credit_card = GravityService.get_credit_card(order.credit_card_id)
-    validate_credit_card!(credit_card)
-    charge_params = construct_charge_params(order, credit_card, merchant_account)
+  attr_accessor :order, :credit_card, :merchant_account, :partner
```
ashkan18: this mostly helps in testing

app/services/order_submit_service.rb
```diff
@@ -1,30 +1,62 @@
-module OrderSubmitService
-  def self.submit!(order, by: nil)
-    # verify price change?
-    raise Errors::OrderError, "Missing info for submitting order(#{order.id})" unless can_submit?(order)
+class OrderSubmitService
+  def self.call!(order, by: nil)
```
ashkan18: not so sure about this method's name

ashkan18: i wanted to provide a helper that encourages minimizing the time between instantiation and actual process since things can change in between but its definitely not forcing anything so i could remove.

app/services/order_service.rb
```diff
@@ -95,7 +95,7 @@ def self.valid_currency_code?(currency_code)
     Order::SUPPORTED_CURRENCIES.include?(currency_code.downcase)
   end
 
-  def self.refund(order)
+  def self.release_inventory_refund(order)
```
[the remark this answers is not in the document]
ashkan18: yeah not a big fan of this name 👍 will change

ashkan18:
alright going to merge this so we can try it on staging, @williardx feel free to comment and i'll address them on follow up PR 🙌

PR #134, opened 2018-08-28
Track order state history

williardx:
This is a follow up PR for https://github.com/artsy/exchange/pull/124, where we decided to create a new `StateHistory` model to keep track of the entire history of an order's state changes.

Addresses [PLATFORM-773](https://artsyproduct.atlassian.net/browse/PLATFORM-773).

### What changed
- Adds `StateHistory` model that's associated with `Order` for tracking state changes in an order. 
- On order creation and after every order action (submit, approve, reject, etc.) a new `StateHistory` record is created that records the time at which the order entered that state.
- Creates `submitted_at` and `approved_at` methods on `Order` that query `StateHistory` for the time at which an order became submitted and approved, respectively.

Metaphysics PR forthcoming.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -93,6 +93,14 @@ class Order < ApplicationRecord
     "Order #{id}"
   end
 
+  def submitted_at
+    state_histories.find_by(state: Order::SUBMITTED)&.updated_at
+  end
+
+  def approved_at
+    state_histories.find_by(state: Order::APPROVED)&.updated_at
+  end
+
   private
 
   def set_code
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -45,6 +45,7 @@ class Order < ApplicationRecord
   validates :state, presence: true, inclusion: STATES
 
   after_create :set_code
+  after_create :create_state_history
   before_save :update_state_timestamps, if: :state_changed?
   before_save :set_currency_code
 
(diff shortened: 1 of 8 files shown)


app/models/order.rb
```diff
@@ -90,6 +93,14 @@ def to_s
     "Order #{id}"
   end
 
+  def submitted_at
```
ashkan18: knowing that it's possible to get to these states multiple times (like admins manually switching states if needed), i'd say lets `sort` and `.first` and call this method `last_submitted_at` to clarify what we are actually doing.

app/models/order.rb
```diff
@@ -55,6 +57,7 @@ class Order < ApplicationRecord
       with_lock do
         state_machine.trigger!(action)
         save!
+        create_state_history
```
ashkan18: magical line...

app/models/order.rb
```diff
@@ -40,10 +40,12 @@ class Order < ApplicationRecord
 
   has_many :line_items, dependent: :destroy, class_name: 'LineItem'
   has_many :transactions, dependent: :destroy
+  has_many :state_histories, dependent: :destroy
 
   validates :state, presence: true, inclusion: STATES
 
   after_create :set_code
+  after_create :create_state_history
```
ashkan18: im ok with this callback here but since we all don't like callbacks and trying to avoid it, we could in this case just add this starting point history in `CreateOrder` service call too.

PR #136, opened 2018-08-29
Add support for totalCount on orders

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PLATFORM-835
From clients we sometimes want to show just the total count of specific query.

# Solution
Added a custom connection type for our `Order` model which defines `total_count` field and used it in returning `orders` query. You can ask for `totalCounts` now on the same level as edges.
```graphql
query($sellerId: String, $buyerId: String, $state: OrderStateEnum, $sort: OrderConnectionSortEnum) {
            orders(sellerId: $sellerId, buyerId: $buyerId, state: $state, sort: $sort, first: 2) {
              totalCount
              edges {
                node {
                  id
                }
              }
            }
          }
```
In sample ☝️ we'll only get first 2 orders (note `first: 2` in the query)back but we'll get total count of all orders.

### DIFF
--- a/app/graphql/types/order_connection_with_total_count_type.rb
+++ b/app/graphql/types/order_connection_with_total_count_type.rb
@@ -1,0 +1,14 @@
+class Types::OrderEdgeType < GraphQL::Types::Relay::BaseEdge
+  node_type(Types::OrderType)
+end
+
+class Types::OrderConnectionWithTotalCountType < GraphQL::Types::Relay::BaseConnection
+  edge_type(Types::OrderEdgeType)
+
+  field :total_count, Integer, null: false
+  def total_count
+    # - `object` is the Connection
+    # - `object.nodes` is the collection of Orders
+    object.nodes.size
+  end
+end
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -7,7 +7,7 @@ class Types::QueryType < Types::BaseObject
     argument :id, ID, required: true
   end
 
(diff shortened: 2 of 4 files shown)


jonallured:
I'm just starting to play with this, do you guys think this makes sense:

```
query($sellerId: String, $sellerType: String, $state: String, $sort: OrdersSortMethodType) {
  submittedCount:orders(sellerId: $sellerId, sellerType: $sellerType, state: "SUBMITTED") {
    totalCount
  }
  approvedCount:orders(sellerId: $sellerId, sellerType: $sellerType, state: "APPROVED") {
    totalCount
  }
  orders(sellerId: $sellerId, sellerType: $sellerType, state: $state, sort: $sort) {
    edges {
      node {
        id
        code
        itemsTotal
        state
        stateUpdatedAt
        stateExpiresAt
        requestedFulfillment {
          __typename
        }
        lineItems {
          edges {
            node {
              id
              editionSet {
                id
              }
              artwork {
                id
                inventoryId
                artist_names
              }
            }
          }
        }
      }
    }
  }
}
```

What I'm working on is adding open and ready counts to CMS in the /orders nav area.

Do you think it's fair for the query that gets the type of orders we're looking at to also have queries for the counts? I could make a separate request for this, but that seemed wasteful.

ashkan18:
This generally makes sense to me, although i may separate the query for getting actual results with the one that only need count. basically make 2 requests, one for list of orders another just for counts which would be the query you have there that only asks for `totalCount`.

PR #139, opened 2018-08-29
Fix compare to `nil` error

### LINKED ISSUE #138: Sidekiq errors
```2018-08-29T15:35:53.480Z 1 TID-gr4okbye1 WARN: ArgumentError: comparison of Time with nil failed
2018-08-29T15:35:53.480Z 1 TID-gr4okbye1 WARN: /app/app/jobs/order_follow_up_job.rb:6:in `>='```

ashkan18:
Fixes #138 

# Cause
When creating an order, we set the state to pending but we don't set `state_expires_at` so https://github.com/artsy/exchange/blob/master/app/jobs/order_follow_up_job.rb#L6 fails.

# Solution
Set it at order creation.

cc: @williardx


williardx:
Nice! Thanks for catching this @ashkan18  💯

ashkan18:
🙌

PR #140, opened 2018-08-29
Adds the buyer id to the event going out to the event service

orta:
Spotted this was missing when reading the code,


ashkan18:
🙌

PR #141, opened 2018-08-29
Use region codes when calculating sales tax for US/CA orders

williardx:
### Problem
TaxJar requires a region code (e.g. "FL" for "Florida") for state/provinces in US and CA-based orders. We use a free-text input field to collect the shipping region from the buyer, opening the possibility that the buyer might input something that TaxJar doesn't recognize.

### Solution
For orders being shipped to US/CA, parse the shipping region field using [Carmen](https://github.com/carmen-ruby/carmen) and return a two-letter region code or `nil` if the region can't be parsed.

A longer-term fix would be to use a more structured input method on the client side, but that's beyond immediate scope.

### What changed
- Adds Carmen gem to help translate shipping region names to codes
- Adds `parse_region` method to `SalesTaxService` that parses the free-text region input and converts to a region code if needed


app/services/sales_tax_service.rb
```diff
@@ -10,10 +11,16 @@ module SalesTaxService
 
   def self.calculate_total_sales_tax(order, fulfillment_type, shipping, shipping_total_cents)
     raise 'Dont know how to calculate tax for non-partner sellers' unless order.seller_type == Order::PARTNER
+    if fulfillment_type == Order::SHIP
```
ashkan18: do you think there is any benefit in moving line 14-19 to a separate private local method? or do this before hand in previous method call, mainly thinking of this method being super important and want to only see tax related stuff here, we could move this logic up all the way to actual mutation too if you think we are going to eventually remove it.

ashkan18: basically thinking of moving it all the way up to https://github.com/artsy/exchange/blob/master/app/graphql/mutations/set_shipping.rb#L12

PR #144, opened 2018-08-29
Rename and change submitted_at and approved_at to fetch last timestamp order entered those states

williardx:
Followup from https://github.com/artsy/exchange/pull/134#discussion_r213509183, this a quick update to rename `submitted_at` -> `last_submitted_at` and `approved_at` -> `last_approved_at` to return the most recent timestamps at which the order entered those states. Also adds new helper method `get_last_state_timestamp`.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -93,12 +93,12 @@ class Order < ApplicationRecord
     "Order #{id}"
   end
 
-  def submitted_at
-    state_histories.find_by(state: Order::SUBMITTED)&.updated_at
+  def last_submitted_at
+    get_last_state_timestamp(Order::SUBMITTED)
   end
 
-  def approved_at
-    state_histories.find_by(state: Order::APPROVED)&.updated_at
+  def last_approved_at
+    get_last_state_timestamp(Order::APPROVED)
   end
 
   private
@@ -110,6 +110,10 @@ class Order < ApplicationRecord
   def update_state_timestamps
     self.state_updated_at = Time.now.utc
     self.state_expires_at = STATE_EXPIRATIONS.key?(state) ? state_updated_at + STATE_EXPIRATIONS[state] : nil
+  end
+
+  def get_last_state_timestamp(state)
+    state_histories.where(state: state).order(:updated_at).last&.updated_at
   end
 
   def create_state_history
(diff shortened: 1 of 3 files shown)


app/models/order.rb
```diff
@@ -112,6 +112,10 @@ def update_state_timestamps
     self.state_expires_at = STATE_EXPIRATIONS.key?(state) ? state_updated_at + STATE_EXPIRATIONS[state] : nil
   end
 
+  def get_last_state_timestamp(state)
+    state_histories.where(state: state).order(:updated_at).last&.updated_at
```
ashkan18: Im really impressed by how ActiveRecord works here!
```ruby
2.5.1 :002 > StateHistory.where(state: 'pending').order(:updated_at).first
  StateHistory Load (2.2ms)  SELECT  "state_histories".* FROM "state_histories" WHERE "state_histories"."state" = $1 ORDER BY "state_histories"."updated_at" ASC LIMIT $2  [["state", "pending"], ["LIMIT", 1]]
 => nil
2.5.1 :003 > StateHistory.where(state: 'pending').order(:updated_at).last
  StateHistory Load (0.3ms)  SELECT  "state_histories".* FROM "state_histories" WHERE "state_histories"."state" = $1 ORDER BY "state_histories"."updated_at" DESC LIMIT $2  [["state", "pending"], ["LIMIT", 1]]
```
By just reading the current
```
state_histories.where(state: state).order(:updated_at).last&.updated_at
```
I thought it would get all the order state histories and order them by `update_at` asc (the default) and then we get the last one from the array which would have been little costly! 

By looking at the experience above, looks like its actually smart enough and translate `order(:update_at).last` to `order by updated_at desc limit 1` 

If you want to map what `ActiveRecord` does in our code to we could change that line to:
```ruby
state_histories.where(state: state).order(created_at: :desc).limit(1).pluck(:updated_at)
```
otherwise we can just 🥇 `ActiveRecord`

app/models/order.rb
```diff
@@ -128,7 +132,7 @@ def build_machine
     machine = MicroMachine.new(state)
     machine.when(:abandon, PENDING => ABANDONED)
     machine.when(:submit, PENDING => SUBMITTED)
-    machine.when(:approve, SUBMITTED => APPROVED)
+    machine.when(:approve, SUBMITTED => APPROVED, FULFILLED => APPROVED)
```
ashkan18: did you add this mainly for testing here? eventually we need to do this i think anyways. Once thing about it though is would be nice to make sure this can only happen by admins but that can happen later.

williardx: Ah! I missed this -- I'd rather take it out for now and be deliberate about when we want to add it, i.e. once we have the associated admin functionality in place.

app/models/order.rb
williardx: I say we leave `ActiveRecord` since it's doing the right thing 🏆

PR #145, opened 2018-08-30
Destination Amount

ashkan18:
# Problem
When charge is captured, we need to transfer only `seller_total_cents` to their Stripe account so this way we don't end up paying for `transaction_fee` and we also take our possible `commission_fee`.

With current setup, all the money goes  to `seller` and we end up paying for transaction fee and not taking commission.

# Solution
Use [Stripe's `destination[:amount]`](https://stripe.com/docs/connect/destination-charges) and set it to `seller_total_cents`. 
This way we only transfer the seller they portion they supposed to receive and we keep the rest on our account which then Stripe takes their fee from that amount.

# Any Solutions Out There?!
our current logic in https://github.com/artsy/exchange/compare/master...ashkan18:destination-amount?expand=1#diff-7d9a8004a01f1e811ed9226d20c6b1e4R21 is somewhat fragile. The order of things happening really matters and we are doing multiple saves on order. 

For example we need to make sure we set commission fee and transaction fee on the model before calling `PaymentService`.

I personally think this is just the nature of this logic and it might be ok, but as you review this code look for possible ways of making this logic less complicated 👀 

cc: @joanna20 

# Other Changes
## Refactor
Did some refactoring in `PaymentService` to return either failed or successful transaction instead of returning charge hash. This way caller services `OrderService` or `OrderSubmitService` don't need to translate Stripe charge to transaction.

## Metadata
Added `metadata` to Stripe charge call:
```ruby
{
      exchange_order_id: @order.id,
      buyer_id: @order.buyer_id,
      seller_id: @order.seller_id,
      type: 'bn-mo'
    }
```
cc: @madeleineb

## Description
Now we use `<partner name> via Artsy` as charge description (this is whats shown on credit card statement). 
Currently we are limiting it to 23 characters which used to be limit on description but don't see that requirement anymore, we may be able to show full partner name.


app/services/order_submit_service.rb
```diff
@@ -49,9 +48,13 @@ def pre_process!
     assert_credit_card!
     @partner = GravityService.fetch_partner(@order.seller_id)
```
williardx: Unrelated to this PR but do we need `partner` anymore? I don't see it used anywhere and it'd be nice to cut out this Gravity call.

ashkan18: we use it https://github.com/artsy/exchange/pull/145/files#diff-7d9a8004a01f1e811ed9226d20c6b1e4R97 when calculating commission fee.

ashkan18:
@williardx im thinking of doing some refactoring on this area, basically thinking of updating `PaymentService` to actually return an `Transaction` instead of Stripe `Charge`. This way we don't have translate stripe charge to our transaction in `OrderService` and `OrderSubmitService`. I'll do it in one commit and let me know what you think about it.

app/models/order.rb
```diff
@@ -59,7 +59,6 @@ class Order < ApplicationRecord
         save!
         create_state_history
         block.call if block.present?
-        self
```
ashkan18: i _think_ this makes more sense to have this method return the result of the block instead of always returning `self` not sure whats the recommended way but this current way allows us doing things like
```ruby
transaction = order.submit! do
  PaymentService....
end
```

starsirius: I'm not sure about the recommended way, but I personally feel having consistent return value is easier to predict from a developer's point of view. Previously version always returns `self` and may be useful for chaining, etc. If I understand correctly, this version behaves like:

- If there is no block given, it returns whatever `create_state_history` returns.
- If there a block given, it returns whatever the block returns.

app/services/order_service.rb
```diff
@@ -68,21 +58,23 @@ def self.fulfill_at_once!(order, fulfillment, by)
   end
 
   def self.reject!(order, by)
-    order.reject! do
+    transaction = order.reject! do
       refund(order)
     end
+    order.transactions << transaction
```
starsirius: Do we need to `order.save!` after appending a new transaction?

app/services/order_service.rb
```diff
@@ -68,21 +58,23 @@ def self.fulfill_at_once!(order, fulfillment, by)
...
     PostNotificationJob.perform_later(order.id, Order::REJECTED, by)
     order
   rescue Errors::PaymentError => e
-    TransactionService.create!(order, e.body)
+    order.transactions << e.transaction
```
joeyAghion: It's a little funky to get the transaction from the block in 2 different ways (and have to record it in both cases, or risk failing to record if not careful). I think it would be cleaner to simply return a transaction but be able to test it for whether it's successful or not, and only do certain follow-up actions if so.

app/services/order_submit_service.rb
```diff
@@ -49,33 +47,27 @@ def pre_process!
...
     @merchant_account = GravityService.get_merchant_account(@order.seller_id)
+    @order.commission_fee_cents = calculate_commission_fee
+    @order.transaction_fee_cents = calculate_transaction_fee
+    @order.save!
```
starsirius: #minor. These can be
```ruby
@order.update! commission_fee_cents: calculate_commission_fee, transaction_fee_cents: calculate_transaction_fee
```

app/services/payment_service.rb
```diff
@@ -1,61 +1,52 @@
...
-      source: source_id,
-      customer: customer_id,
-      destination: destination_id,
+      description: description,
+      source: credit_card[:external_id],
+      customer: credit_card[:customer_account][:external_id],
+      destination: {
+        account: merchant_account[:external_id],
+        amount: seller_amount
+      },
```
starsirius: 👍

app/services/order_service.rb
starsirius: Maybe we can get rid of the `Errors::PaymentError` bubbling up from `PaymentService`, given that we expect to always get back a transaction, succeeded or not.

starsirius:
Trying to summarize, I'm seeing 2 refactors here:

- Change the return value of `order.submit!` and `order.submit! do ... end`.
- `PaymentService` methods always generate a transaction, whether it's a success or failure. Maybe we can get rid of raising errors, so we can get the generated transaction the same way.

No blockers from me in this PR and feel free to address the comments in a followup. Thanks for the fix and refactor, @ashkan18!!

app/models/order.rb
ashkan18: yeah im not sure about the recommended way for this, having it returning the block's result makes it easier for the caller to do things like what i have above, basically:
```
transaction = order.submit! do
  PaymentService....
end
```
this way if they wanted to return `order` we can still do it in blocks return. @joeyAghion do you have any preference on this? I learned blocks from you :)

app/services/order_service.rb
ashkan18: yes we dont 👍 acrtiverecord overrides `<<` method.

app/services/order_service.rb
ashkan18: good point! that does make this cleaner! changing.

ashkan18: so looking at how the change would look, the biggest change there would be, right now we use the `Errors::PaymentError` being raised to rollback current transaction, with this change that logic has to change in the way where we first call Stripe (`PaymentService`) to do its thing and then if it was successful we approve/reject the order.
For example, `OrderService.approve!` will look like:
```ruby
  transaction = PaymentService.capture_charge(order.external_charge_id)
  order.transactions << transaction
  raise Errors::PaymentError("Could not approve order #{order.id}: #{transaction.failure_message}") if transaction.status == Transaction::FAILURE
  order.approve!
  PostNotificationJob.perform_later(order.id, Order::APPROVED, by)
  OrderFollowUpJob.set(wait_until: order.state_expires_at).perform_later(order.id, order.state)
end
```
as opposed to current
```ruby
  transaction = order.approve! do
    PaymentService.capture_charge(order.external_charge_id)
  end
  order.transactions << transaction
  PostNotificationJob.perform_later(order.id, Order::APPROVED, by)
  OrderFollowUpJob.set(wait_until: order.state_expires_at).perform_later(order.id, order.state)
  order
rescue Errors::PaymentError => e
  TransactionService.create!(order, e.body)
  Rails.logger.error("Could not approve order #{order.id}: #{e.message}")
  raise e
end
```
im 👍  with this change, as long as no one can see any issues with calling stripe first, then updating status.

ashkan18:
Looking at this PR, im leaning towards merging this as it is and do a follow up PR mainly addressing the refactoring in `PaymentService` and most likely moving towards having `PaymentService` methods accepting a block. 
Adding that to this PR will make review kind of complicated. Going to resolve this conflict and merge and do a separate PR.

app/services/order_service.rb
ashkan18: we will still lock the order when actually approving, but the idea would be to have `PaymentService` getting a block and in there maybe as a param we provide newly created transaction which in the block we can act on it. Still need to spend some time on it but wanted to have that refactor as its own thing so its easier to read and cleaner without other changed made in this PR.

PR #146, opened 2018-08-30
Add Sidekiq Dashboard, update readme

ashkan18:
# Changes
 - Added Sidekiq Dashboard currently limited to admins and under `/admin/sidekiq`.
- Added `ArtsyAuthHelper` class which has common code needed to decode/verify a JWT and do some validation
- Refactored `AuthHandler` to use ☝️ 
- Updated README to include sidekiq dashboard urls and also removed query section from README since it gets out of date often and changed it to point to our Insomnia file.


README.md
```diff
@@ -5,8 +5,8 @@ Exchange aspires to be responsible for the various types of e-commerce interacti
 ## Meta
 
 * State: development
-* Production: https://exchange.artsy.net, [Admin Dashboard](https://exchange.artsy.net/admin)
-* Staging: https://exchange-staging.artsy.net, [Admin Dashboard](https://exchange-staging.artsy.net/admin)
+* Production: https://exchange.artsy.net, [Admin Dashboard](https://exchange.artsy.net/admin) | [Sidekiq Dashboard](https://exchange.artsy.net/admin/sidekiq)
+* Staging: https://exchange-staging.artsy.net, [Admin Dashboard](https://exchange-staging.artsy.net/admin) | [Sidekiq Dashboard](https://exchange-staging.artsy.net/admin/sidekiq)
```
williardx: This URL is 404'ing

ashkan18: yes, it will be available once this PR is merged (only to admins though)

lib/artsy_auth_helper.rb
```diff
@@ -0,0 +1,22 @@
+class ArtsyAuthHelper
```
williardx: What if this were called something like `ArtsyJWT` instead? `ArtsyAuthHelper` feels a little bit vague given that this class seems to only deal only with the JWT

ashkan18: interesting, my only issue with calling it `ArtsyJWT` is we include implementation detail in the name itself, lets say if in the future we changed it to KWT we don't want to change the class name.

williardx: Maybe `ArtsyToken` then?

ashkan18: yeah! makes sense 👍

ashkan18:
merging 🙌

PR #149, opened 2018-09-03
Add phone_number field to shipping information (or, add buyer_phone_number field to order)

ds300:
Tested locally:
image

https://artsyproduct.atlassian.net/browse/PURCHASE-395

### DIFF
--- a/app/graphql/mutations/set_shipping.rb
+++ b/app/graphql/mutations/set_shipping.rb
@@ -11,8 +11,8 @@ class Mutations::SetShipping < Mutations::BaseMutation
   def resolve(id:, fulfillment_type:, phone_number:, shipping: {})
     order = Order.find(id)
     validate_buyer_request!(order)
-    
-    raise Errors::AuthError, 'Phone number is required' if  fulfillment_type == Order::SHIP && phone_number == nil
+
+    raise Errors::AuthError, 'Phone number is required' if fulfillment_type == Order::SHIP && phone_number.nil?
 
     shipping = AddressParser.parse!(shipping.to_h) if fulfillment_type == Order::SHIP
     {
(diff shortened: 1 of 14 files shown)


ds300:
Currently we will only ask for the phone number for the SHIP fulfilment type. At least that's what it says in the Jira ticket (which I forgot to attach above, apologies) https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=41&projectKey=PURCHASE&modal=detail&selectedIssue=PURCHASE-394

Your suggestion makes sense as future-proofing though. I like it :+1: and will do that.

ds300:
@ashkan18 updated 🍏 

image

app/graphql/mutations/set_shipping.rb
```diff
@@ -3,16 +3,20 @@ class Mutations::SetShipping < Mutations::BaseMutation
...
   argument :shipping, Inputs::ShippingAttributes, required: false
 
   field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure', null: false
 
-  def resolve(id:, fulfillment_type:, shipping: {})
+  def resolve(id:, fulfillment_type:, phone_number:, shipping: {})
     order = Order.find(id)
     validate_buyer_request!(order)
+
+    raise Errors::AuthError, 'Phone number is required' if fulfillment_type == Order::SHIP && phone_number.nil?
```
ashkan18: part of me wants to say lets make `phone_number` required all the time if its mandatory for shipping. I can imagine `pickup` is actually dependent on it! but we can do that in a separate PR.

ashkan18:
As a side note, phone number is PII along with few other fields in Exchange, I've talked to @madeleineb about it and it seems for now we are good with keeping this data in Exchange since there is a good business reason behind it, but just something to keep in mind.

PR #151, opened 2018-09-04
rename migration file

ds300:
Accidentally left this migration file with the wrong name after a refactor.


ashkan18:
🙌 merge on green

PR #152, opened 2018-09-04
Set stateUpdatedAt on newly created orders

ashkan18:
Still seeing issues with `pending` orders not having `state_expires_at` and `state_updated_at`, adding more spec around this and also possible fix.


ashkan18:
cc: @williardx this is still a mystery to me, lot of retries here https://exchange-staging.artsy.net/admin/sidekiq/retries lets see if this fixes it.

PR #153, opened 2018-09-05
Make sure phoneNumber is not a required argument in the method

ashkan18:
# Problem
We get `ArgumentError (missing keyword: phone_number):` when we don't pass `phone_number` as an input argument to `setShipping` mutation, this arg is optional and is ok to not have it in input.

# Solution
Set the default to `nil` in the `resolve` method and add spec for checking that.

### DIFF
--- a/spec/controllers/api/requests/set_shipping_mutation_request_spec.rb
+++ b/spec/controllers/api/requests/set_shipping_mutation_request_spec.rb
@@ -15,6 +15,7 @@ describe Api::GraphqlController, type: :request do
     let(:shipping_region) { 'Tehran' }
     let(:fulfillment_type) { 'SHIP' }
     let(:total_sales_tax) { 2222 }
+    let(:phone_number) { '00123456789' }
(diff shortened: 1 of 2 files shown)


ashkan18:
Sarah:
> merge on green

spec/controllers/api/requests/set_shipping_mutation_request_spec.rb
```diff
@@ -73,7 +73,7 @@
             addressLine1: 'Vanak',
             addressLine2: 'P 80'
           }
-        }
+        }.compact
```
ds300: why is this needed?

ashkan18: in the original test we had, when setting `phoneNumber` to `nil` it was still passed as an argument. So we were testing `setShipping(phoneNumber: nil....)` the error was happening when someone was not passing `phoneNumber` so here `.compact` will remove `nil` params.

PR #154, opened 2018-09-05
Adds Sentry

yuki24:
This PR enables Sentry error reporting. I've already created [exchange-production](https://sentry.io/artsynet/exchange-production/) and [exchange-staging](https://sentry.io/artsynet/exchange-staging) projects in Sentry and set ENV variables (`ENV['SENTRY_DSN']`) to both staging and production. All we have to do is to just merge and deploy :)

### DIFF
--- a/config/initializers/sentry_raven.rb
+++ b/config/initializers/sentry_raven.rb
@@ -1,0 +1,3 @@
+Raven.configure do |config|
+  config.dsn = ENV['SENTRY_DSN']
+end
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ gem 'jwt'
 gem 'micromachine'
 gem 'money' # Library for dealing with money and currency conversion
 gem 'paper_trail'
+gem 'sentry-raven'
 gem 'sidekiq'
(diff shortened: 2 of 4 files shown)


config/initializers/sentry_raven.rb
```diff
@@ -0,0 +1,5 @@
+Raven.configure do |config|
+  config.dsn = ENV['SENTRY_DSN']
+
+  config.processors -= [Raven::Processor::PostData]
```
yuki24: This will configure `Raven` to send HTTP bodies posted to Exchange, which may contain personally identifiable information. It will make it a lot easier to debug, but I'm also happy to  disable it in case this could cause issues in terms of GDPR.

ashkan18: good point, im personally more for not sending them.

yuki24:
> also is there a migration needed here to set env variables?

I've already set an appropriate value to `ENV['SENTRY_DSN']` both in staging and production so this should be good to go.

sweir27:
Niiiiice!

ashkan18:
mergin....

PR #155, opened 2018-09-05
Check artwork versions

saolsen:
Adds version id to line items.


app/services/create_order_service.rb
```diff
@@ -16,6 +16,7 @@ def self.with_artwork!(user_id:, artwork_id:, edition_set_id: nil, quantity:)
       )
       order.line_items.create!(
         artwork_id: artwork_id,
+        artwork_version_id: artwork[:current_version_id],
```
williardx: Should this ever be `nil`? Should we not create orders if there's no version id?

ashkan18: i think we did migrate all artworks to have `current_version_id` but we could add a check here.

ashkan18:
@saolsen looks like there is a `schema.rb` conflict, for resolving that the easiest approach i found locally was to merge `master` and accept incoming changes and then do a
```shell
rails db:drop
rails db:create
rails db:migrate
```

saolsen:
Sorry, forgot about this. I merged master in and regenerated db. Should be good to go.

ashkan18:
🙌

PR #156, opened 2018-09-05
Randomize order codes

williardx:
### Problem
Our human-readable order code is currently sequential and based on the order id, which gives away information about our orders.

### Solution
Generate a randomized human-readable order code.

### What changed
- The code is now a 0-padded 9-digit integer.
- `update_code` checks the database to ensure the code is unique and will attempt to set the code 10 times before giving up and throwing an error.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -90,7 +90,7 @@ class Order < ApplicationRecord
 
   def update_code(attempts = 10)
     while attempts.positive?
-      code = SecureRandom.rand(100000000..999999999)
+      code = format('%09d', SecureRandom.rand(999999999))
       unless Order.where(code: code).exists?
         update!(code: code)
         break
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -89,7 +89,15 @@ class Order < ApplicationRecord
   private
 
   def set_code
-    update!(code: format('B%06d', id))
+    update!(code: generate_code)
+  end
+
+  def generate_code(size = 8, num_chars = 3)
+    nums = %w{ 2 3 4 6 7 9 }
+    chars = %w{ A C D E F G H J K M N P Q R T V W X Y Z}
+    random_nums = (0..(size - num_chars - 1)).map { nums.to_a[rand(nums.size)] }.join
+    random_chars = (0..(num_chars - 1)).map { chars.to_a[rand(chars.size)] }.join
+    random_nums + random_chars
   end
 
   def update_state_timestamps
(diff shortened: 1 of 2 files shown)


app/models/order.rb
```diff
@@ -88,8 +88,24 @@ def last_approved_at
...
+    end
+    raise Errors::OrderError, 'Failed to set order code' if attempts.zero?
+  end
+
+  def generate_code(size = 8, num_chars = 3)
+    nums = %w[2 3 4 6 7 9]
+    chars = %w[A C D E F G H J K M N P Q R T V W X Y Z]
+    random_nums = Array.new(size - num_chars).map { nums.sample }.join
+    random_chars = Array.new(num_chars).map { chars.sample }.join
+    random_nums + random_chars
```
joeyAghion: I trust that this works, but it's a lot of custom code for a function that's fairly common. Could we rely on built-in stuff like `SecureRandom` instead, or is the numbers-followed-by-characters pattern critical?

`SecureRandom.base58(8).upcase` gives random uppercase combinations from an alphabet that excludes ambiguous characters. https://api.rubyonrails.org/classes/SecureRandom.html

williardx: The current pattern is more to have something that feels predictable, easy to read and matches an ecommerce brand from a user perspective. E.g. we had some discussion about how order codes that have too many letters in them feel "airline-y."

That said, though, this could be a random 8-digit number, and maybe that's better just because it's simpler.

app/models/order.rb
```diff
@@ -88,8 +88,16 @@ def last_approved_at
 
   private
 
-  def set_code
-    update!(code: format('B%06d', id))
+  def update_code(attempts = 10)
+    while attempts.positive?
+      code = SecureRandom.rand(100000000..999999999)
```
ashkan18: we could use same code but zero pad anything below 100000000 that way we get more chances. What do you think? that way the code will look like
```ruby
code = format('%09d', SecureRandom.rand(999999999))
```

spec/models/order_spec.rb
```diff
@@ -53,9 +53,15 @@
...
-      expect(order.code).to match(/^B\d{6}$/)
+  describe '#update_code' do
+    it 'raises an error if it is unable to set a code within specified attempts' do
+      expect do
+        order.send(:update_code, 0)
+      end.to raise_error(Errors::OrderError, 'Failed to set order code')
+    end
+
+    it 'sets code on order creation' do
+      expect(order.code).not_to be_nil
```
ashkan18: we can check `order.code.length` and maybe check format (make sure there are no characters in it.

PR #158, opened 2018-09-06
add fulfillment_type

erikdstock:
I just realized that the `fulfillment_type` property on an order is missing, but pulse expects it to be there since some emails include shipment information. I also sorted the lines on this growing list of order event properties.

@ashleyjelks, @ds300, @anson, for context: This `OrderEvent` basically represents an event about an order going out to our rabbitmq service. Pulse will receive the event with a `properties` key, but it isn't an exact copy of our exchange event- so I'm adding a field for the fulfillment type. Please comment or message me if you have questions!


ashkan18:
Nice! makes sense! merge on green

PR #160, opened 2018-09-06
OrderTotalUpdaterService

ashkan18:
# Problem
We were not calling update totals before submission, we were setting commission and transaction fees and were missing updating totals after that which would lead to calling stripe and not holding commission and transaction in our account.

# Solution
Totals were calculated in 2 different places, moved all of them into one service which can get commission rate as an input and based on that calculates all the totals of an order in one place! this makes things little more sane... hopefully.

### DIFF
--- a/app/services/order_total_updater_service.rb
+++ b/app/services/order_total_updater_service.rb
@@ -16,6 +16,7 @@ class OrderTotalUpdaterService
   def self.calculate_transaction_fee(order)
     return 0 unless order.buyer_total_cents.positive?
     # This is based on Stripe US fee, it will be different for other countries
+    # https://stripe.com/us/pricing
     (Money.new(order.buyer_total_cents * 2.9 / 100, 'USD') + Money.new(30, 'USD')).cents
   end
 end
--- a/app/services/order_total_updater_service.rb
+++ b/app/services/order_total_updater_service.rb
@@ -1,4 +1,4 @@
-class OrderTotalUpdaterService
+module OrderTotalUpdaterService
   def self.update_totals!(order, commission_rate = nil)
     raise Errors::OrderError, 'Missing price info on line items' if order.line_items.any? { |li| li.price_cents.nil? }
     order.items_total_cents = order.line_items.map(&:total_amount_cents).sum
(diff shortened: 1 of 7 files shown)


app/services/order_total_updater_service.rb
```diff
@@ -0,0 +1,21 @@
+class OrderTotalUpdaterService
+  def self.update_totals!(order, commission_rate = nil)
```
joeyAghion: This looks good to me at a high level. I'd prefer to call it `OrderTotalService` or `OrderCalculator` or something like that, and actually instantiate it with an `order` and expose the `update_totals!` method publicly while the other helpers can remain private.

ashkan18: good catch, it's a class since i originally had it so you could create new instance and set order on it. Based on 👇 i can change it to `module`.

ashkan18: RE name, i wanted to make it more obvious that it actually updates the order model and doesn't just calculate totals, makes sense?

app/services/order_total_updater_service.rb
```diff
@@ -0,0 +1,21 @@
...
+    raise Errors::OrderError, 'Missing price info on line items' if order.line_items.any? { |li| li.price_cents.nil? }
+    order.items_total_cents = order.line_items.map(&:total_amount_cents).sum
+    order.buyer_total_cents = order.items_total_cents + order.shipping_total_cents.to_i + order.tax_total_cents.to_i
+    order.commission_fee_cents = calculate_commission_fee(order, commission_rate) if commission_rate.present?
+    order.transaction_fee_cents = calculate_transaction_fee(order)
+    order.seller_total_cents = order.buyer_total_cents - order.commission_fee_cents.to_i - order.transaction_fee_cents.to_i
```
williardx: This doesn't need to happen in this PR but we need to account for the fact that sales tax is selectively collected in `seller_total_cents`. We'll need to subtract out the sales tax for any line items where Artsy needs to remit taxes.

ashkan18: is that tracked somewhere?

williardx: Not that I know of! I'll make a ticket.

app/services/order_total_updater_service.rb
joeyAghion: I don't think it's necessarily bad for the class to expose a single public method. Or, at least, that's really the same question as whether to organize this separately from `OrderService` at all. Either way it's fine if you want to merge this as-is in the interest of the demo.

ashkan18: Nah, we still have time till demo :) , updated to use class and added more spec to make sure commission rate is in desired range. going to merge 👍

PR #162, opened 2018-09-07
[COORDINATED MERGE] move phone_number to shipping input and Ship output

ds300:
:warning: Do not merge until we have all PRs related to this change in place and approved. 

here it is working in exchange local:

image

### DIFF
--- a/app/graphql/inputs/shipping_attributes.rb
+++ b/app/graphql/inputs/shipping_attributes.rb
@@ -1,6 +1,7 @@
 class Inputs::ShippingAttributes < Types::BaseInputObject
   description 'Shipping information'
 
+  # shouldn't most of these be required?
   argument :name, String, required: false
   argument :address_line1, String, required: false
   argument :address_line2, String, required: false
@@ -8,4 +9,5 @@ class Inputs::ShippingAttributes < Types::BaseInputObject
   argument :region, String, required: false
   argument :country, String, required: false
   argument :postal_code, String, required: false
+  argument :phone_number, String, required: true
 end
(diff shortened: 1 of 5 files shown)


app/graphql/types/requested_fulfillment_union_type.rb
```diff
@@ -1,4 +1,8 @@
 class Types::Ship < Types::BaseObject
+  field :phone_number, String, null: true
+  def phone_number
```
ashkan18: do you think we should actually call it `buyer_phone_number`? sounds informational but at the same time we don't say `buyer_shipping_address_line1` anywhere so 🤷‍♂️

ds300: I think buyer is implied when it's on the Ship type?

ashkan18: good point! 🙌

alloy: It may also be a person at the shipping location, though. How about also defining a `Order#buyer_phone_number` field and have the various fulfilment `phone_number` fields default to it? (Once needed)

ds300: That's exactly what we have right now. The `buyer_phone_number` field is still there, and it's the canonical truth. We don't store another phone number separately for the fulfilment, it just gets copied over from the `buyer_phone_number` field on the `Order` rails model. It could be overridden with a separate value one day if needed.

app/graphql/inputs/shipping_attributes.rb
```diff
@@ -1,11 +1,13 @@
 class Inputs::ShippingAttributes < Types::BaseInputObject
   description 'Shipping information'
 
+  # shouldn't most of these be required?
```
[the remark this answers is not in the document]
williardx: I'm working on this!

app/graphql/types/requested_fulfillment_union_type.rb
alloy: Oh yeah, I wasn’t looking at the diff 🙈 Well, excellent 👌

PR #164, opened 2018-09-07
Add TaxJar sandbox

williardx:
Adds new environment variable `TAXJAR_API_URL` that sets TaxJar's API URL. Defaults to default URL in TaxJar client if no URL given.

### Migrations

- [x] Add environment variable to staging and set to sandbox URL
- [x] Remove `USE_TAXJAR_SANDBOX` env variable from staging


config/taxjar.yml
```diff
@@ -1,5 +1,6 @@
 development: &default
   taxjar_api_key: <%= ENV['TAXJAR_API_KEY'] || '' %>
+  use_taxjar_sandbox: <%= ENV['USE_TAXJAR_SANDBOX'] || false %>
```
ashkan18: there is this tricky thing with using boolean on Env variable. The thing is the variable comes in as a _string_ so basically when consuming it you have to do string matching otherwise if someone set `USE_TAXJAR_SANDBOX=false` with the check above `use_taxjar_sandbox` will be true.

To make this easier, is there a way we could switch this param to be the actual url? so instead of `USE_TAXJAR_SANDBOX` define `TAXJAR_SANDBOX_API`?

PR #165, opened 2018-09-07
Validate shipping address to match tax requirements

williardx:
### Problem
TaxJar requires certain fields to be filled out for US and Canada-based orders (e.g. state, postal code, etc.). We currently don't validate these fields, resulting in errors from TaxJar if a user inputs the wrong thing.

### Solution
Validate shipping addresses for shipped orders in the US or Canada. `AddressParser` will make an attempt to parse the address and will then pass along the parsed address to `OrderShippingService.set_shipping!`, which will validate it.

### What Changed
- Adds `validate_shipping!` method in `set_shipping!` to validate addresses.
- Allows `AddressParser` to return `nil` for countries / regions it can't recognize instead of throwing an error. This pushes the logic of what we do and don't accept for an address into `OrderShippingService`, which feels correct since it inherently has to do with our shipping implementation.
- Moves `OrderService.set_shipping!` and related methods to a new module `OrderShippingService` since the tax / shipping logic in `OrderService` was getting complicated.


app/services/order_shipping_service.rb
```diff
@@ -0,0 +1,61 @@
+module OrderShippingService
+  def self.set_shipping!(order, fulfillment_type:, phone_number:, shipping: {})
```
ashkan18: Nice on moving this logic out of `OrderService`. Do you think we could follow the same pattern of `initialize` with order and shipping info and have a `process!` method or something? I think it would be good since it'll be more consistent with newer services and also we could make all of those methods below `private`.

williardx: Yep! I'll update it to fit the new style.

app/services/order_shipping_service.rb
```diff
@@ -0,0 +1,97 @@
...
+          shipping_address_line2: @shipping[:address_line2],
+          shipping_city: @shipping[:city],
+          shipping_region: @shipping[:region],
+          shipping_country: @shipping[:country],
+          shipping_postal_code: @shipping[:postal_code]
+        )
+      )
+      OrderTotalUpdaterService.new(@order).update_totals!
+    end
+    @order
```
ashkan18: super #minor we can stop returning `order` from these `process!` methods. for newer services i ended up doing something like https://github.com/artsy/exchange/blob/master/app/graphql/mutations/approve_order.rb#L11

williardx: 👍 👍

app/services/order_shipping_service.rb
```diff
@@ -0,0 +1,96 @@
...
+
+  def validate_artwork!(artwork)
+    raise Errors::OrderError, 'Cannot set shipping, unknown artwork' unless artwork
+    raise Errors::OrderError, 'Cannot set shipping, missing artwork location' if artwork[:location].blank?
+  end
+
+  def validate_shipping!
+    raise Errors::OrderError, 'Valid country required for shipping address' if @shipping[:country].nil?
+    validate_us_shipping! if @shipping[:country] == 'US'
+    validate_ca_shipping! if @shipping[:country] == 'CA'
```
yuki24: At some point we should probably replace these one-off validation rules with `ActiveModel::Validations` so service objects don't grow too much and the logic for validations will be re-useable.

PR #150, opened 2018-09-03
[WIP] First pass in adding state reasons

ashkan18:
Work in progress

PR #166, opened 2018-09-09
Refactors

ashkan18:
# Changes
-  Modified `PaymentService` methods to always return a `Transaction`, based on success or failure, `transaction.status` would be different
- Moved `approve!`, `reject!` and `seller!` out of `OrderService`
- used `ensure` to make sure transactions coming back from `PaymentService` are stored.
- Moved logic for validating order's currency code to `Order` as a validation
- Better error message from transition checking on `Order`
- Added more specs on service level


app/services/order_refund_service.rb
```diff
@@ -0,0 +1,33 @@
+class OrderRefundService
```
williardx: What if we called this `OrderCancelService`? I feel like the refund is an important commonality between `seller_lapse!` and `reject!` but it doesn't tell the whole story with what's happening

app/models/order.rb
```diff
@@ -42,7 +42,10 @@ class Order < ApplicationRecord
   has_many :transactions, dependent: :destroy
   has_many :state_histories, dependent: :destroy
 
+  before_validation { self.currency_code = currency_code.upcase if currency_code.present? }
```
williardx: IIRC we `currency_code.downcase` somewhere else for comparison. I don't have any attachment on which to use but we should pick one and be consistent with it

app/models/order.rb
```diff
@@ -42,7 +42,10 @@ class Order < ApplicationRecord
...
+
   validates :state, presence: true, inclusion: STATES
+  validates :currency_code, inclusion: SUPPORTED_CURRENCIES, allow_nil: true
```
williardx: Why do we allow `nil` for this? This seems like something that should always be defined

app/models/order.rb
```diff
@@ -61,7 +64,7 @@ class Order < ApplicationRecord
         block.call if block.present?
       end
     rescue MicroMachine::InvalidState
-      raise Errors::OrderError, "Invalid action on this #{state} order"
+      raise Errors::OrderError, "Invalid transition for #{state} order: #{id}"
```
williardx: 👏

app/services/order_service.rb
```diff
@@ -38,69 +36,21 @@ def self.set_shipping!(order, fulfillment_type:, phone_number:, shipping: {})
     order
   end
 
-  def self.approve!(order, by: nil)
```
williardx: Loving all the ✂️here

spec/models/order_spec.rb
```diff
@@ -11,7 +17,7 @@
 
     it 'does not update timestamps if state did not change' do
       current_timestamp = order.state_updated_at
-      order.update!(currency_code: 'CAD')
+      order.update!(shipping_total_cents: 12313)
```
williardx: Does updating `currency_code` cause this test to fail?

spec/services/payment_service_spec.rb
```diff
@@ -34,16 +34,14 @@
     end
     it 'catches Stripe errors and raises a PaymentError in its place' do
```
williardx: #minor but this doesn't test for raising a `PaymentError` anymore

williardx:
A few questions but this is overall looking 👌! Thanks for doing this refactor, I like this new direction for how we handle transactions.

app/models/order.rb
ashkan18: i noticed in Gravity its `upcase` so ended up forcing it here too, we were doing `downcase` for validating currency which is not needed anymore since we `upcase` it in `before_validation` now.

app/models/order.rb
ashkan18: good point, i mainly added `allow_nil` for the case where we create the order first and then we populate it with artwork data. But looks like we always get artwork first anyways so we should be 👍 , going to remove that check

app/services/order_refund_service.rb
ashkan18: yes, i wasn't really sure about the name of this service, i'm planning to redo https://github.com/artsy/exchange/pull/150 once this branch merged and lets revisit this name then 👍

spec/models/order_spec.rb
ashkan18: yes, using that currency code for test was failing test mainly since it was raising validation error for `CAD` had to just update something else on the model to verify state timestamp doesn't change.

spec/services/payment_service_spec.rb
ashkan18: good catch!!

app/services/order_refund_service.rb
williardx: Cool, sounds good

williardx:
Feel free to merge whenever you're ready 🙌

ashkan18:
🙌 talked to @sweir27, going to merge that PR and will open a follow up PR to add state reasons and will address her comments in follow up PR there.

app/services/order_refund_service.rb
sweir27: I agree with @williardx -- `OrderRefundService` feels too specific.

I also didn't realize we used the term `refund` both for releasing the hold _and_ actually refunding payment, since both of these refer to the former, right?

williardx: There's no distinction in Stripe's API, so we follow the same convention here.

ashkan18: im ok with calling that specific method `refund` to reflect what we actually are doing Stripe-wise but we could add a comment maybe saying we are trying to release the hold charge.

I renamed the class though to `OrderCancellationService` in  #169

PR #167, opened 2018-09-10
check artwork version on order submit

saolsen:
Checks artwork version for line items matches the current version on an order submit. If they differ raises an error.


williardx:
Thanks @saolsen, looks good 👍 

Do we need to do this for order approval as well?

ashkan18:
@williardx i don't think we need anything on approval, from user's perspective, they should be able to buy the order they have seen and verified. Maybe in CMS UI we want to show something telling partners but i think from Exchange's perspective they should still be allowed to approve it.

app/services/order_submit_service.rb
```diff
@@ -39,6 +39,10 @@ def process!
   private
 
   def pre_process!
+    @order.line_items.map do |li|
+      artwork = GravityService.get_artwork(li[:artwork_id])
+      raise Errors::OrderError, 'Artwork version out of date' if artwork[:current_version_id] != li[:artwork_version_id]
```
ashkan18: we've removed `Errors::OrderError` in https://github.com/artsy/exchange/pull/172, we need to change this line to
```ruby
raise Errors::ProcessingError, :artwork_version_missmatch if ....
```

ashkan18:
🙌

app/services/order_submit_service.rb
```diff
@@ -39,6 +39,10 @@ def process!
   private
 
   def pre_process!
+    @order.line_items.map do |li|
+      artwork = GravityService.get_artwork(li[:artwork_id])
+      raise Errors::ProcessingError, :artwork_version_missmatch if artwork[:current_version_id] != li[:artwork_version_id]
```
starsirius: Possible typo "mismatch"?

ashkan18: 👀 im fixing it in my error handling PR.

ashkan18: https://github.com/artsy/exchange/pull/184

PR #169, opened 2018-09-11
Add State Reason

ashkan18:
# Changes
Following our previous discussions, we want to combine few states and add `reason` as separate field to describe why we ended up in this combined states. 
https://docs.google.com/drawings/d/1WIcdO7GygNICfePxMF1cdrKkUDu-26ZJDU0irYlYMcw/edit

For this to work, we added `state_reason` to `Order` and also `reason` to `StateHistory`.

The idea here is to define a set of possible _reasons_ for states that require reasons.
When transitioning to one of this state that requires reason, you need to provide a valid reason from the all possible reasons for that state.

Currently only `canceled` state requires reason and there are two:
- `seller_lapsed`
- `seller_rejected`
reasons.

We kept `seller_lapse` and `rejected` actions which will basically go to `canceled` state now but they will set the reason properly.


app/models/order.rb
```diff
@@ -56,9 +68,10 @@ class Order < ApplicationRecord
   scope :active, -> { where(state: [SUBMITTED, APPROVED]) }
 
   ACTIONS.each do |action|
-    define_method "#{action}!" do |&block|
+    define_method "#{action}!" do |state_reason = nil, &block|
       with_lock do
         state_machine.trigger!(action)
+        self.state_reason = state_reason || ACTION_REASONS[action]
```
ashkan18: this line sets the reason based on action if no `state_reason` was passed to this method and also if current action has a default `reason`

app/graphql/types/order_state_enum.rb
```diff
@@ -3,7 +3,6 @@ class Types::OrderStateEnum < Types::BaseEnum
   value 'PENDING', 'order is still pending submission by collector', value: Order::PENDING
   value 'SUBMITTED', 'order is submitted by collector', value: Order::SUBMITTED
   value 'APPROVED', 'order is approved by seller', value: Order::APPROVED
-  value 'REJECTED', 'order is rejected by seller', value: Order::REJECTED
+  value 'CANCELED', 'order is canceled by seller', value: Order::CANCELED
```
williardx: Can't the order be canceled by a buyer or admin too?

app/graphql/types/order_state_enum.rb
```diff
@@ -3,7 +3,6 @@ class Types::OrderStateEnum < Types::BaseEnum
   value 'PENDING', 'order is still pending submission by collector', value: Order::PENDING
   value 'SUBMITTED', 'order is submitted by collector', value: Order::SUBMITTED
```
williardx: Don't need to fix this in this PR but we should change `collector` in this file to `buyer`

app/models/order.rb
```diff
@@ -91,6 +104,11 @@ def last_approved_at
 
   private
 
+  def state_reason_inclusion
+    errors.add(:state_reason, "Current state not expecting reason: #{state}") if state_reason.present? && !REASONS.key?(state)
+    errors.add(:state_reason, 'Invalid state reason') if REASONS[state] && !REASONS[state].value?(state_reason)
```
williardx: I'm reading this as this will  throw if you pass in a state reason that's not one of the default reasons specified in `REASONS` - is that right?

app/services/order_cancellation_service.rb
```diff
@@ -18,7 +18,7 @@ def reject!
     @order.reject! do
       refund
     end
-    PostNotificationJob.perform_later(@order.id, Order::REJECTED, @by)
+    PostNotificationJob.perform_later(@order.id, Order::CANCELED, @by)
```
williardx: Should we include the state reason here too? I can imagine we'll have different types of cancellation emails depending on the reason.

db/migrate/20180910194118_add_state_reason_to_orders.rb
```diff
@@ -0,0 +1,6 @@
+class AddStateReasonToOrders < ActiveRecord::Migration[5.2]
+  def change
+    add_column :orders, :state_reason, :string
```
williardx: What's the difference between `orders.state_reason` and `state_histories.reason`? Is `orders.state_reason` just the most recent state reason?

spec/models/order_spec.rb
```diff
@@ -9,6 +9,49 @@
     end
   end
 
+  describe 'validates state_reason' do
+    context 'state requiring reasons' do
+      it 'raises error when missing reason for states with required reason' do
+        expect do
+          order.update!(state: Order::CANCELED)
+        end.to raise_error(ActiveRecord::RecordInvalid, 'Validation failed: State reason Invalid state reason')
```
williardx: I wonder if we can make this error message more useful? An invalid state reason feels different than a missing state reason

app/graphql/types/order_state_enum.rb
ashkan18: good catch!

app/graphql/types/order_state_enum.rb
ashkan18: updating here while we are at it :)

app/models/order.rb
ashkan18: yes, so the idea here is to make sure we only accept specific reasons. We check 2 things:
1) Make sure your state expects a reason
2) Make sure passed in reason is valid for this state

app/services/order_cancellation_service.rb
ashkan18: reason is now part of the actual event https://github.com/artsy/exchange/pull/169/files#diff-e8e817ee0f1e56ba11f94f6752d82f09R40. 
This second parameter is used as the action in the event which eventually used as part of routing key.

db/migrate/20180910194118_add_state_reason_to_orders.rb
ashkan18: yes, do you think we should rename it? `state_reason` is basically _current state's reason_. So if goes to `canceled` state, which requires a reason, and state reason was `seller_lapesed` and then it goes to `submitted`, state without reason, `state_reason` should be `nil`.

spec/models/order_spec.rb
ashkan18: Good point, lets make sure this is covered in #172

spec/services/order_service_spec.rb
```diff
@@ -81,7 +82,7 @@
         Timecop.freeze do
           order.update!(state_updated_at: 10.days.ago)
           OrderService.abandon!(order)
-          expect(order.reload.state_updated_at.to_date).to eq Time.now.to_date
+          expect(order.reload.state_updated_at.to_date).to eq Time.now.utc.to_date
```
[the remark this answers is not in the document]
ashkan18: yeah! this test was failing after 6pm eastern time :)

williardx:
Feel free to merge whenever you're ready 💯

ashkan18:
lets do this 🙌

PR #170, opened 2018-09-11
[WIP] Error Handling

ashkan18:
The idea here is to have a set of predefined error types that can have a set of predefined errors with code and description.

This is WIP but sample error handling is already in `set_shipping.rb` and few other places

@ds300 take a look when had a chance and let me know what you think before we go deep in this route :)


ashkan18:
closing in favor of #172

PR #171, opened 2018-09-11
Improve the schema output to match metaphysics

orta:
Mainly to sync behavior with https://github.com/artsy/metaphysics/pull/1274


Gemfile.lock
```diff
@@ -108,6 +108,7 @@ GEM
       terminal-table (~> 1)
     dante (0.2.0)
     diff-lcs (1.3)
+    diffy (3.2.1)
```
orta: I punted on using diff-lcs for this, but it's too low level to provide a diff in the form that we're used to.

ashkan18:
@orta should we close this?

ashkan18:
talked to @orta going to close this

PR #172, opened 2018-09-12
[COORDINATED MERGE] Error Handling Take II

ashkan18:
# Problem
Current style in raising errors is hard to manage from Client perspective since all we are doing is returning an error description which is not easy to parse.

# Change
Remove descriptions and instead now each error has:

- `type`: string defining error area could be things like validation, processing
- `code` (description/meaning): things like invalid_order_state
- `data`: json including things like order id and possible charge id and etc

Here are some description and when they could happen https://www.notion.so/artsy/Error-and-failure-states-for-Orders-b3dfcdef33bc4f2b9a1f912568cb3c8e


joeyAghion:
I followed the link but couldn't find the discussion. I'm puzzled by the random codes, since this seems like something we'd want to be readable and meaningful, especially since these currently-mysterious values may get copied into other repositories.

In my opinion it's fine to decide on a predetermined set of error types that will be exposed to clients. (That's a sort of interface, after all.) Clients won't update their handling every time Exchange's logic gets enhanced, anyway, so there will have to be some grouping or hierarchy in what gets exposed.

ds300:
Ah, yes, that discussion happened in a slack 1-1. Must remember to take these things out into public!

Our idea was to have three levels of specificity when identifying an error:

- Type: e.g. 'validation error' or 'integration error'
   The idea here is to allow clients to handle classes of error in a forwards-compatible way, if need be.
- Meaning: e.g. 'cannot_find_merchant_account'
   In this PR this is handled by the 'description' field, and the error messages are supplied inline at `raise` sites.
- Location: the exact place in the code where the error was raised. We thought about a using a random string for permanence and maximum `cmd+shift+f`-ability, and also to make it easy to create new identifiers.

So clients should mainly be concerned with the first 2, but we wanted to include the third in case we could show it in a very discreet way somewhere so that it might help us narrow down a problem.

> In my opinion it's fine to decide on a predetermined set of error types that will be exposed to clients. (That's a sort of interface, after all.) Clients won't update their handling every time Exchange's logic gets enhanced, anyway, so there will have to be some grouping or hierarchy in what gets exposed.

:+1: :+1:

 On that note, @ashkan18 my main feedback for this PR is that the 'description' free text field doesn't give much leverage to the front end to decide how to display the errors. I would prefer to see enum-like human-readable codes and then a separate `data` field in case we need to pass extra information to the client. It would also be helpful if the codes were defined in one file like you had before, so we have a documented 'interface' and can track changes over time more easily.

ashkan18:
yes, i basically wanted to try and see how this random id per `raise` works, it was actually pretty tedious process setting it for these orders, you basically have to generate one and search everywhere to see if its already used anywhere or not. I think if we want to keep specific id per raise, we may be better with a sequence stored in a file or something but I suggest 👇 


I think between this PR and original one I like the middle solution where each error has:
- `type`: string defining error area could be things like `validation`, `processing`
- `code` (description/meaning): things like `invalid_order_state`
- `data`: json including things like order id and possible charge id and etc
 
What we lose here is exact location where possibly the same error has been `raised` lets say if there are two places where raise `ValidationError(:not_found)` we can't tell where this was raised from, but i guess for that we could refer to `data`, newrelic, sentry and other things.

Let me know what you think.

ds300:
I'm happy to go forward with your suggestion. It's all the frontend needs. If we decide that it's too annoying to disambiguate error `raise` sites via sentry in the future then we can revisit the guid idea.

ashkan18:
This PR is very conflict-friendly, it changes 50 files, would be good if we get this reviewed/merged sooner 🙌 😱

ds300:
This needs to be a coordinated merge with metaphysics and reaction and force and anything else I don't know about which executes a mutation against exchange. I'm working on the metaphysics PR right now.

alloy:
:ceiling-cat: I’m pretty sure that in another GraphQL backend app we decided to include status codes following the HTTP status codes so clients have a single spec they can follow. I don’t recall offhand which app it was, but it would be good to make sure the two don’t collide.

app/graphql/mutations/base_mutation.rb
```diff
@@ -1,9 +1,9 @@
 class Mutations::BaseMutation < GraphQL::Schema::RelayClassicMutation
   def validate_seller_request!(order)
-    raise Errors::AuthError, 'Not permitted' unless order.seller_type == Order::PARTNER && context[:current_user]['partner_ids'].include?(order.seller_id)
+    raise Errors::ValidationError, :not_found unless order.seller_type == Order::PARTNER && context[:current_user]['partner_ids'].include?(order.seller_id)
```
williardx: Why is this a `ValidationError` instead of `AuthError`?

app/services/create_order_service.rb
```diff
@@ -43,9 +47,7 @@ def self.find_edition_set(external_artwork, edition_set_id)
 
   # TODO: 🚨 update gravity to expose amount in cents and remove this duplicate logic
   # https://github.com/artsy/gravity/blob/65e398e3648d61175e7a8f4403a2d379b5aa2107/app/models/util/for_sale.rb#L221
-  def self.price_in_cents(price_in_dollars, currency)
-    raise Errors::OrderError, 'No price found' unless price_in_dollars&.positive?
-    raise Errors::OrderError, 'Missing currency' if currency.blank?
+  def self.price_in_cents(price_in_dollars)
```
williardx: We can get rid of this method and use `UnitConverter.convert_dollars_to_cents`

app/services/gravity_service.rb
```diff
@@ -4,27 +4,27 @@ def self.fetch_partner(partner_id)
       Adapters::GravityV1.get("/partner/#{partner_id}/all")
     end
   rescue Adapters::GravityNotFoundError
-    raise Errors::OrderError, 'Unable to find partner'
+    raise Errors::ValidationError.new(:unknown_partner, partner_id: partner_id)
   rescue Adapters::GravityError, StandardError => e
-    raise Errors::OrderError, e.message
+    raise Errors::ValidationError.new(:gravity_error, message: e.message)
```
williardx: I wonder if this should be something more generic like `ApplicationError` instead since a Gravity failure can cover many other kinds of failure beyond an invalid response

app/services/gravity_service.rb
```diff
@@ -61,8 +61,8 @@ def self.undeduct_inventory(line_item)
       Adapters::GravityV1.put("/artwork/#{line_item.artwork_id}/inventory", params: { undeduct: line_item.quantity })
     end
   rescue Adapters::GravityNotFoundError
-    raise Errors::OrderError, 'Credit card not found'
-  rescue Adapters::GravityError => e
-    raise Errors::InventoryError.new(e.message, line_item)
+    raise Errors::ValidationError.new(:unknown_artwork, line_item_id: line_item.id)
+  rescue Adapters::GravityError
+    raise Errors::ValidationError.new(:insufficient_inventory, line_item_id: line_item.id)
```
williardx: Is there a better way to test for `insufficient_inventory`? I'm worried about having all generic Gravity errors here raising as a validation error due to insufficient inventory

app/services/order_approve_service.rb
```diff
@@ -9,7 +9,7 @@ def initialize(order, by = nil)
   def process!
     @order.approve! do
       @transaction = PaymentService.capture_charge(@order.external_charge_id)
-      raise Errors::PaymentError, "Could not approve this order. #{@transaction}" if @transaction.failed?
+      raise Errors::ProcessingError.new(:capture_failed, @transaction.failure_data) if @transaction.failed?
```
williardx: 👏

app/services/order_submit_service.rb
```diff
@@ -14,28 +14,22 @@ def initialize(order, by)
...
+    pre_process!
     @order.submit! do
       # Try holding artwork and deduct inventory
       @order.line_items.each do |li|
         GravityService.deduct_inventory(li)
         deducted_inventory << li
       end
       @transaction = PaymentService.authorize_charge(construct_charge_params)
-      raise Errors::PaymentError, "Could not submit this order. #{@transaction}" if @transaction.failed?
+      raise Errors::ProcessingError.new(:failed_charge_authorize, @transaction) if @transaction.failed?
```
williardx: super #minor but would sound a little better as `failed_charge_authorization`

williardx:
Couple of minor comments but otherwise this looks good! Definitely a meaty PR, thanks for all the work @ashkan18!

app/graphql/mutations/base_mutation.rb
ashkan18: had some chat with @sweir27 about this, the thing is, from authentication perspective they are actually _authenticated_, they are not just authorized to access this order, might be better to return `not_found` since from their perspective this order should not exist and not exposed the fact that this record exist in our system.

williardx: Okay cool, that makes sense to me -- should the errors here https://github.com/artsy/exchange/pull/172/files#diff-782222440a34817710d4459d1d894234R56 also be `ValidationError` instead of `AuthError`?

app/services/gravity_service.rb
ashkan18: yes, i wasn't sure about Gravity's errors, it's not really validation and agreed its probably something more like generic application error. lets follow up on separate PR though since this one needs some coordination with other repos.

ashkan18:
merging to release bunch of follow up PRs in other repos.

app/graphql/mutations/base_mutation.rb
alloy: @ashkan18 A great way to start clarifying that distinction in code is to no longer use the shorter ‘auth’, but either ‘authentication’ or ‘authorization’.

PR #173, opened 2018-09-12
Add capability to refund sales tax transactions

williardx:
### Problem
When we refund a _captured charge_ we need to send a reversed transaction to TaxJar to record the refund.

### Solution
Add capability to create refund transactions. This will eventually be implemented by using `PostSalesTaxRefundJob` in `OrderCancellationService`. However, since we're not currently refunding captured charges this is just the code to create refund transactions in TaxJar.

### What Changed
- Adds methods to handle creating refunds in TaxJar
- Creates `PostSalesTaxRefundJob` to process the refund job in the background
- Refactors shared parameters for TaxJar API calls into a single method `construct_tax_params`
- Compares country codes for `artsy_should_remit_taxes?` using Carmen's country code instead of a hard-coded string

### DIFF
--- a/app/services/sales_tax_service.rb
+++ b/app/services/sales_tax_service.rb
@@ -47,7 +47,7 @@ class SalesTaxService
   end
 
   def artsy_should_remit_taxes?
-    return false unless destination_address[:country] == 'US'
+    return false unless destination_address[:country] == Carmen::Country.coded('US').code
     REMITTING_STATES.include? destination_address[:state].downcase
   end
 
--- a/app/services/sales_tax_service.rb
+++ b/app/services/sales_tax_service.rb
@@ -47,27 +47,22 @@ class SalesTaxService
   private
 
   def fetch_sales_tax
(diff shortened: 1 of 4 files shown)


app/services/sales_tax_service.rb
```diff
@@ -39,35 +39,54 @@ def record_tax_collected
...
-      to_street: destination_address[:address],
-      shipping: UnitConverter.convert_cents_to_dollars(@shipping_total_cents)
-    )
+    @tax_client.tax_for_order(construct_tax_params)
+  end
+
+  def get_transaction(id)
+    @tax_client.show_order(id)
+  rescue Taxjar::Error::NotFound
+    nil
```
ashkan18: we should probably log this error for our own tracking. Another question about this is, do you want to attempt of refunding this transaction to show up in sidekiq as a failure (and possibly retry?) if we want, then maybe we should log the error and raise.

williardx:
Closing this out in favor of a new PR

PR #176, opened 2018-09-14
👋 Memcache and Cache

ashkan18:
# Problem
Originally we decided to cache Partner and Artwork data, but as we started testing, looks like caching mostly causes issue and leads to us using stalled data. We want to make sure we are using latest version of all these models when calling our main API.

# Solution
Remove caching.

# Followup
De-provision Memache instances assigned to Exchange.


joeyAghion:
Seems reasonable to me, as long as we still avoid _duplicate_ or unnecessary requests for the same data.

Lesson: add caching only when necessary?

ashkan18:
haha we thought we could cache artwork back then https://github.com/artsy/exchange/pull/79 ... things were simpler back then...

PR #178, opened 2018-09-17
Add Sidekiq check to our health endpoint

ashkan18:
Adding `worker_health` to our health endpoint response, based on our threshold it either returns `OK` or `OY`

### DIFF
--- a/app/controllers/api/health_controller.rb
+++ b/app/controllers/api/health_controller.rb
@@ -1,9 +1,13 @@
 module Api
   class HealthController < Api::BaseApiController
     skip_before_action :authenticate_request!
+    SIDEKIQ_ALERT_THRESHOLD = 100
 
     def index
-      render json: { healthy: true }
+      render json: {
+        healthy: true,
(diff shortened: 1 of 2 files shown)


app/controllers/api/health_controller.rb
```diff
@@ -1,9 +1,13 @@
 module Api
   class HealthController < Api::BaseApiController
     skip_before_action :authenticate_request!
+    SIDEKIQ_ALERT_THRESHOLD = 100
 
     def index
-      render json: { healthy: true }
+      render json: {
+        healthy: true,
+        worker_health: Sidekiq::Queue.new.size < SIDEKIQ_ALERT_THRESHOLD ? 'OK' : 'OY'
```
joeyAghion: 👍 this is helpful for simple health-checks that can match strings.

You also could include the actual queue size here as a separate property, for humans and more sophisticated bots.

ashkan18:
follow up, update pingdom

PR #179, opened 2018-09-17
Update exchange admin overview

ansor4:
Resolves https://artsyproduct.atlassian.net/browse/PLATFORM-842

Delete the dashboard view
Rename Partner/User to Seller/Buyer
Remove Seller/Buyer id columns, add Fulfillment deadline column


app/models/order.rb
```diff
@@ -65,7 +65,9 @@ class Order < ApplicationRecord
   before_save :set_currency_code
 
   scope :pending, -> { where(state: PENDING) }
-  scope :active, -> { where(state: [SUBMITTED, APPROVED]) }
+  scope :active, -> { where.not(state: [ Order::ABANDONED, Order::PENDING, Order::CANCELED ]) }
```
ashkan18: lets work together to add some spec for these new scopes 🙌

PR #180, opened 2018-09-18
Catch all

ashkan18:
# Problem
Currently `RuntimeError`s and any other error that's not rescued results in a non-json response which leads to MP returning a useless `unexpected end of JSON` error.

# Solution
Add a `rescue_from StandardError` as a catch all option which returns `500` as http status code and returns a json error with type `internal` and code `server` with data being `message: exception.message`.

# Other changes
Made sure existing `rescue_from`s also return `ApplicationErrorType` with `code`, `type` and `data`.

# Example Response
```json
{
  "errors": [
    {
      "code": "server",
      "data": {
        "message": "undefined method `[]' for nil:NilClass"
      },
      "type": "internal"
    }
  ]
}

```

### DIFF
--- a/app/graphql/types/application_error_type.rb
+++ b/app/graphql/types/application_error_type.rb
@@ -14,7 +14,7 @@ class Types::ApplicationErrorType < Types::BaseObject
     error_type(type: :internal, code: :server, data: { message: err.message })
   end
 
-  def self.error_type(type:, code:, data: {})
+  def self.error_type(type:, code:, data: nil)
     {
       code: code,
       data: data,
--- a/app/graphql/types/application_error_type.rb
+++ b/app/graphql/types/application_error_type.rb
@@ -7,18 +7,18 @@ class Types::ApplicationErrorType < Types::BaseObject
   field :type, String, null: false, description: 'Type of this error'
 
   def self.from_application(err)
-    {
-      code: err.code,
-      data: err.data.to_json,
-      type: err.type.to_s
-    }
+    error_type(type: err.type, code: err.code, data: err.data)
   end
 
   def self.from_exception(err)
+    error_type(type: :internal, code: :server, data: { message: err.message })
+  end
+
+  def self.error_type(type:, code:, data: {})
     {
-      code: :server,
-      data: { message: err.message },
-      type: :internal
+      code: code,
+      data: data,
+      type: type
     }
   end
 end
--- a/app/graphql/types/application_error_type.rb
+++ b/app/graphql/types/application_error_type.rb
@@ -10,8 +10,8 @@ class Types::ApplicationErrorType < Types::BaseObject
     format_error_type(type: err.type, code: err.code, data: err.data)
   end
 
-  def self.from_exception(err)
-    format_error_type(type: :internal, code: :server, data: { message: err.message })
+  def self.from_generic_exception(err)
+    format_error_type(type: :internal, code: :generic, data: { message: err.message })
   end
 
   def self.format_error_type(type:, code:, data: nil)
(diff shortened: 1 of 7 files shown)


app/graphql/types/application_error_type.rb
```diff
@@ -7,10 +7,18 @@ class Types::ApplicationErrorType < Types::BaseObject
...
 
   def self.from_application(err)
+    error_type(type: err.type, code: err.code, data: err.data)
+  end
+
+  def self.from_exception(err)
+    error_type(type: :internal, code: :server, data: { message: err.message })
+  end
+
+  def self.error_type(type:, code:, data: nil)
```
sweir27: The naming here reads a little weirdly (`ApplicationErrorType.error_type...`)

Maybe `format_error_type` is more accurate?

config/database.yml
```diff
@@ -26,7 +26,7 @@ default: &default
 
 development:
   <<: *default
-  database: exchange_development
+  database: <%= ENV['DATABASE_NAME'] || 'exchange_development' %>
```
sweir27: 😉

spec/controllers/api/requests/error_handling_request_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+        expect(error['data']['message']).to eq 'something went wrong'
+      end
+    end
+
+    context 'ActiveRecord::RecordNotFound' do
+      before do
+        allow(CreateOrderService).to receive(:with_artwork!).and_raise(ActiveRecord::RecordNotFound, 'cannot find')
+        post '/api/graphql', params: { query: mutation, variables: { input: mutation_input } }, headers: auth_headers
+      end
+      it 'returns 500' do
```
sweir27: `'returns 404'`

spec/controllers/api/requests/error_handling_request_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+        expect(error['data']['message']).to eq 'cannot find'
+      end
+    end
+
+    context 'ActionController::ParameterMissing' do
+      before do
+        allow(CreateOrderService).to receive(:with_artwork!).and_raise(ActionController::ParameterMissing, 'id')
+        post '/api/graphql', params: { query: mutation, variables: { input: mutation_input } }, headers: auth_headers
+      end
+      it 'returns 500' do
```
sweir27: `'returns 400'`

app/graphql/types/application_error_type.rb
```diff
@@ -7,10 +7,18 @@ class Types::ApplicationErrorType < Types::BaseObject
   field :type, String, null: false, description: 'Type of this error'
 
   def self.from_application(err)
+    error_type(type: err.type, code: err.code, data: err.data)
+  end
+
+  def self.from_exception(err)
+    error_type(type: :internal, code: :server, data: { message: err.message })
```
sweir27: `type: :internal` and `code: :server` seem to say the same thing What other types of errors would you expect that would have the same `type`?

config/database.yml
ashkan18: this makes connecting to staging db easier if needed

app/graphql/types/application_error_type.rb
ashkan18: maybe we could have `type: internal` and `code: runtime`? since this is coming from catch all i can't think of any other types or codes.

ashkan18: renamed code to `generic` we can revisit in follow up PRs 👍 going to merge

PR #182, opened 2018-09-19
Extra columns, minor formatting to exchange admin overview

ansor4:
Changes:
- Change `Fulfillment Deadline` to `Fulfillment Countdown`
- Change set timestamp to countdown (or was the set timestamp ok?)
- Add `Submitted At` and `Last Updated` columns
- Add `Approval Countdown` column
- `Total` displayed in dollars and cents


ansor4:
Yeah, we had confirmed that was ok with Madeleine right? I was going off the spec but then forgot about that exchange and blindly made the change

ashkan18:
If we could do the time diff on the client side it would be really cool! For now I'd vote for removing it to avoid possible confusion. Feel free to merge when ready 🙌 🙌

PR #183, opened 2018-09-19
Fix setShipping to pickup issue

ashkan18:
# Problem
Currently setting order requested fulfillment to Pickup on staging fails with address validation. Follow up on #175 

# Cause
We are validating `shipping` input even if type is pickup, clients are currently sending `region: ''` and that fails checking region requirements.

# Solution
Don't set/validate shipping address for pickup orders.


sweir27:
Thanks!!

PR #184, opened 2018-09-19
Mo error handling less problems

ashkan18:
# Changes

## Errors::ERROR_TYPES
Added `Errors::ERROR_TYPES` as a Hash with keys being all current possible error `types` and each type has a list of possible `codes` for that type. On creating `ApplicationError` we check if type and code are valid. This way `Errors::ErrorTypes` should be our source of truth for all current errors.

## Root Level Errors
In this PR we also updated root level errors to match [GraphQL specs](http://facebook.github.io/graphql/June2018/#example-fce18) where we are required to provide a `message` and we can add our own specific fields under `extensions`.

## Errors::InternalError
Used `Errors::InternalError` for things like Gravity's generic errors.

### DIFF
--- a/app/controllers/errors/error_types.rb
+++ b/app/controllers/errors/error_types.rb
@@ -15,6 +15,7 @@ module Errors
       invalid_state
       invalid_order
       missing_artwork_location
+      missing_partner_location
       missing_commission_rate
       missing_country
       missing_currency
@@ -33,7 +34,7 @@ module Errors
     ],
     processing: %i[
       capture_failed
-      failed_charge_authorization
+      charge_authorization_failed
       insufficient_inventory
       refund_failed
     ],
--- a/app/controllers/errors/error_types.rb
+++ b/app/controllers/errors/error_types.rb
@@ -13,6 +13,7 @@ module Errors
       invalid_commission_rate
       invalid_seller_address
       invalid_state
+      invalid_order
       missing_artwork_location
       missing_commission_rate
       missing_country
--- a/app/controllers/errors/error_types.rb
+++ b/app/controllers/errors/error_types.rb
@@ -1,0 +1,44 @@
+module Errors
+  ERROR_TYPES = {
+    auth: %i[
+      not_found
+    ],
+    validation: %i[
+      credit_card_deactivated
+      credit_card_missing_customer
+      credit_card_missing_external_id
+      credit_card_not_found
+      failed_order_code_generation
+      invalid_artwork_address
+      invalid_commission_rate
+      invalid_seller_address
+      invalid_state
+      missing_artwork_location
+      missing_commission_rate
+      missing_country
+      missing_currency
+      missing_domestic_shipping_fee
+      missing_international_shipping_fee
+      missing_merchant_account
+      missing_params
+      missing_postal_code
+      missing_price
+      missing_region
+      missing_required_info
+      not_found
+      unknown_artwork
+      unknown_edition_set
+      unknown_partner
+    ],
+    processing: %i[
+      capture_failed
+      failed_charge_authorization
+      insufficient_inventory
+      refund_failed
+    ],
+    internal: %i[
+      generic
+      gravity
+    ]
+  }.freeze
+end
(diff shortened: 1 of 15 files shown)


app/controllers/errors/error_types.rb
```diff
@@ -0,0 +1,45 @@
...
+      missing_currency
+      missing_domestic_shipping_fee
+      missing_international_shipping_fee
+      missing_merchant_account
+      missing_params
+      missing_postal_code
+      missing_price
+      missing_region
+      missing_required_info
+      not_found
```
sweir27: What's this one? A basic 404 doesn't really seem like _validation_, it's more of a generic error type.

app/graphql/types/application_error_type.rb
```diff
@@ -10,15 +10,28 @@ def self.from_application(err)
...
+
+  def self.root_level_error_from_exception(err)
+    format_root_level_error(type: :internal, code: :generic, data: { message: err.message }, message: err.message)
+  end
+
+  # For root level errors expect a message key and our detail data should be under extension
+  def self.format_root_level_error(type:, code:, data: nil, message: nil)
+    message ||= "type: #{type}, code: #{code}, data: #{data}"
+    {
+      message: message,
```
sweir27: This looks like it's going to be another breaking change?

I wonder if we can just keep the `code`, `data`, and `type`, and add `message` as a separate thing. Even though clients will be interpreting the type and often showing their own text, a message from the server can sometimes still be helpful. (i.e. `"Artwork not found"`, `"Couldn't validate order"`... more of a human-readable version in case someone's looking at the response not through a UI).

app/controllers/errors/error_types.rb
```diff
@@ -0,0 +1,45 @@
...
+      missing_country
+      missing_currency
+      missing_domestic_shipping_fee
+      missing_international_shipping_fee
+      missing_merchant_account
+      missing_params
+      missing_postal_code
+      missing_price
+      missing_region
+      missing_required_info
```
starsirius: We can iterate on these error types and make it super clear that when we see the error type, we know immediately what's wrong.

app/controllers/errors/error_types.rb
```diff
@@ -0,0 +1,45 @@
...
+      not_found
+      unknown_artwork
+      unknown_edition_set
+      unknown_partner
+    ],
+    processing: %i[
+      capture_failed
+      failed_charge_authorization
```
starsirius: #minor. The naming can be consistent for the above two.

app/graphql/types/application_error_type.rb
starsirius: Looks like a breaking change, but I think the goal here is to implement the error format according to the GraphQL [spec](http://facebook.github.io/graphql/June2018/#sec-Errors), which other graphql client library might be following:

1. Every error must contain an entry with the key `message` with a string description.
2. An optional `locations`
3. An optional `path`
4. An optional `extensions`
5. GraphQL services should not provide any additional entries to the error format.

app/services/order_submit_service.rb
```diff
@@ -39,11 +39,11 @@ def process!
   private
 
   def pre_process!
+    raise Errors::ValidationError, :missing_required_info unless can_submit?
     @order.line_items.map do |li|
       artwork = GravityService.get_artwork(li[:artwork_id])
-      raise Errors::ProcessingError, :artwork_version_missmatch if artwork[:current_version_id] != li[:artwork_version_id]
+      raise Errors::ProcessingError, :artwork_version_mismatch if artwork[:current_version_id] != li[:artwork_version_id]
```
starsirius: 👍

app/controllers/errors/error_types.rb
ashkan18: so this is the tricky part, there is `not_found` under `auth` because a user doesn't have access to an object, and there is a `not_found` for when an object doesn't exist at all. Whats the best way to separate these two cases? or we don't want to separate them at all from client's perspective?

app/graphql/types/application_error_type.rb
ashkan18: talked to @sweir27 about it, it won't be a breaking change since purchase side clients are currently not checking these errors and for the most part these errors are handled by clients being used `Relay` and `Graphlient`. In my case I actually noticed the problem when `Graphlient` complained about missing `message`.

ashkan18:
Going to merge this and we can followup on our discussions on separate PR 🙌

PR #186, opened 2018-09-20
Sort by state updated at

jonallured:
This PR adds the ability to sort the orders list by the `state_updated_at` timestamp. I started by adding it to the `Types::OrderConnectionSortEnum` and getting a red test. From there I added the code to get back to green which was just adding to the case statement that I found in `Types::QueryType#orders`, but it occurred to me that with a little extract method, I'd be able to support future sorting options that follow the format of `[attribute_name]_[direction]` so I did that with a894e3e. Along the way I also did some work on making the specs a little more consistent by extracting a helper to work with ids in the response data so that comparing them to the ids of fabricated data is easier.

I'm not sure if this is enough - I guess I need to update the schema and propagate that change to MP? Not really sure how to do that so I'd love some direction on that front! ❤️

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -33,6 +33,10 @@ class Types::QueryType < Types::BaseObject
       query.order(updated_at: :asc)
     when 'UPDATED_AT_DESC'
       query.order(updated_at: :desc)
+    when 'STATE_UPDATED_AT_ASC'
+      query.order(state_updated_at: :asc)
+    when 'STATE_UPDATED_AT_DESC'
+      query.order(state_updated_at: :desc)
     else
       query
     end
--- a/app/graphql/types/order_connection_sort_enum.rb
+++ b/app/graphql/types/order_connection_sort_enum.rb
@@ -2,6 +2,8 @@ class Types::OrderConnectionSortEnum < Types::BaseEnum
   description 'Fields to sort by'
   graphql_name 'OrderConnectionSortEnum'
 
-  value 'UPDATED_AT_ASC', 'Sort by latest timestamp order was updated in ascending order'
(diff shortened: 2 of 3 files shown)


app/graphql/types/query_type.rb
```diff
@@ -27,19 +27,21 @@ def orders(params)
...
-    else
-      query
-    end
+    order_by = order_by(sort)
+    query.order(order_by)
   end
 
   private
 
+  def order_by(sort)
```
ashkan18: I like moving the order logic to it's own method. On the other hand, I prefer the `case` statement approach over this current version for two reasons:
1) current approach assumes a very specific format that's not really exposed anywhere and limits us to that format so we can't do things like `last_approved_at` and `last_submitted_at` discussed https://github.com/artsy/volt/pull/3212#discussion_r219295102
2) case approach seems more readable and flexible and allows for possible multi-column sorting options.

if we want to make it more dynamic and limited to only column names we should probably do something like https://github.com/artsy/gravity/blob/master/app/api/util/sort_helpers.rb where we check for existence of the column and etc.

jonallured: ok, fair enough - i've removed the extraction 👍

ashkan18: thanks 🙌

ashkan18:
Lets do this 🙌 🎉 📉 📈

ashkan18:
Followup on this. For this to work from MP side of thing we need to:
- Follow https://github.com/artsy/exchange#did-you-change-graphql-schema
- Update https://github.com/artsy/metaphysics/blob/master/src/schema/ecommerce/types/orders_sort_method_enum.ts to include new options

jonallured:
☝️ excellent docs, thanks for writing that up!! ❤️

PR #189, opened 2018-09-24
Update libs and Graphlient

ashkan18:
# Changes
Update to latest `graphlient` where we get more details about `ServerError`s. 

Also update other gems, `rubocop` update required adding extra line after every one line `if`.


greysteil:
@ashkan18 - would you be up for using [Dependabot](https://dependabot.com) on this repo for gem updates? I built it, and am trying to get more open source repos using it because the data is really useful for our compatibility scores ([here's](https://dependabot.com/compatibility-score/?dependency-name=rubocop&new-version=0.59.1&package-manager=bundler&previous-version=0.58.2) the one for the rubocop upgrade you did here, for example).

I chatted with @orta about it a year ago, when you were using DependencyCI to support Andrew Nesbitt's work, but DependencyCI is owned by TideLift now, and it doesn't look like you're using it anymore.

🙏

PR #190, opened 2018-09-24
Remove NJ from the list of states where we should remit sales taxes

williardx:
Based on continuing conversations around sales tax with business and legal stakeholders, this removes NJ from the list of states where we should remit sales taxes.


ashkan18:
nice and easy 🙌

PR #191, opened 2018-09-24
More CreateOrder errors Raise error

ashkan18:
# Change
Make sure we raise error when artwork was:
- unpublished
- disabled for ecommerce

Updated https://www.notion.so/artsy/Error-and-failure-states-for-Orders-b3dfcdef33bc4f2b9a1f912568cb3c8e with the changes

### DIFF
--- a/app/services/create_order_service.rb
+++ b/app/services/create_order_service.rb
@@ -2,6 +2,8 @@ module CreateOrderService
   def self.with_artwork!(user_id:, artwork_id:, edition_set_id: nil, quantity:)
     artwork = GravityService.get_artwork(artwork_id)
     raise Errors::ValidationError.new(:unknown_artwork, artwork_id: artwork_id) if artwork.nil? || (edition_set_id && !find_edition_set(artwork, edition_set_id))
+    raise Errors::ValidationError.new(:unpublished_artwork, artwork_id: artwork_id) unless artwork[:published]
+    raise Errors::ValidationError.new(:disabled_ecommerce, artwork_id: artwork_id) unless artwork[:ecommerce]
 
     Order.transaction do
       order = Order.create!(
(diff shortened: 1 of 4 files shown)


app/services/create_order_service.rb
```diff
@@ -2,6 +2,8 @@ module CreateOrderService
   def self.with_artwork!(user_id:, artwork_id:, edition_set_id: nil, quantity:)
     artwork = GravityService.get_artwork(artwork_id)
     raise Errors::ValidationError.new(:unknown_artwork, artwork_id: artwork_id) if artwork.nil? || (edition_set_id && !find_edition_set(artwork, edition_set_id))
+    raise Errors::ValidationError.new(:unpublished_artwork, artwork_id: artwork_id) unless artwork[:published]
+    raise Errors::ValidationError.new(:disabled_ecommerce, artwork_id: artwork_id) unless artwork[:ecommerce]
```
sweir27: Should we be testing for `acquireable` here?

ashkan18: oooh good point! do you think i should do that as an extra thing or just check for that?

sweir27: it includes `ecommerce` so I think it should be fine?

app/services/create_order_service.rb
```diff
@@ -2,6 +2,8 @@ module CreateOrderService
   def self.with_artwork!(user_id:, artwork_id:, edition_set_id: nil, quantity:)
     artwork = GravityService.get_artwork(artwork_id)
     raise Errors::ValidationError.new(:unknown_artwork, artwork_id: artwork_id) if artwork.nil? || (edition_set_id && !find_edition_set(artwork, edition_set_id))
+    raise Errors::ValidationError.new(:unpublished_artwork, artwork_id: artwork_id) unless artwork[:published]
+    raise Errors::ValidationError.new(:disabled_ecommerce, artwork_id: artwork_id) unless artwork[:acquireable]
```
sweir27: Agh #minor but we should probably rename the error to something that reflects the new logic, maybe `:not_acquireable`?

PR #193, opened 2018-09-25
Use Address object in RecordSalesTaxJob

williardx:
### Problem
https://github.com/artsy/exchange/issues/185

### Solution
Pass an `Address` object to `SalesTaxService` in `RecordSalesTaxJob`

### What changed
- Updates `RecordSalesTaxJob` and specs to use `Address` object as required by `SalesTaxService`
- Overrides `Address.==` to consider two `Address` objects to be equal if their corresponding address attributes are equal. This was mostly done so that we can compare address objects in testing.

### DIFF
--- a/lib/address.rb
+++ b/lib/address.rb
@@ -12,6 +12,15 @@ class Address
     @postal_code = @address[:postal_code]
   end
 
+  def ==(other)
+    @country == other.country && \
+      @region == other.region && \
+      @city == other.city && \
+      @street_line1 == other.street_line1 && \
+      @street_line2 == other.street_line2 && \
+      @postal_code == other.postal_code
+  end
+
   private
 
   def parse(address)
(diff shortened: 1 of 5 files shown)


lib/address.rb
```diff
@@ -12,6 +12,15 @@ def initialize(address)
     @postal_code = @address[:postal_code]
   end
 
+  def ==(other)
```
ashkan18: nice 👍

lib/address.rb
```diff
@@ -12,6 +12,15 @@ def initialize(address)
...
+    @country == other.country && \
+      @region == other.region && \
+      @city == other.city && \
+      @street_line1 == other.street_line1 && \
+      @street_line2 == other.street_line2 && \
```
ashkan18: do you think `street_line2` is needed fo matching addresses? is Apt B1 same as Apt B?

williardx: Good question -- I can see this going either way. I don't think Apt B1 is the same as B, but for most of our purposes we really only care about `street_line1`. Comparing line 2 feels more correct to me but might be overly rigid. I'll leave it in for now and if it's a problem later on I'll change it, especially since we're only using it in testing anyway.

PR #195, opened 2018-09-25
Remove last admin note attribute/filter

ansor4:
Remove this for now, we can add it back in later if we need it


ashkan18:
merge on green

PR #196, opened 2018-09-26
Expose state_reason in graphql

sepans:
## Problem
`state_reason` existed in the model but not exposed through graphql. 

## Solution
Adds `stateReason` to graphql schema

screen shot 2018-09-26 at 9 40 41 am
(used exchange `rejectOrder` to reject order above)


app/graphql/types/order_type.rb
```diff
@@ -8,6 +8,7 @@ class Types::OrderType < Types::BaseObject
   field :seller, Types::OrderPartyUnionType, null: false
   field :credit_card_id, String, null: true
   field :state, Types::OrderStateEnum, null: false
+  field :state_reason, String, null: true
```
ashkan18: do you think we should make it `seller_only: true` ?

sepans: Nice! I didn't notice we have such an option. I remember @erikdstock and @zephraph wanted to use this field somewhere so I am guessing it appears on buyer pages as well.

PR #197, opened 2018-09-26
Add admin note functionality to exchange

ansor4:
for https://artsyproduct.atlassian.net/browse/PLATFORM-844

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -17,7 +17,9 @@ ActiveAdmin.register Order do
   filter :state_reason, as: :check_boxes, collection: proc { Order::REASONS.values.map(&:values).flatten.uniq.map!(&:humanize) }
 
   index do
-    column :id
+    column :id do |order|
+      link_to order.id, admin_order_path(order.id)
+    end
     column :state
     column :fulfillment_type
     column :last_admin_note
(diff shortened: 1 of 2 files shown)


app/admin/orders.rb
```diff
@@ -32,5 +34,123 @@
...
+        column :description
+      end
+    end
+
+  end
+
+  sidebar :contact_info, only: :show do
+
+    #TODO: why doesn't this work?
+    link_to "artsy.net" do
```
ashkan18: we can use helpers here, we could define something like `artsy_artwork_path` method and pass it a line item and have it return the artwork link.
example: https://github.com/artsy/impulse/blob/master/app/admin/conversation.rb#L50
https://github.com/artsy/impulse/blob/71da5ac1a3a02569480762da5592fd93074a15c3/app/helpers/url_helper.rb
https://github.com/artsy/impulse/blob/71da5ac1a3a02569480762da5592fd93074a15c3/config/initializers/active_admin.rb#L21

app/admin/orders.rb
```diff
@@ -32,5 +34,123 @@
...
+        row :partner_name
+        row :address
+        row :phone
+        row :email
+        row :sales_contacts do
+          #TODO: add sales contacts
+          #table_for
+        end
+      end
+      link_to "artsy.net" do
```
ashkan18: same thing as artwork path example above.

PR #198, opened 2018-09-27
More edition set handling

ashkan18:
# Problem
Currently Eigen doesn't pass `edition_set_id` to `createOrderWithArtwork` mutation when there is one edition set. We need Exchange to check if artwork has only one edition set, assume that was the one that buyer meant to pick. In case there are more than one edition sets and `edition_set_id` was not passed, we now raise a new `validation` error with code `missing_edition_set_id`.

# Changes
`CreateOrderService` was getting more complicated with these changes and as part of this change we did a little refactoring there where followed our new style. We now create a new instance of `CreateOrderService` and we call `process!` on it. We also expose created `order` with `attr_reader` on that class.
New edition set logic is:
- If `edition_set_id` was passed to the service
  - we check if edition set exists on artwork we use that, otherwise we raise error
- else
  - we check if artwork has edition sets
     - if artwork has more than one edition set, we raise error since we don't know which one buyer meant to buy
     - if artwork has one edition set, we make assumption that buyer meant to buy that one 

# New Error
- type: `validation`, code: `missing_edition_set_id`, data: `artwork_id`: happens when artwork has more than one edition sets and we don't provide which one we want during order creation.

# TODO
Once Eigen starts sending edition set id in these cases we can eventually deprecate the logic above that assumes edition set.

### DIFF
--- a/app/services/create_order_service.rb
+++ b/app/services/create_order_service.rb
@@ -45,6 +45,7 @@ class CreateOrderService
     raise Errors::ValidationError.new(:unknown_artwork, artwork_id: @artwork_id) if @artwork.nil?
     raise Errors::ValidationError.new(:unpublished_artwork, artwork_id: @artwork_id) unless @artwork[:published]
     raise Errors::ValidationError.new(:not_acquireable, artwork_id: @artwork_id) unless @artwork[:acquireable]
+
     find_verify_edition_set
   end
 
@@ -53,7 +54,7 @@ class CreateOrderService
   end
 
   def artwork_price
-    item = @edition_set.present? ? @edition_set : @artwork
+    item = @edition_set.presence || @artwork
     raise Errors::ValidationError, :missing_price unless item[:price_listed]&.positive?
 
     raise Errors::ValidationError, :missing_currency if item[:price_currency].blank?
@@ -73,7 +74,8 @@ class CreateOrderService
       # if there is one we are going to assume thats the one buyer meant to buy
       # TODO: ☝ is a temporary logic till Eigen starts supporting editionset artworks
       return unless @artwork[:edition_sets]&.count
-      raise Errors::ValidationError.new(:missing_edition_set_id, artwork_id: @artwork[:id]) if @artwork[:edition_sets]&.count > 1
+      raise Errors::ValidationError.new(:missing_edition_set_id, artwork_id: @artwork_id) if @artwork[:edition_sets].present? && @artwork[:edition_sets].count > 1
+
       @edition_set = @artwork[:edition_sets].first
       @edition_set_id = @edition_set[:id]
     end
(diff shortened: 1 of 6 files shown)


app/services/create_order_service.rb
```diff
@@ -50,7 +64,20 @@ def self.artwork_price(external_artwork, edition_set_id: nil)
...
+    if @edition_set_id
+      @edition_set = @artwork[:edition_sets]&.find { |e| e[:id] == @edition_set_id }
+      raise Errors::ValidationError.new(:unknown_edition_set, artwork_id: @artwork[:id], edition_set_id: @edition_set_id) unless @edition_set
+    else
+      # If artwork has EditionSet but it was not passed in the request
+      # if there are more than one EditionSet we'll raise error
+      # if there is one we are going to assume thats the one buyer meant to buy
+      # TODO: ☝ is a temporary logic till Eigen starts supporting editionset artworks
+      return unless @artwork[:edition_sets]&.count
+      raise Errors::ValidationError.new(:missing_edition_set_id, artwork_id: @artwork_id) if @artwork[:edition_sets].present? && @artwork[:edition_sets].count > 1
```
sweir27: You're already checking for `@artwork[:edition_sets]&.count` above so I think this can just be `if @artwork[:edition_sets].count > 1`

app/services/create_order_service.rb
```diff
@@ -50,7 +64,20 @@ def self.artwork_price(external_artwork, edition_set_id: nil)
...
-    external_artwork[:edition_sets].find { |e| e[:id] == edition_set_id }
+  def find_verify_edition_set
+    if @edition_set_id
+      @edition_set = @artwork[:edition_sets]&.find { |e| e[:id] == @edition_set_id }
+      raise Errors::ValidationError.new(:unknown_edition_set, artwork_id: @artwork[:id], edition_set_id: @edition_set_id) unless @edition_set
+    else
+      # If artwork has EditionSet but it was not passed in the request
+      # if there are more than one EditionSet we'll raise error
+      # if there is one we are going to assume thats the one buyer meant to buy
+      # TODO: ☝ is a temporary logic till Eigen starts supporting editionset artworks
```
ashfurrow: We could add a link to https://artsyproduct.atlassian.net/browse/PURCHASE-505 in this comment.

ashkan18: good point! adding

ashkan18:
thanks @ashfurrow , merging

app/services/create_order_service.rb
```diff
@@ -50,7 +64,21 @@ def self.artwork_price(external_artwork, edition_set_id: nil)
...
+      # https://artsyproduct.atlassian.net/browse/PURCHASE-505
+      return unless @artwork[:edition_sets]&.count
```
sweir27: I think this will _always_ pass through, and so for artworks with `0` edition sets, on line 81 we'll be attempting to access `[:id]` on `null`. This should probably check that the count is greater than 0.

Meta-point but we may want to return from this method immediately (before either `if` statement`) if the artwork has 0 edition sets. It feels a little odd to be doing that down here.

ashkan18: oh good catch! the issue is we can't do `@artwork[:edition_sets]&.count > 0` either since `nil` doesn't have `>` method. 

Let me look at it and add more spec, i think our current spec sends `nil`

PR #199, opened 2018-09-27
Refactor RecordSalesTaxJob

williardx:
### Problem
`RecordSalesTaxJob` was failing on pickup orders with unhelpful error messages.

### Solution
- Refactor `RecordSalesTaxJob` to not try to create an address for pickup orders
- Add `tax_recording_failure` code to error types
- Store the transaction ID that we get back upon creating a transaction in TaxJar so that we can reference it later

### What changed
- Adds string `sales_tax_transaction_id` to `line_items`
- Update `SalesTaxService` to store response from transaction POST request to TaxJar in `@transaction`
- Updates `line_item.sales_tax_transaction_id` after receiving transaction response from TaxJar
- Return early in `RecordSalesTaxJob` if sales tax does not need to be remitted
- Add method `shipping_address` to `Order` model that returns the shipping address wrapped in an `Address` object for shipped orders or `nil` otherwise.

### DIFF
--- a/app/jobs/record_sales_tax_job.rb
+++ b/app/jobs/record_sales_tax_job.rb
@@ -3,7 +3,7 @@ class RecordSalesTaxJob < ApplicationJob
 
   def perform(line_item_id)
     line_item = LineItem.find(line_item_id)
-    return unless line_item.should_remit_sales_tax
+    return unless line_item.should_remit_sales_tax?
 
     artwork = GravityService.get_artwork(line_item.artwork_id)
     artwork_address = Address.new(artwork[:location])
--- a/app/jobs/record_sales_tax_job.rb
+++ b/app/jobs/record_sales_tax_job.rb
@@ -3,17 +3,12 @@ class RecordSalesTaxJob < ApplicationJob
 
   def perform(line_item_id)
     line_item = LineItem.find(line_item_id)
+    return unless line_item.should_remit_sales_tax
+
     artwork = GravityService.get_artwork(line_item.artwork_id)
     artwork_address = Address.new(artwork[:location])
-    shipping_address = Address.new(
-      country: line_item.order.shipping_country,
-      postal_code: line_item.order.shipping_postal_code,
-      region: line_item.order.shipping_region,
-      city: line_item.order.shipping_city,
-      address_line1: line_item.order.shipping_address_line1
-    )
-    service = SalesTaxService.new(line_item, line_item.order.fulfillment_type, shipping_address, line_item.order.shipping_total_cents, artwork_address)
+    service = SalesTaxService.new(line_item, line_item.order.fulfillment_type, line_item.order.shipping_address, line_item.order.shipping_total_cents, artwork_address)
     service.record_tax_collected
-    line_item.update!(sales_tax_transaction_id: service.transaction.transaction_id)
+    line_item.update!(sales_tax_transaction_id: service.transaction.transaction_id) if service.transaction.present?
   end
 end
--- a/app/jobs/record_sales_tax_job.rb
+++ b/app/jobs/record_sales_tax_job.rb
@@ -12,7 +12,8 @@ class RecordSalesTaxJob < ApplicationJob
       city: line_item.order.shipping_city,
       address_line1: line_item.order.shipping_address_line1
     )
-    transaction = SalesTaxService.new(line_item, line_item.order.fulfillment_type, shipping_address, line_item.order.shipping_total_cents, artwork_address).record_tax_collected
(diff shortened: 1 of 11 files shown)


app/jobs/record_sales_tax_job.rb
```diff
@@ -12,6 +12,7 @@ def perform(line_item_id)
       city: line_item.order.shipping_city,
       address_line1: line_item.order.shipping_address_line1
     )
-    SalesTaxService.new(line_item, line_item.order.fulfillment_type, shipping_address, line_item.order.shipping_total_cents, artwork_address).record_tax_collected
+    transaction = SalesTaxService.new(line_item, line_item.order.fulfillment_type, shipping_address, line_item.order.shipping_total_cents, artwork_address).record_tax_collected
```
ashkan18: one thing i've started doing in https://github.com/artsy/exchange/pull/198 was to have service have a `process!` method that doesn't return and then if there is something we need from the service we can have a `attr_reader` for it so maybe we could do something like
```ruby
service = SalesTaxService.new(line_item, line_item.order.fulfillment_type, shipping_address, line_item.order.shipping_total_cents, artwork_address)
service.process!
transaction = service.transaction
```
what do you think?

williardx: That `process` pattern doesn't work as easily here because `SalesTaxService` does two things: calculates sales tax and records sales tax transactions. But it makes sense to me to set the sales tax object and sales tax transaction objects we get back as attributes on the class.

williardx: Actually looks like we already do that with sales tax but I'll go ahead and add in the transaction

ashkan18: awesome! thanks

williardx:
Adding #DoNotMerge tag since I'm going to combine these changes with some other changes to fix errors when recording sales tax.

williardx:
WIP label is now off 😄

app/jobs/record_sales_tax_job.rb
```diff
@@ -3,15 +3,12 @@ class RecordSalesTaxJob < ApplicationJob
 
   def perform(line_item_id)
     line_item = LineItem.find(line_item_id)
+    return unless line_item.should_remit_sales_tax
```
ashkan18: #minor in Ruby methods returning boolean usually end with `?` we could rename this to `line_item.remit_sales_tax?`

app/jobs/record_sales_tax_job.rb
```diff
@@ -3,15 +3,12 @@ class RecordSalesTaxJob < ApplicationJob
...
     artwork_address = Address.new(artwork[:location])
-    shipping_address = Address.new(
-      country: line_item.order.shipping_country,
-      postal_code: line_item.order.shipping_postal_code,
-      region: line_item.order.shipping_region,
-      city: line_item.order.shipping_city,
-      address_line1: line_item.order.shipping_address_line1
-    )
-    SalesTaxService.new(line_item, line_item.order.fulfillment_type, shipping_address, line_item.order.shipping_total_cents, artwork_address).record_tax_collected
+    service = SalesTaxService.new(line_item, line_item.order.fulfillment_type, line_item.order.shipping_address, line_item.order.shipping_total_cents, artwork_address)
```
ashkan18: we are passing `line_item` already to this service, do you think we still need to send all the other params?
this could be `SalesTaxService.new(line_item, artwork_address)`

app/jobs/record_sales_tax_job.rb
williardx: `should_remit_sales_tax` is the name of an attribute in the `line_item` schema. Should we rename the attribute? Or is it more idiomatic to create a method `remit_sales_tax?` that just returns `should_remit_sales_tax`?

ashkan18: oooh missed that part, if its a field on that model, i think you can just add `?` and `ActiveRecord` already handles that

app/jobs/record_sales_tax_job.rb
williardx: Yes -- when we use `SalesTaxService` in `OrderShippingService` those parameters aren't yet saved to the line item.

I'm beginning to think that `SalesTaxService` should be split into two different classes `SalesTaxEstimationService` and `SalesTaxRecordingService`, but that's for another PR!

ashkan18: makes sense, this way we end up with two smaller services 👍 to ✌️

app/jobs/record_sales_tax_job.rb
williardx: Updated!

PR #200, opened 2018-09-27
Enable find Order by code

ashkan18:
# Problem

https://artsyproduct.atlassian.net/browse/PURCHASE-496
We want to be able to find orders by code.

# Solution
make `id` optional and add new `code` argument. Make sure one of them is required. 

# New Error
type: `validation`, code: `missing_required_param`, data: `message: id or code is required`

### DIFF
--- a/spec/controllers/api/requests/order_query_request_spec.rb
+++ b/spec/controllers/api/requests/order_query_request_spec.rb
@@ -31,8 +31,43 @@ describe Api::GraphqlController, type: :request do
 
     let(:query) do
       <<-GRAPHQL
-        query($id: ID!) {
+        query($id: ID) {
           order(id: $id) {
+            id
+            buyer {
+              ... on User {
+                id
+              }
+            }
+            seller {
+              ... on Partner {
+                id
+              }
+            }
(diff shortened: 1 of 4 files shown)


joeyAghion:
The "code" was intended to be friendly for humans (and reference-able, e.g., in support requests). I'm not sure it makes sense to use that in URLs, where a GUID might be more appropriate.

We should still have orders find-able by code, probably, to support admins and so on.

ashkan18:
There has been lot of back and forth on the error code, at the moment its _almost_ a GUID :) and should be url safe but we can always add a separate GUID too if there was a need.

spec/controllers/api/requests/order_query_request_spec.rb
```diff
@@ -89,6 +124,18 @@
         expect(result.data.order.created_at).to eq created_at.iso8601
       end
 
+      it 'returns order when accessing correc order by code' do
```
sweir27: `correct`

ashkan18:
🙌 merging, lets follow up on GUID as a separate thing

PR #202, opened 2018-09-28
Add Test Coverage Analysis #FF

ashkan18:
# FF
Added coveralls. We basically updated our default `rake` task to include pushing coveralls and also we now wrap our spec run process with coveralls.


spec/spec_helper.rb
```diff
@@ -1,3 +1,6 @@
+require 'coveralls'
+Coveralls.wear!('rails')
```
zephraph: 😄 clever naming on their part

ashkan18:
lets do this

PR #204, opened 2018-09-28
Enforce domestic shipping only validation on adding address to order

javamonn:
As part of [PURCHASE-371](https://artsyproduct.atlassian.net/browse/PURCHASE-371), we want to validate against adding a non-continental US shipping address to an artwork with a nil `international_shipping_fee_cents` i.e. the work ships to continental us only. Exchange will return an error of type `invalid_shipping_address`, which reaction will use to display an appropriate modal. Reaction already disables the country field from being changed in this case, so this validation logic should only trigger when a user attempts to enter an address in Alaska or Hawaii.

This is the first time I've written Ruby in a while, so feel free to nit anything and everything.

### DIFF
--- a/app/controllers/errors/error_types.rb
+++ b/app/controllers/errors/error_types.rb
@@ -35,6 +35,7 @@ module Errors
       unknown_edition_set
       unknown_partner
       unpublished_artwork
+      unsupported_shipping_location
       wrong_fulfillment_type
     ],
     processing: %i[
@@ -44,6 +45,7 @@ module Errors
       insufficient_inventory
       refund_failed
       tax_recording_failure
+      tax_calculator_failure
     ],
     internal: %i[
       generic
--- a/app/services/order_shipping_service.rb
+++ b/app/services/order_shipping_service.rb
@@ -10,6 +10,8 @@ class OrderShippingService
   def process!
     raise Errors::ValidationError.new(:invalid_state, state: @order.state) unless @order.state == Order::PENDING
 
+    validate_shipping_location!
+
     Order.transaction do
       attrs = {
         shipping_total_cents: shipping_total_cents,
@@ -58,6 +60,14 @@ class OrderShippingService
     raise Errors::ValidationError, :missing_artwork_location if artwork[:location].blank?
   end
 
+  def validate_shipping_location!
+    domestic_shipping_only = artworks.any? do |(_, artwork)|
+      artwork[:international_shipping_fee_cents].nil? && international_shipping?(artwork[:location])
+    end
+
+    raise Errors::ValidationError.new(:unsupported_shipping_location, failure_code: :domestic_shipping_only) if domestic_shipping_only
+  end
+
   def shipping_total_cents
     @shipping_total_cents ||= @order.line_items.map { |li| artwork_shipping_fee(artworks[li.artwork_id]) }.sum
   end
@@ -65,15 +75,20 @@ class OrderShippingService
   def artwork_shipping_fee(artwork)
     if @fulfillment_type == Order::PICKUP
       0
-    elsif domestic?(artwork[:location])
+    elsif domestic_shipping?(artwork[:location])
       calculate_domestic_shipping_fee(artwork)
     else
       calculate_international_shipping_fee(artwork)
     end
   end
 
-  def domestic?(artwork_location)
-    artwork_location[:country].casecmp(@shipping_address.country).zero?
+  def domestic_shipping?(artwork_location)
+    artwork_location[:country].casecmp(@shipping_address.country).zero? &&
+      (@shipping_address.country != Carmen::Country.coded('US').code || @shipping_address.continental_us?)
+  end
+
+  def international_shipping?(artwork_location)
+    !domestic_shipping?(artwork_location)
   end
 
   def calculate_domestic_shipping_fee(artwork)
--- a/spec/services/order_shipping_service_spec.rb
+++ b/spec/services/order_shipping_service_spec.rb
@@ -3,7 +3,9 @@ require 'support/gravity_helper'
 
 describe OrderShippingService, type: :services do
   let(:order) { Fabricate(:order) }
+  let(:continental_us_order) { Fabricate(:order) }
   let!(:line_item) { Fabricate(:line_item, order: order, artwork_id: 'a-1') }
+  let!(:continental_us_line_item) { Fabricate(:line_item, order: continental_us_order, artwork_id: 'a-2') }
   let(:domestic_shipping) do
     {
       address_line1: '401 Broadway',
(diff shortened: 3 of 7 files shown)


app/services/order_shipping_service.rb
```diff
@@ -10,6 +10,8 @@ def initialize(order, fulfillment_type:, shipping:)
   def process!
     raise Errors::ValidationError.new(:invalid_state, state: @order.state) unless @order.state == Order::PENDING
 
+    validate_shipping_location!
```
javamonn: Would this be better placed as a validation on the `Order` model? I've previously had mixed experiences with async validations but not sure what the best practice is.

app/controllers/errors/error_types.rb
```diff
@@ -13,6 +13,7 @@ module Errors
       invalid_commission_rate
       invalid_order
       invalid_seller_address
+      invalid_shipping_address
```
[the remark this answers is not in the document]
ashkan18: good question, each `ValidationError` also gets a `data` field along with `code` we could leave the code be generic, as you have now, and in the data field describe more details, what do you think?

app/controllers/errors/error_types.rb
```diff
@@ -42,6 +43,7 @@ module Errors
       insufficient_inventory
       refund_failed
       tax_recording_failure
+      tax_calculator_failure
```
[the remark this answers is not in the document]
ashkan18: good catch! yes requiring `code` be defined in this file was introduced in one PR and i think all parallel PRs around that are missing this, although its little tricky if the test looking for this error is only checking for raise error it may not catch this problem.

app/services/order_shipping_service.rb
ashkan18: yes we haven't had a specific approach here either, if this is something that will be used from other places then maybe moving it to `order` make sense, otherwise if it's only used in `OrderShippingService` we tend to keep it close to the service layer.

app/services/order_shipping_service.rb
```diff
@@ -58,6 +60,13 @@ def validate_artwork!(artwork)
     raise Errors::ValidationError, :missing_artwork_location if artwork[:location].blank?
   end
 
+  def validate_shipping_location!
+    ships_to_continental_us_only = artworks.any? do |(_, artwork)|
```
ashkan18: super #minor but i would rename this from `ships_to_continental_us_only ` to `domestic_shipping_only` or something like that since all its doing is checking if all artworks international shipping flag.

spec/services/order_shipping_service_spec.rb
```diff
@@ -39,10 +59,35 @@
...
+  let(:continental_us_artwork) { gravity_v1_artwork(continental_us_artwork_config) }
 
   before do
     @service_domestic_shipping = OrderShippingService.new(order, fulfillment_type: Order::SHIP, shipping: domestic_shipping)
     @service_international_shipping = OrderShippingService.new(order, fulfillment_type: Order::SHIP, shipping: international_shipping)
+    @service_continental_us_shipping = OrderShippingService.new(continental_us_order, fulfillment_type: Order::SHIP, shipping: non_continental_us_shipping)
+  end
+
+  describe '#process!' do
+    context 'with continental US only shipping' do
```
ashkan18: nice specs 🙌

app/services/order_shipping_service.rb
```diff
@@ -58,6 +60,13 @@ def validate_artwork!(artwork)
...
+      artwork[:international_shipping_fee_cents].nil?
+    end
+    raise Errors::ValidationError, :invalid_shipping_address if ships_to_continental_us_only && !@shipping_address.continental_us?
```
ashkan18: looking more into this, i think we need to check if `artwork`'s location is also in US. With current logic, a work in Berlin that's enabled for domestic but being shipped to HI or AK will get errors, right?

javamonn:
Addressed PR feedback and broadened the check performed in `validate_shipping_location!` to check if a work only ships domestically and expands the definition of `#domestic?` to treat non-continental US addresses as international with respect to the rest of the U.S. I think this makes more sense, rather than trying to explicitly special-case how we treat the continental US in this validation.

app/services/order_shipping_service.rb
```diff
@@ -58,6 +60,14 @@ def validate_artwork!(artwork)
     raise Errors::ValidationError, :missing_artwork_location if artwork[:location].blank?
   end
 
+  def validate_shipping_location!
+    domestic_shipping_only = artworks.any? do |(_, artwork)|
+      artwork[:international_shipping_fee_cents].nil? && !domestic?(artwork[:location])
+    end
+
+    raise Errors::ValidationError.new(:invalid_shipping_address, failure_code: 'domestic_shipping_only') if domestic_shipping_only
```
ashkan18: you could use a symbol for `domestic_shipping_only` should we rename this to `:unsupported_shipping_location`?

app/services/order_shipping_service.rb
```diff
@@ -58,6 +60,14 @@ def validate_artwork!(artwork)
     raise Errors::ValidationError, :missing_artwork_location if artwork[:location].blank?
   end
 
+  def validate_shipping_location!
+    domestic_shipping_only = artworks.any? do |(_, artwork)|
+      artwork[:international_shipping_fee_cents].nil? && !domestic?(artwork[:location])
```
ashkan18: what do you think about reversing the `domestic?` check and have a check for `international_shipping?` just to avoid the negative check. It will basically be
```ruby
artwork[:international_shipping_fee_cents].nil? && international_shipping?(artwork[:location])
```

ashkan18: super #minor

app/services/order_shipping_service.rb
```diff
@@ -58,22 +60,36 @@ def validate_artwork!(artwork)
     raise Errors::ValidationError, :missing_artwork_location if artwork[:location].blank?
   end
 
+  def validate_shipping_location!
+    domestic_shipping_only = artworks.any? do |(_, artwork)|
+      artwork[:international_shipping_fee_cents].nil? && international_shipping?(artwork[:location])
+    end
+
+    raise Errors::ValidationError.new(:unsupported_shipping_location, failure_code: :domestic_shipping_only) if domestic_shipping_only
```
javamonn: It looks like this is being serialized out as the Hash's `to_s` perhaps? I'm seeing this in Reaction currently: 

```
...
data: "{:failure_code=>:domestic_shipping_only}"
```

Which doesn't parse as JSON. I found [this discussion](https://artsy.slack.com/archives/CB110C6LE/p1538144343000200) in slack, perhaps there's something I can override in this specific case to avoid introducing a larger breaking change.

javamonn: We resolved this by adding serializing error data with [`to_json`](https://github.com/artsy/exchange/pull/204/commits/ada2d5cc2bb059e5c676f1c2083bb03f7b1f2e59#diff-0ee4f9f5b82bd09da5190bcca64d4586R10)

ashkan18:
Thanks! this is awesome! a feature + few bug fixes 🙌

PR #205, opened 2018-10-01
Fix empty array EditionSet issue

ashkan18:
# Problem
https://sentry.io/artsynet/exchange-staging/issues/706978737/?query=is:unresolved
With current logic if artwork's `edition_sets` is empty array we try to fetch first edition set and call `[:id]` on it which raises runtime error.

https://github.com/artsy/exchange/pull/198#pullrequestreview-160285287

# Change
use `present?` and do it earlier before calling `find_verify_edition_set`.

### DIFF
--- a/app/services/create_order_service.rb
+++ b/app/services/create_order_service.rb
@@ -46,7 +46,7 @@ class CreateOrderService
     raise Errors::ValidationError.new(:unpublished_artwork, artwork_id: @artwork_id) unless @artwork[:published]
     raise Errors::ValidationError.new(:not_acquireable, artwork_id: @artwork_id) unless @artwork[:acquireable]
 
-    find_verify_edition_set
+    find_verify_edition_set if @edition_set_id.present? || @artwork[:edition_sets].present?
   end
 
   def post_process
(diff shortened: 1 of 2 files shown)


app/services/create_order_service.rb
```diff
@@ -46,7 +46,7 @@ def pre_process!
     raise Errors::ValidationError.new(:unpublished_artwork, artwork_id: @artwork_id) unless @artwork[:published]
     raise Errors::ValidationError.new(:not_acquireable, artwork_id: @artwork_id) unless @artwork[:acquireable]
 
-    find_verify_edition_set
+    find_verify_edition_set if @edition_set_id.present? || @artwork[:edition_sets].present?
```
sweir27: #minor but in case that method gets called from somewhere else, should we put this logic at the top of `find_verify_edition_set`? If you think this is fine then I'm 👍 .

ashkan18: it's private method in this class, we don't call it from anywhere else but def something to be careful about, on the other hand i think new spec would catch wrong calls 🤞

ashkan18: actually, let me just move that inside the method, like how thats more self-contained

sweir27: (did you want to remove it from here then?)

ashkan18:
merging... 🙌

PR #206, opened 2018-10-01
add datadog tracing

saolsen:
This gets datadog tracing set up. Then once that's good I'll submit another PR for statsd/metric tracking.

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -28,6 +28,10 @@ spec:
             value: 'true'
           - name: RAILS_ENV
             value: production
+          - name: DATADOG_TRACE_AGENT_HOSTNAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
           envFrom:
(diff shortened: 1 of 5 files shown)


saolsen:
Yeah probably, I don't know if we've done any graphql tracing yet. Can look into it once the automatic stuff is working.

ashkan18:
that could totally be a follow and as per need. was just curious about it, Exchange GraphQL is probably not that complicated at the moment at least.

hokusai/production.yml
```diff
@@ -28,6 +28,10 @@ spec:
             value: 'true'
           - name: RAILS_ENV
             value: production
+          - name: DATADOG_TRACE_AGENT_HOSTNAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
```
dleve123: @saolsen where does `spec.nodeName` come from? Where is that value defined?

PR #207, opened 2018-10-01
Switch to GUID for all ids

ashkan18:
# Problem
This has come up few times already where we discussed using GUIDs for `Order`s. Current integer ids have few issues:
- it's a sequential number which exposes the size of our database and etc.
- it's easy to predict and you don't want to expose them in urls and etc.

https://github.com/artsy/exchange/pull/200

# Solution
Switch primary keys of all of our models to GUID using Rails 5.1 way of doing this.

# Migration
A graceful migration seemed pretty complicated, we need to add a new column to each model, populate it with some default, go through all existing foreign keys and update them to the new uuid column data, then go remove existing `id` fields from all models and rename `uuid` to `id`.

Since we are not _live_ yet, it seems lot easier to just update existing migrations (which i know is something we normally should not do) and recreate the database on both staging and prod. 

Im voting for clean up instead of more involved migration but let me know if you think otherwise.

@yuki24 please don't merge without coordination with me, im thinking of doing this on some off hours

# Is this a breaking change?! 
No, in GraphQL schema we use `ID` everywhere for these ids which is actually string, so we should be 👍


ashkan18:
@yuki24 can you think of a better way for migration?

_schema.graphql
```diff
@@ -1,438 +0,0 @@
-# An generic error type for mutations
```
williardx: Why did this file get deleted? It seems unrelated to the changes you're making

williardx:
LGTM as well, thanks @ashkan18!

_schema.graphql
ashkan18: yes, wasn't sure what happened there! i'll bring it back!

yuki24:
@ashkan18 I don't mind rebuilding the entire db if there's no order data in production.

ashkan18:
merging this! I'll look into `_schema.graphql` in follow up PR

PR #208, opened 2018-10-01
let me turn on debugging

saolsen:
because it's broke

### DIFF
--- a/config/initializers/datadog.rb
+++ b/config/initializers/datadog.rb
@@ -3,9 +3,10 @@
 Datadog.configure do |c|
   enabled = ENV.fetch('DATADOG_ENABLED', false)
   hostname = ENV.fetch('TRACE_AGENT_HOSTNAME', 'localhost')
+  debug = ENV.fetch('DATADOG_DEBUG', false)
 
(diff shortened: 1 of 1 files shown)


config/initializers/datadog.rb
```diff
@@ -3,9 +3,10 @@
 Datadog.configure do |c|
   enabled = ENV.fetch('DATADOG_ENABLED', false)
   hostname = ENV.fetch('TRACE_AGENT_HOSTNAME', 'localhost')
+  debug = ENV.fetch('DATADOG_DEBUG', false)
```
ashkan18: one thing though! does `debug` flag down in tracer accept string? basically these `ENV.fetch` will return String and not boolean.

orta:
You merge on green'd in a github review (not a comment) you should add support - https://github.com/artsy/peril-settings/issues/81

config/initializers/datadog.rb
saolsen: oh maybe that's the whole problem, what's the move to parse as a bool?

ashkan18: yes your `== 'true'` should work

PR #211, opened 2018-10-02
Remove all states from list of states where we need to remit taxes

williardx:
### Problem
Legal has determined that we should not be remitting taxes sellers to any states right now. This may change in the future as legislation evolves.

### Solution
Remove all states from `SalesTaxService::REMITTING_STATES` but keep code that allows us to remit sales tax to states.

### What Changed
 - Removed WA and PA from `SalesTaxService::REMITTING_STATES`
- Updated specs

### DIFF
--- a/spec/jobs/record_sales_tax_job_spec.rb
+++ b/spec/jobs/record_sales_tax_job_spec.rb
@@ -2,13 +2,13 @@ require 'rails_helper'
 require 'support/taxjar_helper'
 
 describe RecordSalesTaxJob, type: :job do
-  let(:order) { Fabricate(:order, fulfillment_type: Order::PICKUP, shipping_country: 'US', shipping_postal_code: '10013', shipping_region: 'WA', shipping_city: 'New York', shipping_address_line1: '401 Broadway', shipping_total_cents: 100, state: Order::APPROVED) }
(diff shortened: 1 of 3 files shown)


spec/jobs/record_sales_tax_job_spec.rb
```diff
@@ -2,20 +2,23 @@
...
+      city: 'New York',
+      state: 'NY',
       address: '22 Fake St',
       postal_code: 10013
     }
   end
   describe '#perform' do
     before do
       stub_tax_for_order
+      silence_warnings do
```
ashkan18: oh wow! didn't know about this MAGIC! 🙌

ashkan18:
do we need to update anything https://www.notion.so/artsy/403bca0f9f954019bc9752bb7c7d364f?v=e8e5d991dee440f89ed21cea539c37e8 ?

williardx:
What do you mean by the rules? Do you mean which states adhere to the rules, what the rules are or both?

williardx:
And yes, I need to update some stuff in Notion - thanks for the reminder!

ashkan18:
awesome! I just meant to keep track of tax rule changes that we had done in exchange purely for our history :)

PR #212, opened 2018-10-02
Bring back _schema.graphql …

ashkan18:
it was accidentally deleted in  https://github.com/artsy/exchange/pull/207


ashkan18:
self merge on green

PR #213, opened 2018-10-03
fix uncaptured nil error

ashkan18:
Resolves https://sentry.io/artsynet/exchange-staging/issues/678478367
and
https://sentry.io/artsynet/exchange-staging/issues/711198622


ashkan18:
cc: @williardx moved one of tax jar error codes from validation to processing, let me know if you had any concerns.

PR #215, opened 2018-10-03
Add ability to refund transactions from TaxJar

williardx:
### Problem
When we refund a captured charge we need to send a reversed transaction to TaxJar to record the refund if we previously collected sales tax and planned to remit it.

### Solution
Add capability to create refund transactions in TaxJar. This will eventually be implemented in an async job in `OrderCancellationService`, but since we're not currently posting transactions to TaxJar I'm leaving it out for now. 

### What Changed
- Adds methods to handle creating refunds in TaxJar
- Refactors shared parameters for TaxJar API calls into a single method `construct_tax_params`
- Adds tests to ensure that we properly handle errors from TaxJar

### DIFF
--- a/app/services/sales_tax_service.rb
+++ b/app/services/sales_tax_service.rb
@@ -36,7 +36,7 @@ class SalesTaxService
   end
 
   def refund_transaction(refund_date)
-    @transaction = get_transaction(transaction_id) if @transaction.blank?
+    @transaction = get_transaction(transaction_id)
     @refund = post_refund(refund_date) if @transaction.present?
   rescue Taxjar::Error => e
     raise Errors::ProcessingError.new(:tax_refund_failure, message: e.message)
@@ -74,7 +74,7 @@ class SalesTaxService
   def post_refund(refund_date)
     @tax_client.create_refund(
       construct_tax_params(
-        transaction_id: "#{transaction_id}_refund",
+        transaction_id: "refund_#{transaction_id}",
         transaction_date: refund_date.iso8601,
         transaction_reference_id: transaction_id,
         sales_tax: UnitConverter.convert_cents_to_dollars(@line_item.sales_tax_cents)
@@ -112,6 +112,6 @@ class SalesTaxService
   end
 
   def transaction_id
-    "#{@line_item.order.code}-#{@line_item.id[0..4]}"
+    "#{@line_item.order.id}__#{@line_item.id}"
   end
 end
--- a/app/services/sales_tax_service.rb
+++ b/app/services/sales_tax_service.rb
@@ -20,6 +20,7 @@ class SalesTaxService
     @shipping_address = shipping_address
     @shipping_total_cents = artsy_should_remit_taxes? ? shipping_total_cents : 0
     @transaction = nil
+    @refund = nil
   end
 
   def sales_tax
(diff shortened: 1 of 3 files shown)


app/services/sales_tax_service.rb
```diff
@@ -91,4 +110,8 @@ def destination_address
   def seller_address
     @seller_address ||= GravityService.fetch_partner_location(@line_item.order.seller_id)
   end
+
+  def transaction_id
+    "#{@line_item.order.code}-#{@line_item.id[0..4]}"
```
javamonn: It looks like we enforce uniqueness on `@line_item.order.code`, but curious as to why this changed from from `@line_item.order_id` as I would assume they're also unique?

app/services/sales_tax_service.rb
```diff
@@ -42,28 +50,40 @@ def artsy_should_remit_taxes?
...
-      from_city: origin_address.city,
-      from_street: origin_address.street_line1,
-      to_country: destination_address.country,
-      to_zip: destination_address.postal_code,
-      to_state: destination_address.region,
-      to_city: destination_address.city,
-      to_street: destination_address.street_line1,
-      shipping: UnitConverter.convert_cents_to_dollars(@shipping_total_cents)
-    )
+    @tax_client.tax_for_order(construct_tax_params)
```
javamonn: Nice 👍

app/services/sales_tax_service.rb
ashkan18: is there a limit on the size of `transaction_id` if not, maybe we can use just `@line_item.id`, with that we should be able to find the order if needed. One issue that happens with this formatted ids on third party systems is, if we ever change the format, it will be hard to later map old data and new data.

app/services/sales_tax_service.rb
```diff
@@ -34,6 +35,13 @@ def record_tax_collected
     raise Errors::ProcessingError.new(:tax_recording_failure, message: e.message)
   end
 
+  def refund_transaction(refund_date)
+    @transaction = get_transaction(transaction_id) if @transaction.blank?
```
ashkan18: should we always `get_transaction` even if it's there already? just to make sure we always get the latest one before calling refund on it.

app/services/sales_tax_service.rb
```diff
@@ -34,6 +35,13 @@ def record_tax_collected
...
+    @refund = post_refund(refund_date) if @transaction.present?
```
ashkan18: should we `raise` a `ProcessingError` if `@transaction` was not present? that way once we move this job to sidekiq it will get automatic retries. Thinking of the case where Taxjar was down or something

app/services/sales_tax_service.rb
williardx: There's no limit that I'm aware of -- I mostly did this for human readability since I can imagine the only people who will look at the `transaction_id` will probably be Artsy employees. I'm not opposed to to using the GUIDs, though, and I can have a way to visually break them up so that they're more distinct. I'll change that.

app/services/sales_tax_service.rb
williardx: If TaxJar is down then we'll get a different error than `Taxjar::Error::NotFound`. It's not an error state if `@transaction` is missing since a given refund may or may not include a transaction. We could infer if a refund should include a transaction from `line_item.should_remit_sales_tax` or from `artsy_should_remit_tax?` but I think it's safer to just always try so that we don't inadvertently leave something out in case the logic changes later on.

PR #216, opened 2018-10-03
Add commission_fee_cents to LineItem

ashkan18:
# Problem
When creating `purchase` for each approved order in Gravity, we want to know the commission fee for each artwork sold. Currently `commission_fee_cents` is only stored on `Order`, this makes it hard for us to calculate the commission per artwork.

In general, it seems like a good idea to store `commission_fee_cents` on `LineItem` and also store the total on `Order` so we can go back and track what was the commission at the time of that order.

# Solution
- Added `commission_fee_cents` to `LineItem`
- Updated `OrderTotalUpdater` to first set `commission_fee_cents` on `LineItem` and then set the total on `Order` based on `line_items`.
  - For ☝️ , ended up locking the order to make sure totals are all set together now that updating line items is also involved.
- Updated `OrderEvent` to include `commission_fee_cents` in the event posted
- Updated Schema to include `commission_fee_cents` as `seller_only`, `null: true` field on `LineItem`.
  - For ☝️ to work, had to update `seller_only` definition to understand how to handle `LineItem` fields

# Is this a breaking change?
No, we added a field to `LineItem`, if new field is not used by clients, they don't have to update schema.

### DIFF
--- a/app/services/order_total_updater_service.rb
+++ b/app/services/order_total_updater_service.rb
@@ -9,22 +9,29 @@ class OrderTotalUpdaterService
   def update_totals!
     raise Errors::ValidationError.new('Missing price info on line items', 'd2e1cc') if @order.line_items.any? { |li| li.price_cents.nil? }
 
-    @order.items_total_cents = @order.line_items.map(&:total_amount_cents).sum
-    @order.buyer_total_cents = @order.items_total_cents + @order.shipping_total_cents.to_i + @order.tax_total_cents.to_i
-    @order.commission_fee_cents = calculate_commission_fee if @commission_rate.present?
-    @order.transaction_fee_cents = calculate_transaction_fee
-    @order.seller_total_cents = @order.buyer_total_cents - @order.commission_fee_cents.to_i - @order.transaction_fee_cents.to_i - calculate_remittable_sales_tax
-    @order.save!
+    @order.with_lock do
+      @order.items_total_cents = @order.line_items.map(&:total_amount_cents).sum
+      @order.buyer_total_cents = @order.items_total_cents + @order.shipping_total_cents.to_i + @order.tax_total_cents.to_i
+      if @commission_rate.present?
+        set_commission_on_line_items
+        @order.commission_fee_cents = @order.line_items.map(&:commission_fee_cents).sum
+      end
+      @order.transaction_fee_cents = calculate_transaction_fee
+      @order.seller_total_cents = @order.buyer_total_cents - @order.commission_fee_cents.to_i - @order.transaction_fee_cents.to_i - calculate_remittable_sales_tax
+      @order.save!
+    end
   end
 
   private
 
-  def calculate_remittable_sales_tax
-    @order.line_items.select(&:should_remit_sales_tax).sum(&:sales_tax_cents)
+  def set_commission_on_line_items
+    @order.line_items.each do |li|
+      li.update!(commission_fee_cents: li.total_amount_cents * @commission_rate)
+    end
   end
 
-  def calculate_commission_fee
-    @order.items_total_cents * @commission_rate
+  def calculate_remittable_sales_tax
+    @order.line_items.select(&:should_remit_sales_tax).sum(&:sales_tax_cents)
   end
 
   def calculate_transaction_fee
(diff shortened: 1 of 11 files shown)


app/services/order_total_updater_service.rb
```diff
@@ -9,22 +9,29 @@ def initialize(order, commission_rate = nil)
...
-    @order.commission_fee_cents = calculate_commission_fee if @commission_rate.present?
-    @order.transaction_fee_cents = calculate_transaction_fee
-    @order.seller_total_cents = @order.buyer_total_cents - @order.commission_fee_cents.to_i - @order.transaction_fee_cents.to_i - calculate_remittable_sales_tax
-    @order.save!
+    @order.with_lock do
+      @order.items_total_cents = @order.line_items.map(&:total_amount_cents).sum
+      @order.buyer_total_cents = @order.items_total_cents + @order.shipping_total_cents.to_i + @order.tax_total_cents.to_i
+      if @commission_rate.present?
+        set_commission_on_line_items
+        @order.commission_fee_cents = @order.line_items.map(&:commission_fee_cents).sum
```
joeyAghion: Maybe the `commission_fee_cents` should be set [to `0`] whether or not a rate is present?

ashkan18: i kind of wanted to have it this way to distinguish between a calculated commission fee compared to a not yet calculated case. so `nil` being we haven't yet calculated commission, `0` being we calculated but ended up being `0`.

ashkan18:
we pretty much call this service after each buyer side event, `create`, `setShipping`, `submit`

app/services/order_total_updater_service.rb
```diff
@@ -9,22 +9,29 @@ def initialize(order, commission_rate = nil)
...
 
-    @order.items_total_cents = @order.line_items.map(&:total_amount_cents).sum
-    @order.buyer_total_cents = @order.items_total_cents + @order.shipping_total_cents.to_i + @order.tax_total_cents.to_i
-    @order.commission_fee_cents = calculate_commission_fee if @commission_rate.present?
-    @order.transaction_fee_cents = calculate_transaction_fee
-    @order.seller_total_cents = @order.buyer_total_cents - @order.commission_fee_cents.to_i - @order.transaction_fee_cents.to_i - calculate_remittable_sales_tax
-    @order.save!
+    @order.with_lock do
+      @order.items_total_cents = @order.line_items.map(&:total_amount_cents).sum
+      @order.buyer_total_cents = @order.items_total_cents + @order.shipping_total_cents.to_i + @order.tax_total_cents.to_i
```
williardx: I didn't think about this before, but why do we need to convert `shipping_total_cents` and `tax_total_cents` to `int`? Isn't it already an integer?

spec/services/create_order_service_spec.rb
```diff
@@ -29,11 +29,11 @@
           end.to change(Order, :count).by(1).and change(LineItem, :count).by(1)
         end
         it 'sets state_expires_at for newly pending order' do
-          Timecop.freeze(Time.now.utc) do
+          Timecop.freeze(Time.now.beginning_of_day) do
```
williardx: How come you switched to `beginning_of_day`?

williardx:
💯 Feel free to merge when ready, thanks @ashkan18!

spec/services/create_order_service_spec.rb
ashkan18: this was a weird circleci issue with `timecop`

app/services/order_total_updater_service.rb
ashkan18: yeah good point, i don't remember why we added it either. going to remove

ashkan18: turns out they were there for a reason hehehe, `to_i` makes `nil` a `0`. Without `to_i` we either have to set default for all of these to `0` or check for `nil` for each of these.
The reason i wont be up for setting default to `0` is, i like to differentiate between the case we haven't set these fields vs. the case we've set them but ended up being `0`.

williardx: Ah! That makes sense 👍

PR #219, opened 2018-10-08
Support Refunds by Admins

ashkan18:
# Problem
We want admins to be able to refund an order.

# Solution
Added a new _refund_ `action_item` to Order show page. This button only shows up for `approved` and `fulfilled` orders.

Also added a  new `REFUNDED` status to `Order` and used it from newly added `OrderCancellationService.refund!` method.

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -34,6 +34,55 @@ ActiveAdmin.register Order do
     end
   end
 
+  member_action :refund, method: :post do
+    OrderCancellationService.new(resource).refund!
+    redirect_to resource_path, notice: "Refunded!"
+  end
+
+  action_item :refund, only: :show do
+    link_to 'Refund', refund_admin_order_path(order), method: :post if [Order::APPROVED, Order::FULFILLED].include? order.state
+  end
+
+  sidebar :contact_info, only: :show do
+    #TODO: why doesn't this work?
+    link_to "artsy.net" do
+      button "View Artwork on Artsy"
+    end
+    panel "Buyer Information" do
+      attributes_table_for order do
+        #TODO: fill this in
+        row :name
+        row :shipping_address do
+          if order.shipping_info?
+            #TODO: shipping info
+            #table_for order.admin_notes
+          else
+            'None'
+          end
+        end
+        row :shipping_phone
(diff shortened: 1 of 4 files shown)


app/services/order_cancellation_service.rb
```diff
@@ -23,6 +23,15 @@ def reject!
     @order.transactions << @transaction if @transaction.present?
   end
 
+  def refund!
```
williardx: A different name here would be good since this name is too similar to the private `refund` method

app/admin/orders.rb
```diff
@@ -34,6 +34,55 @@
...
+  end
+
+  sidebar :contact_info, only: :show do
+    #TODO: why doesn't this work?
+    link_to "artsy.net" do
+      button "View Artwork on Artsy"
+    end
+    panel "Buyer Information" do
+      attributes_table_for order do
+        #TODO: fill this in
```
williardx: Could you make this comment more specific, or maybe link to a spec?

ashkan18:
note to self: set `reverse_transfer` when calling stripe

app/admin/orders.rb
```diff
@@ -34,6 +34,55 @@
...
+        row :name
+        row :shipping_address do
+          if order.shipping_info?
+            #TODO: shipping info
+            #table_for order.admin_notes
```
williardx: Is this supposed to be commented out?

app/admin/orders.rb
```diff
@@ -65,7 +114,7 @@
     panel "Transaction" do
       #TODO: finish this payment summary
```
williardx: What else needs to be done?

app/services/payment_service.rb
```diff
@@ -27,7 +27,7 @@ def self.capture_charge(charge_id)
   end
 
   def self.refund_charge(charge_id)
-    refund = Stripe::Refund.create(charge: charge_id)
+    refund = Stripe::Refund.create(charge: charge_id, reverse_transfer: true)
```
williardx: Nice catch 😅

williardx:
Looks good, thanks @ashkan18!

PR #220, opened 2018-10-09
More generic `seller_type`

ashkan18:
# Problem
For Auction Buy Now works, we want to use a different `type` for Stripe's `metadata` when creating charges.

https://artsyproduct.atlassian.net/browse/PURCHASE-547

# Solution
Currently `seller_type` was hardcoded to be `partner`, we now make sure `seller_type` reflects the `partner[:type]` of the artwork. This way we store the partner type at the time of order on `Order` properly.

We use ☝️ to detect what should we send Stripe for `type` field of charge metadata. 

We also updated `OrderPartyUnion` to work with new cases of `seller_type` field.

# Migration
```ruby
Order.update_all!(seller_type: 'gallery')
```

### DIFF
--- a/app/graphql/types/order_party_union_type.rb
+++ b/app/graphql/types/order_party_union_type.rb
@@ -1,5 +1,6 @@
 class Types::Partner < Types::BaseObject
   field :id, String, null: false
+  field :type, String, null: false
 end
 
 class Types::User < Types::BaseObject
@@ -13,10 +14,8 @@ class Types::OrderPartyUnionType < Types::BaseUnion
     case object.type
     when Order::USER
       Types::User
-    when Order::PARTNER
-      Types::Partner
     else
-      raise "Unexpected Return value: #{object.inspect}"
+      Types::Partner
     end
   end
 end
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -49,10 +49,7 @@ class Order < ApplicationRecord
     PARTNER = 'partner'.freeze
   ].freeze
 
-  SELLER_TYPES = [
-    AUCTION = 'auction'.freeze,
-    GALLERY = 'gallery'.freeze
-  ].freeze
+  AUCTION_SELLER_TYPE = 'auction'.freeze
 
   has_many :line_items, dependent: :destroy, class_name: 'LineItem'
   has_many :transactions, dependent: :destroy
@@ -98,7 +95,7 @@ class Order < ApplicationRecord
   end
 
   def auction_seller?
-    seller_type == AUCTION
+    seller_type == AUCTION_SELLER_TYPE
   end
 
   def to_s
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -49,6 +49,11 @@ class Order < ApplicationRecord
     PARTNER = 'partner'.freeze
   ].freeze
 
+  SELLER_TYPES = [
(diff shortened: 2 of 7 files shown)


app/services/order_submit_service.rb
```diff
@@ -92,7 +92,7 @@ def charge_metadata
       buyer_type: @order.buyer_type,
       seller_id: @order.seller_id,
       seller_type: @order.seller_type,
-      type: 'bn-mo'
+      type: @order.seller_type == 'auction' ? 'auction-bn' : 'bn-mo'
```
sweir27: Would there be any use in pulling this out as a constant on `Order` similar to https://github.com/artsy/exchange/blob/master/app/models/order.rb#L47-L50?

app/graphql/types/order_party_union_type.rb
```diff
@@ -1,5 +1,6 @@
 class Types::Partner < Types::BaseObject
   field :id, String, null: false
+  field :type, String, null: false
```
sweir27: If this is returned in the schema, when it's stitched with gravity will this get precedence or the up-to-date _actual_ partner's type?

If the latter, I don't know if it's necessary to return this at all.

ashkan18: i think from GraphQL perspective, not assuming MP stitched version, returning the type here makes sense. MP will ignore that `type` field in both stitched and non-stitched version but would be useful for us to have it exposed for debugging and etc.

app/services/order_submit_service.rb
ashkan18: i thought about having all partner types here, but looking at https://github.com/artsy/gravity/blob/master/app/models/domain/partner.rb#L14 it seems there is no need at the moment to have them in exchange. Do you think it'll be easier if we make this check a `Order` model thing? something like, `order.auction_seller?`

app/graphql/types/order_party_union_type.rb
sweir27: It does seem nice to return the `type` here for querying from exchange.

I can imagine confusion if a partner changed type and we returned and outdated type from exchange vs. the updated one from metaphysics (due to stitching in the correct version).

app/models/order.rb
```diff
@@ -49,6 +49,11 @@ class Order < ApplicationRecord
     PARTNER = 'partner'.freeze
   ].freeze
 
+  SELLER_TYPES = [
```
sweir27: Seems like we should go all or nothing on these-- as in, include all partner types (since something like an institutional seller is missing here) _or_ just update the `auction_seller?` method to reference the string?

ashkan18: the gallery one is def not used and not sure if we need to add others, maybe i just make auction one a const.

app/graphql/types/order_party_union_type.rb
ashkan18: from MP's side you won't even see this field since the whole thing is going to be replaced with Gravity's partner.

sweir27: Exactly, but if you query exchange and you query metaphysics you'll see different things which is confusing

ashkan18: but that might be ok and expected, when querying Exchange you get `id` and `type` back but when querying MP you get `PartnerType` or `UserType` back

PR #221, opened 2018-10-10
Fix money bug, add links and flesh out Buyer/Seller info, add transaction details and state history

ansor4:
Changes:

- Previously was incorrectly displaying all prices/fees by a factor of 100, `number_to_currency` assumes cents in decimal places
- Add buyer info
- Add seller info
- Add user admin link
- Add partner admin link
- Add order status link
- Add state history
- Add transaction details
- Remove sales contacts info for now


app/admin/orders.rb
```diff
@@ -44,100 +44,125 @@
   end
 
   sidebar :contact_info, only: :show do
-    #TODO: why doesn't this work?
-    link_to "artsy.net" do
-      button "View Artwork on Artsy"
-    end
+
+    #TODO: how do I get artwork_id
```
ashkan18: we can do this in a follow up PR, we basically need to iterate through `line_items` and for each add a link to artwork.

app/admin/orders.rb
```diff
@@ -44,100 +44,125 @@
...
+    h5 link_to("View Artwork on Artsy", artsy_view_artwork_url())
+
     panel "Buyer Information" do
       attributes_table_for order do
-        #TODO: fill this in
-        row :name
+        row :shipping_name
```
ashkan18: we may want to limit this section to `if order.fulfillment_type == Order::SHIP`

app/admin/orders.rb
```diff
@@ -44,100 +44,125 @@
...
           end
         end
-        row :shipping_phone
+        row 'Shipping Phone' do
+          number_to_phone order.buyer_phone_number
+        end
+        #TODO: fill in email
         row :email
       end
+      h5 link_to("View User in Admin", artsy_view_user_admin_url(order.buyer_id))
```
ashkan18: nice! i'll just add a guard checking if `order.buyer_type == 'user'`

app/helpers/url_helper.rb
```diff
@@ -0,0 +1,18 @@
+module UrlHelper
+  def artsy_order_status_url(order_id)
+    "https://www.artsy.net/order/#{order_id}/status"
```
ashkan18: lets make `https://www.artsy.net` a config we set in Env, this way we can have correct link for staging

app/helpers/url_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  # TODO: make this take an artwork id
+  def artsy_view_artwork_url
+    'https://www.artsy.net/'
+  end
+
+  def artsy_view_user_admin_url(user_id)
+    "https://admin.artsy.net/user/#{user_id}"
```
ashkan18: we can make this a config too

app/helpers/url_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def artsy_view_partner_admin_url(partner_id)
+    "https://admin-partners.artsy.net/partners/#{partner_id}"
```
ashkan18: and this

app/helpers/url_helper.rb
sweir27: Where does this show up? afaik admins don't have blanket-access to view orders so I'm not sure this would be helpful.

Also, the current url is `/orders`. We'll eventually switch to `/order` but right now that points to the legacy ecommerce flow. 😄

config/force.yml
```diff
@@ -0,0 +1,6 @@
+development: &default
+  artwork_url: <%= "#{ENV['FORCE_URL']}/artwork" %>
```
ashkan18: we can take `/artwork` out of this and just call it `url` then in helper we can add `/artwork` this way if we ended up adding new pages we don't have to update config. what do you think?

config/torque.yml
```diff
@@ -0,0 +1,6 @@
+development: &default
+  user_url: <%= "#{ENV['TORQUE_URL']}/user" %>
```
ashkan18: same question as above

config/vibrations.yml
```diff
@@ -0,0 +1,6 @@
+development: &default
+  partners_url: <%= "#{ENV['VIBRATIONS_URL']}/partners" %>
```
ashkan18: and here

PR #224, opened 2018-10-10
Add commission_rate and processing_fee_schedule to line item

javamonn:
Adds new fields to line item and exposes formatted display strings via GraphQL as part of [SELL-976](https://artsyproduct.atlassian.net/browse/SELL-976). Initial pass based off of discussions in Slack [here](https://artsy.slack.com/archives/CB110C6LE/p1539027899000100) and [here](https://artsy.slack.com/archives/CB110C6LE/p1539027970000100).

From the discussion [here](https://artsy.slack.com/archives/CB110C6LE/p1539026352000100), the current plan is to display the transaction fee as "3%" even though we calculate the fee using 2.9% + $0.30, and this may change in the future.

### DIFF
--- a/app/services/order_total_updater_service.rb
+++ b/app/services/order_total_updater_service.rb
@@ -14,6 +14,7 @@ class OrderTotalUpdaterService
       @order.buyer_total_cents = @order.items_total_cents + @order.shipping_total_cents.to_i + @order.tax_total_cents.to_i
       if @commission_rate.present?
         set_commission_on_line_items
+        @order.commission_rate = @commission_rate
         @order.commission_fee_cents = @order.line_items.map(&:commission_fee_cents).sum
       end
       @order.transaction_fee_cents = calculate_transaction_fee
--- a/app/graphql/types/order_type.rb
+++ b/app/graphql/types/order_type.rb
@@ -25,6 +25,8 @@ class Types::OrderType < Types::BaseObject
   field :state_expires_at, Types::DateTimeType, null: true
   field :last_submitted_at, Types::DateTimeType, null: true
   field :last_approved_at, Types::DateTimeType, null: true
+  field :commission_rate, Float, null: true
+  field :display_commission_rate, String, null: true
   field :line_items, Types::LineItemType.connection_type, null: true
 
   def buyer
@@ -48,5 +50,15 @@ class Types::OrderType < Types::BaseObject
     return if object.fulfillment_type.blank?
 
     object
+  end
+
+  def display_commission_rate
+    return if object.commission_rate.nil?
+
+    ActiveSupport::NumberHelper.number_to_percentage(
+      object.commission_rate * 100,
+      precision: 2,
+      strip_insignificant_zeros: true
+    )
   end
 end
--- a/db/migrate/20181009163308_add_commission_rate_to_order.rb
+++ b/db/migrate/20181009163308_add_commission_rate_to_order.rb
@@ -1,0 +1,7 @@
+class AddCommissionRateToOrder < ActiveRecord::Migration[5.2]
+  def change
+    change_table :orders, bulk: true do |t|
+      t.column :commission_rate, :float
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_10_03_191806) do
+ActiveRecord::Schema.define(version: 2018_10_09_163308) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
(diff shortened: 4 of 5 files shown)


app/services/order_total_updater_service.rb
```diff
@@ -24,9 +24,13 @@ def update_totals!
 
   private
 
-  def set_commission_on_line_items
+  def set_fees_on_line_items
     @order.line_items.each do |li|
-      li.update!(commission_fee_cents: li.total_amount_cents * @commission_rate)
+      li.update!(
+        commission_fee_cents: li.total_amount_cents * @commission_rate,
+        commission_rate: @commission_rate,
```
javamonn: Should these fields live on Order instead of the Line Item?

app/services/order_total_updater_service.rb
```diff
@@ -24,9 +24,13 @@ def update_totals!
...
+        transaction_fee_schedule: LineItem::DEFAULT
```
javamonn: Is there a better way to namespace constants like this in ruby? Referencing it as `DEFAULT` without qualifying it with `TRANSACTION_FEE_SCHEDULES` feels a little off.

app/graphql/types/line_item_type.rb
```diff
@@ -10,5 +10,22 @@ class Types::LineItemType < Types::BaseObject
...
+  field :transaction_fee_schedule, Types::LineItemTransactionFeeScheduleEnum, null: true
+  field :display_transaction_fee_schedule, String, null: true
+  field :display_commission_rate, String, null: true
   field :fulfillments, Types::FulfillmentType.connection_type, null: true
+
+  def display_transaction_fee_schedule
+    LineItem::DISPLAY_TRANSACTION_FEE_SCHEDULES[object.transaction_fee_schedule] unless object.transaction_fee_schedule.nil?
+  end
+
+  def display_commission_rate
```
javamonn: I originally had these defined on the LineItem model, but they are strictly display formatters / "view" code.

app/services/order_total_updater_service.rb
williardx: What made you choose line item over order? Order seems more natural to me since to my knowledge the commission fee and transaction fee don't change on a line item basis.

app/graphql/types/line_item_type.rb
williardx: We might want these display views (or an additional set of views) for the admin app, in which case they should probably live on the model.

app/graphql/types/line_item_type.rb
```diff
@@ -10,5 +10,22 @@ class Types::LineItemType < Types::BaseObject
   field :commission_fee_cents, Integer, null: true, seller_only: true
   field :created_at, Types::DateTimeType, null: false
   field :updated_at, Types::DateTimeType, null: false
+  field :transaction_fee_schedule, Types::LineItemTransactionFeeScheduleEnum, null: true
+  field :display_transaction_fee_schedule, String, null: true
+  field :display_commission_rate, String, null: true
   field :fulfillments, Types::FulfillmentType.connection_type, null: true
+
+  def display_transaction_fee_schedule
+    LineItem::DISPLAY_TRANSACTION_FEE_SCHEDULES[object.transaction_fee_schedule] unless object.transaction_fee_schedule.nil?
```
joeyAghion: Very minor, but I prefer the positive variant of these conditionals when it's not more complicated (`if object.transaction_fee_schedule` in this case).

app/graphql/types/line_item_transaction_fee_schedule_enum.rb
```diff
@@ -0,0 +1,3 @@
+class Types::LineItemTransactionFeeScheduleEnum < Types::BaseEnum
+  value 'DEFAULT', 'default transaction fee schedule', value: LineItem::DEFAULT
```
joeyAghion: This specifies a capitalized `DEFAULT` value, but the list of constants includes a lowercase `default`. I'm probably just misunderstanding the two "values" in this line. Could they be made consistent?

app/services/order_total_updater_service.rb
joeyAghion: Since commissions vary by partner, fees don't yet vary at all, and `Order`s are already associated with a specific partner, I agree that it's sensible to store these on `Order` instead of `LineItem`. Calculating and storing a single commission total also avoids potential problems of summing rounded values in the eventual case that there are orders with multiple line items.

app/services/order_total_updater_service.rb
joeyAghion: Agree. You could just declare a module (embedded in another class is fine) like:

    module TransactionFeeSchedule
      ALL = [
        DEFAULT = '...'
      ]
    end

Then you'd refer to them as `Order::TransactionFeeSchedule::ALL` or could drop the `Order::...` within that class.

app/models/line_item.rb
```diff
@@ -3,6 +3,14 @@ class LineItem < ApplicationRecord
   has_many :line_item_fulfillments, dependent: :destroy
   has_many :fulfillments, through: :line_item_fulfillments
 
+  TRANSACTION_FEE_SCHEDULES = [
+    DEFAULT = 'default'.freeze
+  ].freeze
+
+  DISPLAY_TRANSACTION_FEE_SCHEDULES = {
+    DEFAULT => '3%'
```
joeyAghion: It's a little odd how this uses the constant declared above as the key in this other constant. Maybe you should just declare a single constant hash, with the key being the string that gets stored and the value reflecting the friendlier string. That would make the usage more clear.

ashkan18:
There is a migration that we could run to set this field for existing orders on staging and prod :)

PR #226, opened 2018-10-12
Db migration to have ability to add admin notes

ansor4:
List of changes:
- Add separate new admin note page
- change some link styles to buttons
- add check for credit card info before call to gravity
- change column name to from keyword type


ashkan18:
Looks great! feel free to merge after `byebug`

PR #228, opened 2018-10-15
Sending message to admins before orders expire

orta:
Relates to [PURCHASE-554](https://artsyproduct.atlassian.net/browse/PURCHASE-554)\


spec/jobs/order_internal_follow_up_job_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+describe OrderInternalFollowUpJob, type: :job do
+  let(:state) { Order::PENDING }
+  let(:order) { Fabricate(:order, state: state) }
+  describe '#perform' do
+    context 'with an order in the same state before its expiration time' do
+      context 'SUBMITTED' do
+        let(:state) { Order::SUBMITTED }
+        it 'sends an internal message to admins about seller lapsing' do
+          Timecop.freeze(order.state_expires_at + 1.second) do
+            expect_any_instance_of(PostNotificationJob).to receive(:perform_later)
```
orta: These tests were failing

orta:
Alright - this should fix it up with all the reviewer feedback

app/services/create_order_service.rb
```diff
@@ -50,7 +50,10 @@ def pre_process!
   end
 
   def post_process
+    # The official timeout
     OrderFollowUpJob.set(wait_until: @order.state_expires_at).perform_later(@order.id, @order.state)
+    # The timeout for our admins to reach out to folks
+    ReminderFollowUpJob.set(wait_until: @order.state_expires_at - 2.hours).perform_later(@order.id, @order.state)
```
ashkan18: I think we need to call this from `OrderSubmitService` this is going to send these for pending orders

ashkan18: and also maybe from `OrderApproveService`.

orta: Great point! Done

orta:
OK - this is back

app/models/order.rb
```diff
@@ -51,6 +51,11 @@ class Order < ApplicationRecord
     PARTNER = 'partner'.freeze
   ].freeze
 
+  REMINDER_EVENT_VERB = {
+    'seller_delay' => 'seller_delay_reminder'.freeze,
```
ashkan18: @orta sorry for another late comment, i would rename: `seller_delay` to `pending_approval` and `buyer_delay` to `pending_fulfillment`

orta:
Cool, done that 👍

PR #229, opened 2018-10-15
Add Orders sorting over state_expires_at

javamonn:
As part of [SELL-1042](https://artsyproduct.atlassian.net/browse/SELL-1042), enable orders to be sorted over `state_expires_at`. This will enable Volt to sort its "Ready to Ship" tab to show the soonest to expire orders to the top.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -195,9 +195,11 @@ type Order {
   buyerTotalCents: Int
   code: String!
   commissionFeeCents: Int
+  commissionRate: Float
   createdAt: DateTime!
   creditCardId: String
   currencyCode: String!
+  displayCommissionRate: String
   id: ID!
   itemsTotalCents: Int!
   lastApprovedAt: DateTime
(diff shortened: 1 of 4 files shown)


_schema.graphql
```diff
@@ -195,9 +195,11 @@ type Order {
   buyerTotalCents: Int
   code: String!
   commissionFeeCents: Int
+  commissionRate: Float
```
javamonn: Should these changes be committed?

ashkan18: i think we missed adding them from https://github.com/artsy/exchange/pull/224/files

app/graphql/types/query_type.rb
```diff
@@ -40,6 +40,10 @@ def orders(params)
       query.order(state_updated_at: :asc)
     when 'STATE_UPDATED_AT_DESC'
       query.order(state_updated_at: :desc)
+    when 'STATE_EXPIRES_AT_ASC'
+      query.order(state_expires_at: :asc)
+    when 'STATE_EXPIRES_AT_DESC'
+      query.order(state_expires_at: :desc)
```
jonallured: I think it's time to extract this map of sort string to order clause field and direction because it's making the `orders` method harder to read. I'd like to suggest something like this:

```ruby
def orders(params)
    validate_orders_request!(params)
    sort = params.delete(:sort)
    order_clause = sort_to_order[sort]
    Order.where(params).order(order_clause)
end

private

def sort_to_order
  {
    'STATE_UPDATED_AT_DESC' => { state_updated_at: :desc },
    # ...
  }
end
```

I think the signal to noise ratio there is much better and worth doing! Def open to naming ideas here, just trying to convey my thought. ❤️

javamonn: Great suggestion! I've seen this pattern a couple times in Volt, it definitely makes things more readable.

ashkan18:
merging 🙌

PR #231, opened 2018-10-16
Make postal code optional

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-564

Postal code is not mandatory for submitting an order.

# Solution
Remove it from the list of mandatory shipping fields and add spec for it.


williardx:
Looks good, thanks @ashkan18!

PR #232, opened 2018-10-16
Stripe Webhooks

ashkan18:
# Problem
When partner's setup their Stripe account, they get full access to Stripe's dashboard. They refund charges from their dashboard (it's really hard to do it from there, but it's possible 🤷‍♂️ )

Stripe can send us webhooks on different events, we want to be sync with Stripe data and know if a charge got refunded on the dashboard.

# Change
Add a new `@POST /api/webhooks/stripe` which will be used to receive Stripe's webhooks. Also added a new `StripeWebhookService` which will receive an event and in case of `charge.refund` it handles mark the charge as refunded on our system and undeduct the inventory.

One thing to note is, we do get these webhooks even if we are the one who initiated the refund, so we need to make sure we ignore these webhooks if order is already refunded.

# Migration
We need to setup Stripe to deliver these events.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,18 @@ Exchange aspires to be responsible for the various types of e-commerce interacti
 * GitHub: https://github.com/artsy/exchange/
 * Point People: [@ashkan18][ashkan18], [@williardx][williardx]
 
+
+## Deployment
+### Staging
+PRs merged to `master` are automatically deployed to staging by CI.
+
+### Production
+Mention in #dev slack channel that you are going to deploy and then run:
+```
+hokusai pipeline promote --git-remote upstream
+```
+Assuming `upstream` points to Artsy's repository.
+
 ## Setup
 
 * Fork the project to your GitHub account
--- a/app/controllers/api/webhooks_controller.rb
+++ b/app/controllers/api/webhooks_controller.rb
@@ -7,7 +7,7 @@ module Api
       error!('Missing required header.', 400) unless sig_header
       begin
         event = Stripe::Webhook.construct_event(request.body.read, signature_header, Rails.application.config_for(:stripe)['webhook_secret'])
-        WebhookService.process_stripe_event(event)
+        StripeWebhookService.new(event).process!
       rescue JSON::ParserError => e
         error!('Invalid payload.', 400)
       rescue Stripe::SignatureVerificationError => e
--- a/app/controllers/api/webhooks_controller.rb
+++ b/app/controllers/api/webhooks_controller.rb
@@ -1,0 +1,25 @@
+module Api
+  class WebhookController < Api::BaseApiController
+    skip_before_action :authenticate_request!
+
+    def stripe
+      signature_header = request.headers['Stripe-Signature']
+      error!('Missing required header.', 400) unless sig_header
+      begin
+        event = Stripe::Webhook.construct_event(request.body.read, signature_header, Rails.application.config_for(:stripe)['webhook_secret'])
+        WebhookService.process_stripe_event(event)
+      rescue JSON::ParserError => e
+        error!('Invalid payload.', 400)
+      rescue Stripe::SignatureVerificationError => e
+        Rails.logger.info("Received request with invalid stripe signature: #{sig_header}")
+        error!('Invalid request header.', 400)
+      end
+    end
+
+    private
+
+    def stripe_params
+      params.require(:id)
+    end
+  end
+end
--- a/app/controllers/api/webhooks_controller.rb
+++ b/app/controllers/api/webhooks_controller.rb
@@ -1,25 +1,19 @@
 module Api
-  class WebhookController < Api::BaseApiController
+  class WebhooksController < Api::BaseApiController
     skip_before_action :authenticate_request!
 
     def stripe
       signature_header = request.headers['Stripe-Signature']
-      error!('Missing required header.', 400) unless sig_header
+      error!('Missing required header.', 400) unless signature_header
       begin
         event = Stripe::Webhook.construct_event(request.body.read, signature_header, Rails.application.config_for(:stripe)['webhook_secret'])
         StripeWebhookService.new(event).process!
       rescue JSON::ParserError => e
-        error!('Invalid payload.', 400)
+        render json: { error: 'Invalid payload' }, status: 400
       rescue Stripe::SignatureVerificationError => e
-        Rails.logger.info("Received request with invalid stripe signature: #{sig_header}")
-        error!('Invalid request header.', 400)
+        Rails.logger.info("Received request with invalid stripe signature: #{signature_header}")
+        render json: { error: 'Invalid request header' }, status: 400
       end
-    end
-
-    private
-
-    def stripe_params
-      params.require(:id)
     end
   end
 end
--- a/app/controllers/api/webhooks_controller.rb
+++ b/app/controllers/api/webhooks_controller.rb
@@ -8,11 +8,11 @@ module Api
       begin
         event = Stripe::Webhook.construct_event(request.body.read, signature_header, Rails.application.config_for(:stripe)['webhook_secret'])
         StripeWebhookService.new(event).process!
-      rescue JSON::ParserError => e
-        render json: { error: 'Invalid payload' }, status: 400
-      rescue Stripe::SignatureVerificationError => e
+      rescue JSON::ParserError
+        render json: { error: 'Invalid payload' }, status: :bad_request
+      rescue Stripe::SignatureVerificationError
         Rails.logger.info("Received request with invalid stripe signature: #{signature_header}")
-        render json: { error: 'Invalid request header' }, status: 400
+        render json: { error: 'Invalid request header' }, status: :bad_request
       end
     end
   end
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -182,6 +182,6 @@ class Order < ApplicationRecord
   end
 
   def complete_shipping_details?
-    [shipping_name, shipping_address_line1, shipping_city, shipping_country, shipping_postal_code, buyer_phone_number].all?(&:present?)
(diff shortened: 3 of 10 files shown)


app/models/order.rb
```diff
@@ -182,6 +182,6 @@ def build_machine
   end
 
   def complete_shipping_details?
-    [shipping_name, shipping_address_line1, shipping_city, shipping_country, shipping_postal_code, buyer_phone_number].all?(&:present?)
```
ashfurrow: Is this meant for this PR?

app/services/stripe_webhook_service.rb
```diff
@@ -0,0 +1,30 @@
...
+      Rails.logger.debug("ignore event #{@event[:id]} with type: #{@event[:type]}")
+    end
+  end
+
+  private
+
+  def process_refund_event
+    order = Order.find_by(external_charge_id: @event.data.object.id)
+    raise Errors::ProcessingError.new(:unknown_event_charge, event_id: @event.id, charge_id: @event.data.object.id) unless order
+    return if order.state == Order::REFUNDED # ignore if it's already refunded
```
ashfurrow: Should we log this? It's truly exceptional, but couldn't hurt.

config/stripe.yml
```diff
@@ -1,5 +1,6 @@
 development: &default
   stripe_api_key: <%= ENV['STRIPE_API_KEY'] || '' %>
+  webhook_secret: <%= ENV['STRIPE_WEBHOOK_SECRET'] || 'https://media.giphy.com/media/NdKVEei95yvIY/giphy.gif' %>
```
ashfurrow: lol

README.md
```diff
@@ -9,6 +9,18 @@ Exchange aspires to be responsible for the various types of e-commerce interacti
...
+## Deployment
+### Staging
+PRs merged to `master` are automatically deployed to staging by CI.
+
+### Production
+Mention in #dev slack channel that you are going to deploy and then run:
+```
+hokusai pipeline promote --git-remote upstream
+```
+Assuming `upstream` points to Artsy's repository.
```
starsirius: Maybe we can automate production deployment via CI and avoid relying on local git setup of the person kicking off deployment!

app/controllers/api/webhooks_controller.rb
```diff
@@ -0,0 +1,19 @@
+module Api
+  class WebhooksController < Api::BaseApiController
+    skip_before_action :authenticate_request!
+
+    def stripe
+      signature_header = request.headers['Stripe-Signature']
+      error!('Missing required header.', 400) unless signature_header
```
starsirius: Is `error!` a Grape thing?

app/controllers/api/webhooks_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      begin
+        event = Stripe::Webhook.construct_event(request.body.read, signature_header, Rails.application.config_for(:stripe)['webhook_secret'])
+        StripeWebhookService.new(event).process!
+      rescue JSON::ParserError
+        render json: { error: 'Invalid payload' }, status: :bad_request
+      rescue Stripe::SignatureVerificationError
+        Rails.logger.info("Received request with invalid stripe signature: #{signature_header}")
```
starsirius: #minor. I don't feel too strongly about log level, but invalid signature can be malicious.

app/services/stripe_webhook_service.rb
```diff
@@ -0,0 +1,30 @@
...
+    raise Errors::ProcessingError.new(:received_partial_refund, event_id: @event.id, charge_id: @event.data.object.id) unless @event.data.object.refunded
```
starsirius: So this means in case of partial refund (likely via Stripe dashboard), we don't update order state or inventory for now, and we only do that when it's fully refunded. This makes sense now and we can follow up with products and determine the desired business requirements. A scenario that _could_ happen now is that a partner partially refunds a charge in their Stripe dashboard (for whatever reason), and since the order/inventory is not updated, they might need to contact us to make the work buy-able again.

starsirius: Also, from reading the Stripe doc about [responding to webhooks](https://stripe.com/docs/webhooks#responding-to-a-webhook), responses with non-`2xx` code will be treated by Stripe as not received and Stripe will retry. Maybe we can consider, instead of raising errors, always returning `2xx` and log the exception cases.

README.md
ashkan18: yes, makes sense, will add an issue for it.

app/controllers/api/webhooks_controller.rb
ashkan18: don't tell anyone, but sounds like copy/paste issue... interesting enough there is a test for it and it works! will update

app/services/stripe_webhook_service.rb
williardx: It feels dangerous to ignore these partial refund events since it means our systems' records won't match reality, but I agree with @starsirius that we can follow up on that and get better requirements first. Would it make sense to just record the partial refund as a `transaction`, that way at least it's somewhere in our database?

app/controllers/api/webhooks_controller.rb
ashkan18: ignore my comment, there are no tests for not having signature, adding one.

spec/controllers/api/webhooks_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      )
+    end
+    let(:random_event_payload) { StripeMock.mock_webhook_payload('plan.updated') }
+    let(:random_event) { StripeMock.mock_webhook_event('plan.updated') }
+    it 'returns 400 for unauthorized requests with wrong signature' do
+      allow(Stripe::Webhook).to receive(:construct_event).and_raise(Stripe::SignatureVerificationError.new('invalid signature', '402'))
+      post '/api/webhooks/stripe', params: charge_refunded_payload, headers: { 'HTTP_STRIPE_SIGNATURE' => 'test_header' }
+      expect(response.status).to eq 400
+    end
+    it 'refunds the order' do
```
williardx: We should have a test for the partial refund event error.

spec/services/stripe_webhook_service_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+          expect(e.type).to eq :processing
+          expect(e.code).to eq :unknown_event_charge
+          expect(e.data[:event_id]).to eq charge_refunded_event.id
+          expect(e.data[:charge_id]).to eq event_charge_id
+        end
+      end
+    end
+    context 'partial refunds' do
+      let(:fully_refunded) { false }
+      it 'raises unknown_event_charge' do
```
williardx: This should be `raises received_partial_refund`

app/models/order.rb
ashkan18: good point! no! and im confused a little since that part happened in another PR and `master` already is updated https://github.com/artsy/exchange/blob/master/app/models/order.rb#L194

app/services/stripe_webhook_service.rb
ashkan18: good point, this happens actually quite a lot, cases where we refund from our UI, we will get one of these events too and we need to ignore it, i wonder if log would confuse things more or not. what do you think?

app/services/stripe_webhook_service.rb
ashkan18: both good points! the tricky part is to ignore the already existing one but that should be find if we do `find_or_create` 

As for 200, thats good point i guess we don't have to return 400 in any of these cases 👍

app/controllers/api/webhooks_controller.rb
```diff
@@ -8,16 +8,16 @@ def stripe
...
     rescue Stripe::SignatureVerificationError
       Rails.logger.warn("Received request with invalid stripe signature: #{@signature_header}")
-      render json: { error: 'Invalid request header' }, status: :bad_request
+      render json: { error: 'Invalid request header' }
     end
 
-
     def require_stripe_header
       @signature_header = request.headers['Stripe-Signature']
-      render json: { error: 'Missing required header.' }, status: :bad_request unless @signature_header
```
starsirius: Sorry for not being clear [here](https://github.com/artsy/exchange/pull/232/files#r225983583). I think we _can_ still return `401` in 3 cases above that failed validation because we can assume they are not from Stripe. Where we need to return `2xx` (but still log exceptions, etc) is down the flow that we think we are responding to Stripe, i.e. [:unkonw_event_charge](https://github.com/artsy/exchange/pull/232/files#diff-801383a2478f1457206c0219408ed34aR210), [:received_partial_refund](https://github.com/artsy/exchange/pull/232/files#diff-801383a2478f1457206c0219408ed34aR23) (they are currently responding [`500`](https://github.com/artsy/exchange/blob/65eea93df60b2adec4e16fb0857648e9ccd7ed07/app/controllers/concerns/error_handler.rb#L7)?). But I could be misunderstanding things here.

app/controllers/api/webhooks_controller.rb
```diff
@@ -8,16 +8,16 @@ def stripe
...
+      render json: { error: 'Missing required header.' } unless @signature_header
```
starsirius: Sorry for not being clear [here](https://github.com/artsy/exchange/pull/232/files#r225983583). I think we _can_ still return `401` in 3 cases above that failed validation because we can assume they are not from Stripe. Where we need to return `2xx` (but still log exceptions, etc) is down the flow that we think we are responding to Stripe, i.e. [:unkonw_event_charge](https://github.com/artsy/exchange/pull/232/files#diff-801383a2478f1457206c0219408ed34aR210), [:received_partial_refund](https://github.com/artsy/exchange/pull/232/files#diff-801383a2478f1457206c0219408ed34aR23) (they are currently responding [`500`](https://github.com/artsy/exchange/blob/65eea93df60b2adec4e16fb0857648e9ccd7ed07/app/controllers/concerns/error_handler.rb#L7)?). But I could be misunderstanding things here.

PR #233, opened 2018-10-16
Post Events on Failed Charges

ashkan18:
# Problem
CRT team wants to get notified every time a charge fails to get processed during BNMO checkout, similar to what they currently have in #fraud

# Solution
Added a new `TransactionEvent`, for now only with `failed` as the only possible action. We call this now from the `ensure` block in submit order service only if transaction failed.

# Follow up
- [ ] Update APRB to listen on these events and show proper messages.
- [ ] This adds new type of events to the same topic for transactions, make sure Pulse does not break and only listens on `order` routing key.

### DIFF
--- a/app/events/transaction_event.rb
+++ b/app/events/transaction_event.rb
@@ -1,0 +1,53 @@
+class TransactionEvent < Events::BaseEvent
+  TOPIC = 'commerce'.freeze
+  ACTIONS = [
+    FAILED = 'failed'.freeze
+  ].freeze
+
+  def self.post(transaction, action, user_id)
+    event = new(user: user_id, action: action, model: transaction)
+    Artsy::EventService.post_event(topic: TOPIC, event: event)
+  end
+
+  def subject
+    {
+      id: @subject
+    }
+  end
+
+  def properties
+    order = @object.order
+    {
+      order: {
+        buyer_id: order.buyer_id,
+        buyer_total_cents: order.buyer_total_cents,
+        buyer_type: order.buyer_type,
+        code: order.code,
+        created_at: order.created_at,
+        currency_code: order.currency_code,
+        fulfillment_type: order.fulfillment_type,
+        items_total_cents: order.items_total_cents,
+        line_items: order.line_items.map { |li| line_item_detail(li) },
+        seller_id: order.seller_id,
+        seller_type: order.seller_type,
+        state: order.state,
+        state_reason: order.state_reason,
+        state_expires_at: order.state_expires_at,
+        updated_at: order.updated_at
+      },
+      failure_code: @object.failure_code,
+      failure_message: @object.failure_message,
+      transaction_type: @object.transaction_type
+    }
+  end
+
+  def line_item_detail(line_item)
+    {
+      price_cents: line_item.price_cents,
+      artwork_id: line_item.artwork_id,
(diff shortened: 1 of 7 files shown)


williardx:
Looks good! Feel free to merge when ready 🥇

app/events/transaction_event.rb
```diff
@@ -0,0 +1,53 @@
+class TransactionEvent < Events::BaseEvent
```
dleve123: It seems like this Event is only fired if the transaction failed -- thoughts on naming `TransactionFailedEvent`? I think such a name would make the purpose of the event a bit more self-evident in the code.

app/services/order_submit_service.rb
```diff
@@ -33,7 +33,10 @@ def process!
     deducted_inventory.each { |li| GravityService.undeduct_inventory(li) }
     raise e
   ensure
-    @order.transactions << @transaction if @transaction.present?
+    if @transaction.present?
+      @order.transactions << @transaction
+      notify_failed_charge if @transaction.failed?
```
dleve123: is there a difference between a charge and a transaction? If not, thoughts on standardizing the language to one or the other?

spec/services/order_submit_service_spec.rb
```diff
@@ -156,6 +156,7 @@
           allow(GravityService).to receive(:get_merchant_account).with(partner_id).and_return(partner_merchant_accounts.first)
           allow(GravityService).to receive(:get_credit_card).with(credit_card_id).and_return(credit_card)
           expect(PostNotificationJob).not_to receive(:perform_later)
+          expect(PostTransactionNotificationJob).to receive(:perform_later)
```
dleve123: as we just discussed IRL, thoughts on:

1. stubbing `PostTransactionNotificationJob` via `allow` generally to ensure that that class isn't actually called (and no network interactions are made)
2. asserting that `PostTransactionNotificationJob` isn't called in at least 1 other context

app/events/transaction_event.rb
ashkan18: so far we've been defining Events mapping to actual models and we used verb to define the actions that can happen on that model, at the moment we only care about failed transactions but we could also post events in case of successful ones later too, we tend to keep the event more generic in case of other possible actions.
as im writing this, i started thinking maybe i should rename the verb to `created` instead of `failed`, failed would be a property on this event. going to update

app/services/order_submit_service.rb
ashkan18: charge is Stripe's object, Transaction is basically the stored version of charge attempts in our end. Transaction could be also related to our attempts for refund and etc too.

spec/services/order_submit_service_spec.rb
ashkan18: makes total sense! updating now 🙌

ashkan18: actually ended up checking for job being enqueued and not being enqueued on these cases, since we use test `ActiveJob` queue here https://github.com/artsy/exchange/blob/master/spec/rails_helper.rb#L20 so we should be good with not calling that method.

ashkan18:
going to merge this and start looking at follow up tasks, this will be pretty hard to test...

PR #234, opened 2018-10-17
Make :not_found return 404 instead of 401

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-539

We decided a while back that users should receive 404 when trying to look at orders that they should not have access to. Users currently receive a 401 and this PR is changing it to 404 as intended. This is technically a breaking change, but I couldn't find any evidence of anyone relying on specific status codes. Certainly not from the reaction/force side. But would still be good to check about volt and etc

integration tested manually against metaphysics 
image

### DIFF
--- a/app/controllers/concerns/error_handler.rb
+++ b/app/controllers/concerns/error_handler.rb
@@ -24,7 +24,8 @@ module ErrorHandler
         else
           status = :unauthorized
         end
-        render json: { errors: [Types::ApplicationErrorType.root_level_from_application_error(exception)] }, status: status
+        http_status = exception.code == :not_found ? :not_found : :unauthorized
+        render json: { errors: [Types::ApplicationErrorType.root_level_from_application_error(exception)] }, status: http_status
       end
 
       rescue_from Errors::ValidationError do |exception|
--- a/app/controllers/concerns/error_handler.rb
+++ b/app/controllers/concerns/error_handler.rb
@@ -19,11 +19,6 @@ module ErrorHandler
 
       rescue_from Errors::AuthError do |exception|
         log_exception(exception)
-        if exception.code == :not_found
-          status = :not_found
-        else
-          status = :unauthorized
-        end
         http_status = exception.code == :not_found ? :not_found : :unauthorized
         render json: { errors: [Types::ApplicationErrorType.root_level_from_application_error(exception)] }, status: http_status
       end
--- a/app/controllers/concerns/error_handler.rb
+++ b/app/controllers/concerns/error_handler.rb
@@ -19,7 +19,12 @@ module ErrorHandler
 
       rescue_from Errors::AuthError do |exception|
         log_exception(exception)
-        render json: { errors: [Types::ApplicationErrorType.root_level_from_application_error(exception)] }, status: :unauthorized
+        if exception.code == :not_found
+          status = :not_found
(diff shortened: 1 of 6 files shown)


williardx:
Looks good, thanks @ds300! I'm also not aware of any hard dependencies on particular status codes, but would definitely be good to do some kind of end-to-end test for this just to make sure. And feel free to ping one of us if you're having trouble getting things set up!

app/controllers/concerns/error_handler.rb
```diff
@@ -19,7 +19,12 @@ def self.included(clazz)
 
       rescue_from Errors::AuthError do |exception|
         log_exception(exception)
-        render json: { errors: [Types::ApplicationErrorType.root_level_from_application_error(exception)] }, status: :unauthorized
+        if exception.code == :not_found
+          status = :not_found
+        else
+          status = :unauthorized
+        end
+        render json: { errors: [Types::ApplicationErrorType.root_level_from_application_error(exception)] }, status: status
```
ashkan18: Nice, #super minor, we could make it something like 👇 too
```suggestion
        http_status = exception.code == :not_found ? :not_found : :unauthorized
        render json: { errors: [Types::ApplicationErrorType.root_level_from_application_error(exception)] }, status: http_status
```

app/controllers/concerns/error_handler.rb
```diff
@@ -19,7 +19,12 @@ def self.included(clazz)
 
       rescue_from Errors::AuthError do |exception|
         log_exception(exception)
-        render json: { errors: [Types::ApplicationErrorType.root_level_from_application_error(exception)] }, status: :unauthorized
+        if exception.code == :not_found
```
starsirius: If we want to 404 when the user is not authorized to access order(s), maybe we can consider just raising `ActiveRecord::RecordNotFound` or something upstream (e.g. [order](https://github.com/artsy/exchange/blob/531b2dddf917e25f3cec71eb346143a653f931cd/app/graphql/types/query_type.rb#L63)) and avoid passing a special `:not_found` status code and checking here? Is there particular error message we want to preserve differently here?

yuki24: @starsirius is right, [`ActiveRecord::RecordNotFound` is mapped to 404](https://github.com/artsy/exchange/blob/6c0ec7220a21866a09ce5d53c3aee6665caeb7d3/app/controllers/concerns/error_handler.rb#L15-L18) and we shouldn't have to do anything to return 404. I guess there's something more going on in the controller.

ds300: Sounds sensible to me. @ashkan18 any opinion on why we'd want to use AuthError instead?

joeyAghion: It looks to me like [all the instances of `AuthError`](https://github.com/artsy/exchange/search?q=autherror&unscoped_q=autherror) are accompanied by the `:not_found` code, so yeah I'd agree that simply raising a `RecordNotFound` is simpler.

ashkan18: Yes makes total sense! Sorry for late response 🙌

ds300:
Alright this should be 🍏

ashkan18:
thanks @ds300 ! lets do this!

PR #235, opened 2018-10-17
Upgrade Ruby and bundle update

jonallured:
This is upgrades Ruby to that latest patch so we get the security fix. While I was at it I also did a bundle update. The only one that stood out to me there was jwt, so I took a look at the CHANGELOG:

https://github.com/jwt/ruby-jwt/blob/master/CHANGELOG.md

And nothing stood out to me, but we can be sure once this hits staging and I smoke test things there.


ashkan18:
oh nice! looks like `oauth2` opened up their JWT requirement 👍  merge on green

jonallured:
Ugh, Ruby hasn't released the new images yet:

https://hub.docker.com/r/library/ruby/tags/

ashkan18:
ooh interesting! Cutting Edge :) Looks like CircleCI is also having issues

jonallured:
Fixed this and pushed a new commit squashing everything down into one. Once GitHub is back to working correctly, I'll make sure a build is triggered and then we can get this sucker merged.

jonallured:
GitHub webhooks finally caught up and this is green so merging! ❤️

PR #236, opened 2018-10-17
Finally add filtering by admin action

ansor4:
This uses a custom `ransacker` function to do something that should be very simple to do otherwise.

One thing to note is that on every refresh of the index page, it will run this query for every row. This is ok b/c activeadmin actually smartly only runs it for the orders that are shown. Since the orders are paginated, that query will at most be run 30 times (the max # of rows per page) per page load. Not great, but not that bad.


app/helpers/url_helper.rb
```diff
@@ -1,7 +1,7 @@
 module UrlHelper
   def artsy_order_status_url(order_id)
```
ashkan18: I think @sweir27 mentioned this at somepoint, this URL is not useful for admins, we may want to just remove it. Seems little scary to have admins accessing the order page where they can checkout

app/models/order.rb
```diff
@@ -130,6 +130,16 @@ def shipping_address
     )
   end
 
+  ransacker :last_admin_note_type, type: :string do
```
ashkan18: wow this is some magical stuff! had no idea we can do these kind of things 👏

app/helpers/url_helper.rb
sweir27: Gonna second this! I'm also not sure if admins are even able to access orders that they haven't created themselves?

ansor4: sounds good, i'll remove this then

ansor4:
Yeah it looks like it'll run that query for every row no matter the filter. I'll give the scopes thing a shot.

ansor4:
Hmm ok after giving it a shot, I'm not sure how this would be done with our current tables. Since this is a has_many relationship, wouldn't we need to at least run the admin_notes query as a subquery or double join to filter on the last admin_note? To me, this equates to equal performance but is actually a bit more complicated than a standalone query.

Any suggestions? Is there a rails way to do this? I kinda want to say let's roll with the ransacker function for now, and if performance ever becomes an issue, let's add a `last_admin_action` column to order and just make sure that value and the admin_notes table are synced.

ashkan18:
@ansor4 lets try https://blog.codeship.com/folding-postgres-window-functions-into-rails/ maybe we could use it to do things by `last_admin` if overhead ended up being too much we can always de-normalize

PR #239, opened 2018-10-19
Set item sales tax to be the collectable state sales tax when available

williardx:
### Problem
The response from TaxJar may include taxes for jurisdictions other than states, but we _only_ want to consider state sales tax. This results in us potentially overcharging on sales tax.

### Solution
If a tax breakdown from TaxJar is available (https://developers.taxjar.com/api/reference/#post-calculate-sales-tax-for-an-order), set the sales tax for that item to be the collectable amount for the state. Otherwise, if it's not available continue to use the overall `amount_to_collect`.

### What Changed
- Uses TaxJar's `breakdown` object to get the state tax collectable (if it's available) and sets sales tax to the state sales tax collectable.
- Adds line item details to the TaxJar parameters, which are required to get the `breakdown` object in the response.

### DIFF
--- a/app/services/sales_tax_service.rb
+++ b/app/services/sales_tax_service.rb
@@ -24,7 +24,11 @@ class SalesTaxService
   end
 
   def sales_tax
-    @sales_tax ||= UnitConverter.convert_dollars_to_cents(fetch_sales_tax.amount_to_collect)
+    @sales_tax ||= begin
+      tax_response = fetch_sales_tax
+      collectable_tax = tax_response.breakdown.present? ? tax_response.breakdown.state_tax_collectable : tax_response.amount_to_collect
+      UnitConverter.convert_dollars_to_cents(collectable_tax)
+    end
   rescue Taxjar::Error => e
     raise Errors::ProcessingError.new(:tax_calculator_failure, message: e.message)
   end
@@ -57,7 +61,14 @@ class SalesTaxService
   end
 
   def fetch_sales_tax
-    @tax_client.tax_for_order(construct_tax_params)
+    @tax_client.tax_for_order(
+      construct_tax_params(
+        line_items: [{
+          unit_price: UnitConverter.convert_cents_to_dollars(@line_item.price_cents),
+          quantity: @line_item.quantity
+        }]
+      )
+    )
   end
 
   def post_transaction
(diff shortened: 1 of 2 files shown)


app/services/sales_tax_service.rb
```diff
@@ -24,7 +24,11 @@ def initialize(
   end
 
   def sales_tax
-    @sales_tax ||= UnitConverter.convert_dollars_to_cents(fetch_sales_tax.amount_to_collect)
+    @sales_tax ||= begin
+      tax_response = fetch_sales_tax
+      collectable_tax = tax_response.breakdown.present? ? tax_response.breakdown.state_tax_collectable : tax_response.amount_to_collect
```
joeyAghion: Could be `tax_response.breakdown&.state_tax_collectable || tax_response.amount_to_collect`

spec/services/sales_tax_service_spec.rb
```diff
@@ -79,9 +79,21 @@
...
       allow(GravityService).to receive(:fetch_partner_location).with(order.seller_id).and_return(partner_address)
     end
-    it 'calls fetch_sales_tax and returns the sales tax in cents' do
-      expect(@service_ship).to receive(:fetch_sales_tax).and_return(double(amount_to_collect: 1.00))
-      expect(@service_ship.sales_tax).to eq 100
+    context 'with a sales tax breakdown' do
+      it 'calls fetch_sales_tax and returns the amount of sales tax to collect on a state level' do
+        breakdown_double = double(state_tax_collectable: 2.00)
+        tax_double = double(amount_to_collect: 3.00, breakdown: breakdown_double)
+        expect(@service_ship).to receive(:fetch_sales_tax).and_return(tax_double)
```
joeyAghion: Very minor and totally optional, but sometimes I find it helpful to declare and name stubs like these at the top of the file. E.g., these could be:

    let(:tax_response) { double(amount_to_collect: 1.00) }
    let(:tax_response_with_breakdown) { double(amount_to_collect: 1.00, breakdown: double(...)) }

Even if each is only used once, it keeps the specs focused and the names can be instructive.

williardx:
I don't have a specific reason for why we can ignore non-state taxes, but my understanding is that we only deal with state taxes. I've got the #DoNotMerge label up just to double check before I merge.

app/services/sales_tax_service.rb
williardx: 👍 👍

spec/services/sales_tax_service_spec.rb
williardx: Good suggestion! I'll change that.

williardx:
Merging after consulting with @joanna20

PR #240, opened 2018-10-22
enable statsd for exchange

saolsen:
Only records 1 metric for now, order create. We can circle back and track lots of stuff once we verify this works.


saolsen:
For specs it's disabled and nothing will happen. I'll add that mismatch case but after this gets merged we should think about what other stuff we want to track and do another followup PR to track those.

ashkan18:
🙌 makes sense

PR #241, opened 2018-10-22
Use all seller sales tax nexuses when calculating sales tax

williardx:
### Problem
Per legal's request, we need to consider every partner location as a sales tax nexus when calculating sales tax. Previously we were either using the artwork location or the partner's primary billing address as the seller's nexus when calculating sales tax.

### Solution
Fetch all partner locations and pass them in as part of the base parameters we send to TaxJar when we make requests, i.e. calculating sales tax.

### What Changed
- Creates `GravityService.fetch_partner_locations` that fetches all locations from a partner. Deleted `GravityService.fetch_partner_location`, which previously served the same purpose when we were only considering one location at a time.
- Passes in all partner locations in call to TaxJar for estimating sales tax.

This PR also takes care of a few other issues opportunistically, namely the fact that we ask for tax estimates on non-US orders even though only US-based addresses are taxable for our purposes. This isn't currently an issue since BN is only launched in the US. 

- Adds `address_taxable?` method to `SalesTaxService` that checks if an address is taxable for our purposes. Currently only US-based addresses are taxable.
- Raises error on initialization of `SalesTaxService` if the seller has no taxable addresses and sets sales tax to 0.

### DIFF
--- a/app/services/order_shipping_service.rb
+++ b/app/services/order_shipping_service.rb
@@ -48,9 +48,17 @@ class OrderShippingService
     @tax_total_cents ||= @order.line_items.map do |li|
       artwork_address = Address.new(artworks[li.artwork_id][:location])
       seller_addresses = GravityService.fetch_partner_locations(@order.seller_id)
-      service = SalesTaxService.new(li, @fulfillment_type, @shipping_address, shipping_total_cents, artwork_address, seller_addresses)
-      li.update!(sales_tax_cents: service.sales_tax, should_remit_sales_tax: service.artsy_should_remit_taxes?)
-      service.sales_tax
+      begin
+        service = SalesTaxService.new(li, @fulfillment_type, @shipping_address, shipping_total_cents, artwork_address, seller_addresses)
+        li.update!(sales_tax_cents: service.sales_tax, should_remit_sales_tax: service.artsy_should_remit_taxes?)
+        service.sales_tax
+      rescue Errors::ValidationError => e
+        raise e unless e.code == :no_taxable_addresses
+
+        # If there are no taxable addresses then we set the sales tax to 0.
+        li.update!(sales_tax_cents: 0, should_remit_sales_tax: false)
+        0
+      end
     end.sum
   rescue Errors::AddressError
     raise Errors::ValidationError, :invalid_artwork_address
--- a/app/services/order_shipping_service.rb
+++ b/app/services/order_shipping_service.rb
@@ -47,7 +47,8 @@ class OrderShippingService
   def tax_total_cents
     @tax_total_cents ||= @order.line_items.map do |li|
       artwork_address = Address.new(artworks[li.artwork_id][:location])
-      service = SalesTaxService.new(li, @fulfillment_type, @shipping_address, shipping_total_cents, artwork_address)
+      seller_addresses = GravityService.fetch_partner_locations(@order.seller_id)
+      service = SalesTaxService.new(li, @fulfillment_type, @shipping_address, shipping_total_cents, artwork_address, seller_addresses)
       li.update!(sales_tax_cents: service.sales_tax, should_remit_sales_tax: service.artsy_should_remit_taxes?)
       service.sales_tax
     end.sum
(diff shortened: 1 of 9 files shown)


app/services/order_shipping_service.rb
```diff
@@ -47,9 +47,18 @@ def artworks
   def tax_total_cents
     @tax_total_cents ||= @order.line_items.map do |li|
       artwork_address = Address.new(artworks[li.artwork_id][:location])
-      service = SalesTaxService.new(li, @fulfillment_type, @shipping_address, shipping_total_cents, artwork_address)
-      li.update!(sales_tax_cents: service.sales_tax, should_remit_sales_tax: service.artsy_should_remit_taxes?)
-      service.sales_tax
+      seller_addresses = GravityService.fetch_partner_locations(@order.seller_id)
```
ashkan18: this can go outside of the loop so we don't call this for each line item since `seller_id` for all of the will be the same.

williardx:
Cool -- I'll make those changes in a followup PR.

app/services/order_shipping_service.rb
williardx: Oh, good catch - I'll move that up.

app/services/sales_tax_service.rb
```diff
@@ -96,30 +107,26 @@ def post_refund(refund_date)
   def construct_tax_params(args = {})
     {
       amount: UnitConverter.convert_cents_to_dollars(@line_item.total_amount_cents),
-      from_country: origin_address.country,
-      from_zip: origin_address.postal_code,
-      from_state: origin_address.region,
-      from_city: origin_address.city,
-      from_street: origin_address.street_line1,
```
starsirius: @williardx Looks like we may want to [always send `from_` address](https://developers.taxjar.com/api/guides/#when-to-use-from_-address-information)? Otherwise, TaxJar will use Artsy's State Settings in TaxJar account.

> When to Use from_ Address Information
>
> In general, we recommend always passing in the address information from which an order ships. This gives us what we need to make sourcing determinations in both origin and destination states (as well as provinces and countries). If you know that you are shipping within/to a destination state, you may elect not to send from_ address information. In addition, in the absence of from_address parameters, we will look to your State Settings in your TaxJar account, using the address information there. However, we recommend a best practice of always sending from_ address information.

app/services/sales_tax_service.rb
```diff
@@ -96,30 +107,26 @@ def post_refund(refund_date)
...
       to_country: destination_address.country,
       to_zip: destination_address.postal_code,
       to_state: destination_address.region,
       to_city: destination_address.city,
       to_street: destination_address.street_line1,
+      nexus_addresses: @seller_nexus_addresses.map do |ad|
```
starsirius: This is probably a question for Legal, but I'm still not sure how accurate it is to use existing partner locations as nexus addresses. The uses of partner locations has been somewhat arbitrary.

app/services/sales_tax_service.rb
williardx: AFAIK we don't store nexus information on our account, though I'll double check that. In the absence of our own nexus information and a `from_` address, TaxJar defaults to using what is the nexus addresses that we pass in. But, given we always know that the origin address is the artwork location there's no reason not to pass it in along with everything else. I'll change that in a followup PR.

app/services/sales_tax_service.rb
williardx: Yeah, I was wondering about that, too. The requirements from legal for these changes are to use all partner locations but we should follow up and make sure that the locations we use are actually valid tax nexuses.

app/services/sales_tax_service.rb
starsirius: Yeah _I think_ we should not use Artsy's nexus information. TaxJar is a black box to me but from the doc it sounds recommended to pass in `from_` address anyway. 👍

app/services/sales_tax_service.rb
starsirius: Just raised this to @joanna20 and we should followup with the Legal team! 👍

sweir27: Do we need to revert in the meantime @starsirius?

joanna20: I have just followed up with legal, and waiting for their input. In the meantime I don't think we need to revert.

PR #243, opened 2018-10-22
[WIP] Offer Order

ashkan18:
# Problem
Add new mutation for creating offer orders.
https://artsyproduct.atlassian.net/browse/PURCHASE-594
Requirement: we want to be able to distinguish between Orders created by offers and buy now orders.

# Changes
- Added `mode` to `Order` model. It's a `not null` field.
- Exposed mode in `OrderType` GraphQL query
- Added optional argument for querying `orders` for filtering by `mode`
- Added new `CreateOfferOrderWithArtwork` mutation, which creates an offer order. The logic is really similar to `CreateOrderWithArtwork` with small differences.
- `CreateOfferOrderService` is slightly different compared to `CreateOrderService`, for _buy_ orders we check is work is `acquireable`, for _offer_ orders we check if work is _offerable_(?). For this to work, ended up making `CreateOrderService` to raise a `NotImplementedError` for `assert_create!` and ended up having two separate services `CreateBuyOrderService` and `CreateOfferOrderService` that implement `assert_create!` and properly `initialize` the service with `mode`. The difference is so minimal you could argue we could have one service with conditions, this makes testing little easier, other than that both approaches are 👍 to me.

# Migration
`mode` should be required. during migration, when adding this new field, we set it as nullable, then we update existing orders to have proper `mode` (all `Order::BUY`) and in another step we add constraint for making it required.

# GraphQL Question
Added a new `OrderModeEnum` for `mode` that can be `BUY` or `OFFER`. @orta @alloy , do you think we need something like
```graphql
... on OfferOrder {
}
... on BuyOrder {
}
```
Is ☝️ the preferred way and something we want/can do?

One BIG issue with ☝️ is i think this is breaking change (unless there is a magical way to do it). Also at the moment the fields in both cases are pretty much the same at the moment. So I think Enum approach might be ok, but wondering what you think.


app/graphql/types/order_mode_enum.rb
```diff
@@ -0,0 +1,5 @@
+class Types::OrderModeEnum < Types::BaseEnum
+  Order::MODES.each do |m|
+    value m.upcase, "#{m.humanize} Order", value: m
+  end
```
alloy: Tiny nit: I think that if you only have 2 values it’s probably more readable if you just write them out.

spec/controllers/api/requests/create_offer_order_with_artwork_mutation_request_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+        quantity: quantity
+      }
+    end
+    let(:mutation) do
+      <<-GRAPHQL
+        mutation($input: CreateOfferOrderWithArtworkInput!) {
+          createOfferOrderWithArtwork(input: $input) {
+            orderOrError {
+              ... on OrderWithMutationSuccess {
+                order {
```
alloy: I understand that this follows the existing mutation, but as shown in [my GraphQL errors talk/blog](http://artsy.github.io/blog/2018/10/19/where-art-thou-my-error/#Make.exceptions.first-class.citizens.of.your.schema) I don’t think we actually need to add this extra level of nesting. It could just be:

```graphql
             orderOrError {
               ... on Order {
                 id
                 buyer {
                   # ...
                 }
                 seller {
                   # ...
                 }
               }
               ... on ApplicationError {
                 code
                 data
                 type
               }
             }
```

I would prefer if we can follow that simple form and deprecate the existing one, so that we get consistent examples to follow.

If I understand it correctly all this work is for new functionality, so would that be feasible to do now?

jonallured:
I'm under the impression that orders made with the Make Offer button will end up having an offer row written to the database, correct? I'm asking because I'm not clear on why we need `mode` when it could be computed - maybe something like this:

```ruby
class Order
  has_many :offers

  def mode
    offers.any? ? MODES::OFFER : MODES::BUY
  end
end
```

So I guess what I'm saying is that rather than writing this string down in the database, I'd like to see us compute it.

---

I might be missing something, but it appears that the concept of an order `mode` exists to differentiate orders. We see this in the way the query might go in the PR description above with the `OrderModeEnum` and the accompanying question re: `on OfferOrder` and `on BuyOrder`, but I don't think this is necessary.

From the CMS POV, I think all I want is to be able to query for the offers on an order. That value along with the state of the order (`offered`, I think?) gives me all I need. I think having `mode` on top of this is overkill. And I think promoting the concept of a `OfferOrder` and a `BuyOrder` is making a distinction that isn't necessary. In my mind they are all orders and some of them just happen to be in the state `offered` with a non-empty array of `offer` records.

sweir27:
@jonallured in the checkout flow, we'll actually create a pending order (with an "offer"  type) in order to show the user the correct first page (and allow them to refresh, go back, etc.). So it's helpful from that perspective to be able to have some signifier, which would be set before any offers are created.

I was also thinking about potential other places we may want to keep track of orders-- maybe after an inquiry or an auction. This is hypothetical but I could imagine the type being helpful in those scenarios as well.

jonallured:
Gotcha, makes sense @sweir27, thanks for the clarification!!

Do you think you'll need to have the concepts of `OfferOrder` and `BuyOrder` or will the string `mode` attribute be enough? It makes sense to me now why mode is important, but I'm still unclear about the value of complicating the queries with this wrinkle.

spec/controllers/api/requests/create_offer_order_with_artwork_mutation_request_spec.rb
ashkan18: yes this makes lot of sense! going to follow this pattern for this new mutation. we may be able to do some magic deprecation stuff for existing mutations but that can come as a separate PR.

app/graphql/mutations/create_offer_order_with_artwork.rb
```diff
@@ -0,0 +1,19 @@
+class Mutations::CreateOfferOrderWithArtwork < Mutations::BaseMutation
+  null true
+
+  argument :artwork_id, String, 'Artwork Id', required: true
+  argument :edition_set_id, String, 'EditionSet Id', required: false
+  argument :quantity, Integer, 'Number of items in the line item', required: false
```
joeyAghion: The description is a good place to document the default.

app/graphql/mutations/create_offer_order_with_artwork.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  field :order_or_error, Mutations::OrderOrErrorUnionType, 'A union of Order/ApplicationError', null: false
+
+  def resolve(artwork_id:, edition_set_id: nil, quantity: 1)
+    service = CreateOfferOrderService.new(user_id: context[:current_user][:id], artwork_id: artwork_id, edition_set_id: edition_set_id, quantity: quantity)
+    service.process!
+    {
+      order_or_error: service.order
+    }
```
joeyAghion: Why the extra whitespace here if not below? (I think both can be 1-line.)

app/graphql/types/order_type.rb
```diff
@@ -3,6 +3,7 @@ class Types::OrderType < Types::BaseObject
   graphql_name 'Order'
 
   field :id, ID, null: false
+  field :mode, Types::OrderModeEnum, null: false
```
joeyAghion: Won't this be broken until existing orders are migrated?

app/graphql/types/order_mode_enum.rb
```diff
@@ -0,0 +1,4 @@
+class Types::OrderModeEnum < Types::BaseEnum
+  value Order::BUY.upcase, 'Buy Order', value: Order::BUY
+  value Order::OFFER.upcase, 'Offer Order', value: Order::OFFER
```
joeyAghion: I find the uppercase and lowercase combinations confusing... any reason they can't all match the existing [lowercase] constants?

db/migrate/20181022140918_add_mode_to_orders.rb
```diff
@@ -0,0 +1,8 @@
+class AddModeToOrders < ActiveRecord::Migration[5.2]
+  def change
+    add_column :orders, :mode, :string, null: true
+    add_index :orders, :mode
```
joeyAghion: An index on a column containing 1 of 2 possible values won't be that helpful. At best it shrinks the number of records that need to be scanned or sorted by half. Do we query by this column today? If so, we should probably add whatever secondary filter/sort fields to this index. If we don't use it today, maybe we should wait on creating the index.

db/migrate/20181022140918_add_mode_to_orders.rb
```diff
@@ -0,0 +1,8 @@
...
+    Order.update_all(mode: Order::BUY)
```
joeyAghion: This migration isn't reversible. You should either declare it as separate `up` and `down` implementations, or use a reversible block:

https://edgeguides.rubyonrails.org/active_record_migrations.html#migration-overview

app/graphql/types/order_mode_enum.rb
ashkan18: i think it's recommended to be uppercase since all examples i see are uppercase http://graphql-ruby.org/type_definitions/enums.html and https://graphql.org/learn/schema/#enumeration-types

db/migrate/20181022140918_add_mode_to_orders.rb
ashkan18: good point, we did add it as an argument to `orders` so it's possible (and i think from CMS we will) filter orders by it. not sure about the secondary filter though.

db/migrate/20181022140918_add_mode_to_orders.rb
ashkan18: makes sense, updated.

app/graphql/types/order_type.rb
sweir27: @ashkan18 I think adding the `null: false` in a separate PR would be cleanest. We don't really lose anything anyways since nothing depends on this being present right now.

PR #244, opened 2018-10-23
Fix RuboCop

jonallured:
I'm not sure how this happened, but I got a green build on #235 so I merged it and then it failed on master because of a couple RuboCop things. This PR fixes those violations.

@ashkan18 please take a look at the changes I made to the creation of that statsd thingy - RuboCop was complaining about the formatting there, so I just broke off the building into a separate method - how does that look to you?


ashkan18:
thanks for quick fix!

PR #245, opened 2018-10-23
Track money

saolsen:
Track money for dashboard.

### DIFF
--- a/app/services/order_approve_service.rb
+++ b/app/services/order_approve_service.rb
@@ -19,8 +19,6 @@ class OrderApproveService
   private
 
   def post_process
-<<<<<<< HEAD
-    # order total cents and commission cents
     record_stats
     @order.line_items.each { |li| RecordSalesTaxJob.perform_later(li.id) }
     PostNotificationJob.perform_later(@order.id, Order::APPROVED, @by)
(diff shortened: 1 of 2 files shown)


saolsen:
Once we push this I'll add it to the dashboard

app/services/order_approve_service.rb
```diff
@@ -19,6 +19,10 @@ def process!
   private
 
   def post_process
+    # order total cents and commission cents
```
saolsen:
```suggestion
```

app/services/order_approve_service.rb
```diff
@@ -19,6 +19,9 @@ def process!
   private
 
   def post_process
+    Exchange.dogstatsd.increment 'order.approve'
```
ashkan18: what do you think about adding all these 3 calls into one `post_tracking_data` or something with less scary name than what i suggested and call that private method from here? 

Also is there a chance these thing would fail? if yes, maybe we want to queue a job for these, basically trying to avoid failing order approval because of possible statsd issues.

saolsen: It can't fail. It just sends a message to localhost over UDP and doesn't care if it got there. Statsd was designed to work like that so you can emit events inline wherever you want and not have to worry about slowing anything down or throwing exceptions.

PR #247, opened 2018-10-23
Call the appropriate GravityService method to get partner location info

pepopowitz:
A method was renamed; we had one call that didn't get switched over to the new name. For now, we're just showing the first partner location returned by Gravity.

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -113,10 +113,12 @@ ActiveAdmin.register Order do
 
       partner_info = GravityService.fetch_partner(order.seller_id)
       if partner_info.present?
-        partner_location = GravityService.fetch_partner_location(order.seller_id)
+        partner_locations = GravityService.fetch_partner_locations(order.seller_id)
 
-        if partner_location.present?
+        if partner_locations.present?
 
+          #TODO - handle multiple partner_locations properly, instead of just taking the first.
+          partner_location = partner_locations.first
           partner_info[:partner_location] = partner_location
 
           attributes_table_for partner_info do


dleve123:
This looks good to me. Notably, there’s follow up to create integration tests to avoid these types of errors in the future. For now, I’d recommend shipping to fix this error.

Also, assuming that Sentry is connected to Exchange, we can use Sentry to see this error, and ensure that it doesn’t re-occur

ashkan18:
Nice catch! 🙌 

Yes, integration tests for active admin would be really nice to add! specially when it comes to actions like refund and etc. We've never done this before for any of our apps using `activeadmin`. In the end `activeadmin` ends up generating normal rails controller's underneath so we should be able to write some integration tests there.

Sentry is setup on Exchange, they should show up [here](https://sentry.io/artsynet/exchange-staging/)

app/admin/orders.rb
```diff
@@ -113,10 +113,12 @@
...
       partner_info = GravityService.fetch_partner(order.seller_id)
       if partner_info.present?
-        partner_location = GravityService.fetch_partner_location(order.seller_id)
+        partner_locations = GravityService.fetch_partner_locations(order.seller_id)
 
-        if partner_location.present?
+        if partner_locations.present?
 
+          #TODO - handle multiple partner_locations properly, instead of just taking the first.
+          partner_location = partner_locations.first
```
oxaudo: Depending on what this location is used for - I think now artwork that is available for BuyNow has a location as well, that is most likely the partner location that we are interested in, no?

ashkan18:
going to merge to fix currently broken details page and we can follow up on showing all partner's locations and artwork location.

PR #248, opened 2018-10-24
add fulfillment info, need tests later

ansor4:
Co-authored-by: Daniel Levenson <dleve123@gmail.com>

We plan on adding feature tests as a part of https://artsyproduct.atlassian.net/browse/PLATFORM-1021 so let's just do the test work there


dleve123:
This looks good to me, but I'd recommend that @ashkan18 or @williardx look that the `fulfillments` method to confirm that we're understanding the database tables correctly. gonna assign them now as reviewers if that's cool!

app/admin/orders.rb
```diff
@@ -155,7 +155,15 @@
           link_to "#{order.id}", artsy_order_status_url(order.id)
         end
         row 'Shipment' do |order|
-          #TODO: shipment info
+          table_for order.fulfillments do
```
ashkan18: i think we should should only show this row if `order.state == Order::FULFILLED` and `order.fulfillment_type == Order::PICKUP`

app/models/order.rb
```diff
@@ -134,6 +134,11 @@ def last_admin_note
     admin_notes.order(:created_at).last
   end
 
+  def fulfillments
+    # TODO: add tests
+    Fulfillment.joins(line_items: :order).where(orders: { id: id }).distinct
```
ashkan18: we can do this also by 
```ruby
order.line_items.map(&:fulfillments).flatten
```

app/admin/orders.rb
ashkan18: i'd suggest we track `fulfillments` per `LineItem` so maybe do a 
```ruby
order.line_items.each do |li|
  table_for li.fulfillments do
   ...
  end
end
```
this way we can also remove `fulfillments` method added to `order.rb` below.

ashkan18:
oh you guys should use `git-duet` when pairing 
https://github.com/git-duet/git-duet

app/admin/orders.rb
ansor4: For some reason when I try to do it this way, I get an extra artifact. Any ideas why? See code below:

```ruby
order.line_items.each do |li|
  table_for li.fulfillments do
    column '' do |fulfillment|
      attributes_table_for fulfillment do
        row :courier
        row :tracking_id
        row :estimated_delivery
      end
    end
  end
end
```

screen shot 2018-10-25 at 2 56 06 pm

ansor4: This doesn't happen when I do something like
```
fulfillments = order.line_items.map(&:fulfillments).flatten
table_for fulfillments do
  ...
end
```
Might just be an activeadmin thing

ashkan18:
Nice! Manually fixed rubocop conflict, merge on green 🙌

PR #250, opened 2018-10-24
PLATFORM-977 - add confirmation dialog to “Refund” button in admin

pepopowitz:
Adds a confirmation dialog to the "Refund" button in the admin, so people don't accidentally refund orders.

Note: This PR does not include tests, because I don't really know how to add them for the admin. If anyone reviewing feels like they can help me with that, that would be grand. If not, we have another story for adding admin tests later (https://artsyproduct.atlassian.net/browse/PLATFORM-1021).

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -44,7 +44,7 @@ ActiveAdmin.register Order do
   end
 
   action_item :refund, only: :show do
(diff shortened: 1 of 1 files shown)


ashkan18:
awesome!! Thanks! 🙌 going to merge

PR #251, opened 2018-10-24
Add skip address validation option, credit card rescue to Exchange

ansor4:
A couple changes here:

- Add rescue to gravity call for credit card info. I don't think we want the possibility of this call to ever not be validated, so we add a rescue only on the admin view.
- Add option to override address validation. This will only be triggered on the admin view as well
- Add tests for overriding address validation.

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -189,8 +189,14 @@ ActiveAdmin.register Order do
     panel "Transaction" do
 
       if order.credit_card_id.present?
-        credit_card_info = GravityService.get_credit_card(order.credit_card_id)
-        if credit_card_info.present?
+        no_credit_card_found = false
+        begin
+          credit_card_info = GravityService.get_credit_card(order.credit_card_id)
+          no_credit_card_found = !credit_card_info.present?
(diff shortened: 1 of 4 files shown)


pepopowitz:
The credit card logic, I understand. 👍 on that part.

The skipping of address validation, I don't know much about. I'd like to await @ashkan18 's review before merging, unless this is urgent.

williardx:
@ansor4 Can you give some more context for this? Why do you want to skip address validation?

ashkan18:
Sorry `skip_valiadation` was my suggestion. The issue there is, when we fetch partner's location currently we automatically validate all of their addresses, in this case in our admin view we just want to show them and an invalid address is not as critical as the cases where address is invalid during processing the order.

That said, the more i think about it, it might actually be better to follow the same pattern here as we did for credit card. I'd suggest lets `rescue Errors::ValidationError` when fetching partner's location and if they had invalid addresses, show a big red error so admins can take action on it.

what do you think?

ashkan18:
thanks @ansor4 , sorry for original bad suggestion :)

PR #252, opened 2018-10-24
Add scope for filtering orders by last_admin_note

ashkan18:
# Problem
We want to be able to filter orders by last admin note type.

# (Possible) Solution
Added `by_last_admin_note` scope to `Order`. We can now do things like 
```ruby
Order.by_last_admin_note('case_opened_cancellation')
```

# Possible other options
We could de-normalize `last_admin_note` but maybe we can wait till there is a need for it

# Possible improvement
Is there a ActiveRecord way for the subquery we are doing? instead of SQL?


joeyAghion:
Hard to say without seeing how such a scope would be _used_, but it does smell like what you _actually_ care about isn't being modeled here. Like, if the "last admin note type" indicates a status, maybe that should simply be stored on the record being filtered.

ashkan18:
it's actually related mostly to https://github.com/artsy/exchange/pull/236, we've been thinking about de-normalizing `last_admin_note` on `Order`. `admin_notes` are not really states since these are basically CRT team's workflow items, kind of the same as assessments in impulse conversations.

PR #253, opened 2018-10-25
bind to 0.0.0.0 for dev

saolsen:
Lets you use hokusai dev.


dleve123:
Ran into some subsequent issues with Postgresql when running `hokusai dev`, but def got past the first issue. Merging.

PR #254, opened 2018-10-26
Update Readme and new danger rule

ashkan18:
Added a danger rule for warning people when they add/update models. 

@orta i was wondering it would be nice if we could automatically label PRs with something like `notifyAnaltics` label or something, but now sure how to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,9 @@ rails db:setup
 $ rails s
 ```
 
+## Did You Change Models?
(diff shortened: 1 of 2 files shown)


anipetrov:
Ah this is awesome @ashkan18! cc @cavvia

orta:
OK, cool, I think this should end up in the auctions channel room too now

PR #256, opened 2018-10-29
Rename CreateOrderService to BaseCreateOrderService

ashkan18:
# Change
@zephraph had this awesome suggestion during our purchase team knowledge share about make it more obvious that `CreateOrderService` is actually our base class and implemented by offer and buy create order services.

# Solution
Rename `CreateOrderService` to `BaseCreateOrderService`.

### DIFF
--- a/app/services/base_create_order_service.rb
+++ b/app/services/base_create_order_service.rb
@@ -1,4 +1,4 @@
-class CreateOrderService
+class BaseCreateOrderService
   attr_reader :order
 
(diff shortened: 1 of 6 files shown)


zephraph:
@ashkan18 looks good! Only question I have... there's a new file `spec/services/create_offer_order_service_spec.rb` is that expected to be a part of this commit? I thought it might've been a weird git renaming thing, but it doesn't look like it.

ashkan18:
yes, i basically added a missing spec as part of this since i had to rename the base class anyways. should be 👍 since it's just a missing spec

PR #257, opened 2018-10-29
Skip verify_authenticity_token for API endpoints

ashkan18:
# Problem
Getting LOT of these warnings
```
Can't verify CSRF token authenticity.
```
in all of our API requests.

We've set
```
protect_from_forgery with: :null_session
```
in `BaseApiConroller` before which seems to allow calls without CSRF and just nullifies the session, but the warning message is still there!

# Possible Solution?
Skip `verify_authenticity_token` in API controller all together!

I tried to do few other things like adding 👇 to main controller
```
protect_from_forgery unless: -> { request.format.json? }
```
which didn't work, I think we are safe with skipping this on API controllers, but let me know if you think otherwise.

### DIFF
--- a/app/controllers/api/base_api_controller.rb
+++ b/app/controllers/api/base_api_controller.rb
@@ -1,7 +1,7 @@
 module Api
   class BaseApiController < ApplicationController
     skip_before_action :require_artsy_authentication
+    skip_before_action :verify_authenticity_token
     before_action :authenticate_request!
-    protect_from_forgery with: :null_session
   end
 end


yuki24:
I think a simpler way may be to change `ApplicationController` to inherit from `ActionController::API` rather than `ActionController::Base`. I believe Exchange is JSON-only and it shouldn't hurt that much even if we stepped away from HTML rendering. Is there any compatibility issue with `ActionContoller:API` that we are aware of?

ashkan18:
would that impact active admin pages in any ways?

yuki24:
Seems like [it does](https://github.com/activeadmin/activeadmin/issues/4712#issuecomment-272123123). I don't really like ActiveAdmin's design decision, but either way, the admin UI and the JSON interface are two distinct things. I would recommend having two controllers for each otherwise things will start coupling and one of them will affect the other someday:

```ruby
# app/controllers/api_controller.rb
class ApiController < ActionController::API
  # same stuff in ApplicationController
end

module Api
  class GraphqlController < ApiController
    ...
  end
end
```

What do you think?

yuki24:
I'll just go ahead and merge this but we should re-visit this as it seems a little more coupled than it should.

PR #258, opened 2018-10-29
[WIP] Create Buyer Offer Mutation

ashkan18:
Opening this early to get early feedback while working on it with @ds300.

@ds300 , went through offer flow again with @sweir27 changed few things from our original design together:
- When buyer submits their first offer, order is still in `pending` state. Then they set shipping and payment and then submit. Ended up removing `negotiation` state.
- Still leaving `negotiate` action on `order` which mainly does a lock on the order and makes things little easier and nicer to read, but im not yet 100% sure about it.
- We need to also switch to old style of mutation responses, following #259 
- Looks like it'll be easier to update `BuyerOffer` mutation to return `order` instead of newly created offer, we may still want to somehow offer `current_offer` as a helper field along with `offers` but for that one we can wait till we find the usage for it.


db/migrate/20181029122615_create_offers.rb
```diff
@@ -0,0 +1,15 @@
+class CreateOffers < ActiveRecord::Migration[5.2]
+  def change
+    create_table :offers, id: :uuid do |t|
+      t.references :order, foreign_key: true, type: :uuid
+      t.integer :amount_cents
+      t.string :offerer_id
+      t.string :offerer_type
+      t.string :state
+      t.string :offered_by
```
ashkan18: - `from_id`
- `from_type`
- `creator_id`
- `resolved_at`
- `resolved_by_id`
- `state`

app/models/order.rb
```diff
@@ -67,6 +67,7 @@ class Order < ApplicationRecord
   has_many :transactions, dependent: :destroy
   has_many :state_histories, dependent: :destroy
   has_many :admin_notes, dependent: :destroy
+  has_many :offers, dependent: :destroy
```
ashkan18: add `last_offer_id`

app/models/order.rb
```diff
@@ -182,6 +183,7 @@ def state_machine
 
   def build_machine
     machine = MicroMachine.new(state)
+    machine.when(:negotiate, PENDING => PENDING, SUBMITTED => SUBMITTED)
```
ashkan18: remove this transition since there is no transition.

app/models/order.rb
```diff
@@ -45,7 +45,7 @@ class Order < ApplicationRecord
     SHIP = 'ship'.freeze
   ].freeze
 
-  ACTIONS = %i[abandon submit approve reject fulfill seller_lapse refund].freeze
+  ACTIONS = %i[abandon negotiate submit approve reject fulfill seller_lapse refund].freeze
```
ashkan18: remove `negotiate` action

sweir27:
re: when to add `current_offer`-- the moment we start using the mutation we'll need some way of figuring out what the recently-created offer was. So it can come via a helper (maybe `last_buyer_offer`) or we will need some way of filtering it out.

ashkan18:
to make things easier for me and @ds300 we are going to merge this to a feature branch on artsy/exchange and open a PR from that to `master` so we can both push to that feature branch

ashkan18:
circle is having issues, builds are fine locally, going to merge and open a new WIP to `master`

PR #259, opened 2018-10-29
Use old-style for new createOfferOrderWithArtwork mutation

sweir27:
Per our discussion on https://github.com/artsy/metaphysics/pull/1354 this updates the new mutation to use the old-style, until we land on a solution that we feel confident in.

Note that all other mutations still use this same scheme, so we'd have to go back and update anyways.


db/schema.rb
```diff
@@ -109,7 +109,6 @@
     t.string "mode", null: false
     t.index ["buyer_id"], name: "index_orders_on_buyer_id"
     t.index ["code"], name: "index_orders_on_code"
-    t.index ["mode"], name: "index_orders_on_mode"
```
sweir27: I didn't change anything, but maybe the schema file wasn't updated after the changes last time?

sweir27:
Updated!

PR #260, opened 2018-10-30
Add system specs to Exchange Admin

ansor4:
This performs the following capybara tests:

Add mock order with status pending
1. Open the exchange admin page, should default to Active Orders and so doesn't show the mock order.
2. Click the "All" orders tab to show all orders, the mock order should show up
3. Click into mock order, this should render the order details page (this currently doesn't work)


spec/rails_helper.rb
```diff
@@ -20,3 +22,9 @@
     ActiveJob::Base.queue_adapter = :test
   end
 end
+
+RSpec.configure do |config|
```
dleve123: I think this config was part of a previous attempt to use the engine routes directly. Can it be removed?

spec/system/visiting_order_details_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      created_at: created_at,
+      updated_at: 1.day.ago,
+      shipping_total_cents: 100_00,
+      commission_fee_cents: 50_00,
+      commission_rate: 0.10,
+      seller_total_cents: 50_00,
+      buyer_total_cents: 100_00,
+      items_total_cents: 0,
+      state: state,
+      state_reason: state == Order::CANCELED ? 'seller_lapsed' : nil
```
dleve123: Considering that we state the state L6, I don’t think this conditional is needed. Thoughts on removing and listing as nil?

spec/system/visiting_order_details_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    )
+  end
+
+  it 'renders order overview page' do
+    allow_any_instance_of(ApplicationController).to receive(:require_artsy_authentication)
+    visit '/admin'
```
dleve123: Typically for any sort of test, we want an assertion. It’s good that we visit the page, but I would recommend an assertion on the page having some content, for example, the ID of the order. Such a test would effectively test that we’ve “wired up” active admin correctly

spec/system/visiting_order_details_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    row_selector = 'td.col.col-code'
+
+    expect(page).to_not have_selector(row_selector)
+
+    tab_selector = 'li.scope.all'
+    within tab_selector do
+      click_link 'All'
+    end
+
+    expect(page).to have_selector(row_selector)
```
dleve123: Considering that feature tests are slow, I typically prefer to have multiple assertions in the same test for identical user flows. Seems like this test and the below are the same user flow. Thoughts on combing?

spec/system/visiting_order_details_spec.rb
```diff
@@ -0,0 +1,72 @@
+require 'rails_helper'
+
+RSpec.describe 'visit order details', type: :system do
+  let(:partner_id) { 'partner1' }
+  let(:user_id) { 'user1' }
+  let(:state) { Order::PENDING }
+  let(:created_at) { 2.days.ago }
+  let!(:order) do
+    Fabricate(
```
dleve123: Thoughts on moving all of the attribute values that are irrelevant to the test to the factory definition itself? It would lead to a more succinct test IMO.

dleve123:
Also, looks like there's an issue with the chrome driver on CI. I _think_ we'll need to edit the docker image include's hokusai to include that driver, or edit the build to use a browser driver that's already included in the Docker image.

izakp:
@jonallured maybe you have some insight on the best way forward to install Chromedriver?  Some relevant commits in Currents, installing ChromeDriver via a CI script (but requires root):
- https://github.com/artsy/currents/commit/e25a9d4bbd987a1a0c19fd3a447b64df733d9d07
- https://github.com/artsy/currents/commit/31c1b1af7ddf5067bd82bc3be9bfaaeea723161c

Looking at [how CircleCI sets up their base image](https://github.com/CircleCI-Public/circleci-dockerfiles/blob/master/ruby/images/2.4.3-jessie/node-browsers/Dockerfile#L58) I wonder if we can use a [Multi-Stage Build](https://docs.docker.com/develop/develop-images/multistage-build/) and simply copy over the compiled `chromedriver` binary into either the image build, or on top of the base build via a `Dockerfile.test`, rather than going through the rigamarole of repeating the installation everywhere.  Will do some more research to see what other patterns may be out there.

jonallured:
I might be missing some nuance here, but my general approach to Rails apps on k8s has been to use base images that I control, like this one:

https://github.com/artsy/docker-images/blob/0cab83fc35f35ce4e9c4b699986260d27d8c2170/ruby/2.5.3-node-chrome/Dockerfile

So that simplifies the Dockerfile for currents like so:

https://github.com/artsy/currents/blob/5cda65be763d218ea1b243dba52d2bc7dd4c9ae8/Dockerfile#L1

I update those images as new versions of Ruby come out, so maybe not as fast as some automated process, but generally pretty good! Does that answer your questions @izakp?

ansor4:
thanks @jonallured! I took a couple lines out of your ruby:2.5.3-node-chrome image to install chrome and tests are happy now. Looks like I can just use that image as a base image for exchange, so I'm going to roll with that.

spec/system/visiting_order_details_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    expect(page).to_not have_selector(row_selector)
+
+    tab_selector = 'li.scope.all'
+    within tab_selector do
+      click_link 'All'
+    end
+
+    expect(page).to have_selector(row_selector)
+
+    within row_selector do
```
starsirius: Minor but _I think_ if we are scoping down with `within`, the `expect` line above becomes unnecessary. Capybara will find the selector and fail if not found.

Dockerfile
```diff
@@ -1,7 +1,7 @@
-FROM ruby:2.5.3
+FROM artsy/ruby:2.5.3-node-chrome
```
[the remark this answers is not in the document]
joeyAghion: Ugh sorry... I forgot to "submit" this^ review after writing this comment yesterday.

Have discussed this with a few of you in the meantime, and I think I understand that this is the simplest/straightforward way to accomplish this until we figure out a better pattern for separating test dependencies from build artifacts. In the meantime, worth keeping an eye on image size to see how it's impacted.

spec/rails_helper.rb
```diff
@@ -6,12 +6,23 @@
...
+require 'selenium/webdriver'
 # Add additional requires below this line. Rails is not loaded until this point!
 
+WebMock.disable_net_connect!(allow_localhost: true)
+
 Dir['./spec/support/**/*.rb'].each { |f| require f }
 
 ActiveRecord::Migration.maintain_test_schema!
 
+Capybara.register_driver :chrome_headless do |app|
```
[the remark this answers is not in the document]
ansor4: Ah yeah this is relic from when I was getting chrome to work, I'll switch it out. The biggest thing here is actually disabling sandbox. Since @jonallured's base image already disables sandbox via `sed`'s find-replace on the chrome options file, it's not needed here.

In the future, do you think the chrome options should live in Dockerfile or rails_helper? I actually think the way to do this from now on would be to install chrome as usual in the Dockerfile, but change the options in the app itself, which I find more reliable and clear. But really this is all stemming from my PTSD from using `sed`, so I'd love to get everyone's take on it.

Dockerfile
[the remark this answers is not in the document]
starsirius: Chrome may be big. fwiw for now Volt uses the official ruby image and  [installs Chrome in `ci.sh`](https://github.com/artsy/volt/blob/f22d9e43d76796d76fa020defa147285a70dc0c9/hokusai/ci.sh#L28-L32) every time until we tried out more advanced solution like multi-stage build.

spec/rails_helper.rb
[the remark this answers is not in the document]
starsirius: Yeah, if it's configurable here, I feel that's more flexible.

spec/system/visiting_order_details_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  before do
+    stub_request(:get, "http://exchange-test-gravity.biz/user/#{order.buyer_id}")
+      .with(
+        headers: {
+          'Accept' => '*/*',
+          'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
+          'User-Agent' => 'Faraday v0.15.3',
+          'X-Xapp-Token' => 'https://media.giphy.com/media/yow6i0Zmp7G24/giphy.gif'
+        }
+      )
```
yuki24: You don't have to specify headers if that's trivial in test.

spec/rails_helper.rb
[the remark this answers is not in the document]
izakp: Aha this is becoming clearer - I would also _love_ to not have to use `sed` as part of a build step, and likewise I think any configuration, i.e. chrome options should be out of the Dockerfile, which we should think of as toolchain / build deps only, and thus am very pro- setting it in teh app itself. 🎉

izakp:
That just leaves this last consideration... do we even need to do this?  CircleCI is setting the X virtual framebuffer opts like so...

```
# start xvfb automatically to avoid needing to express in circle.yml
ENV DISPLAY :99
RUN printf '#!/bin/sh\nXvfb :99 -screen 0 1280x1024x24 &\nexec "$@"\n' > /tmp/entrypoint \
  && chmod +x /tmp/entrypoint \
        && sudo mv /tmp/entrypoint /docker-entrypoint.sh
```

But we don't seem to care in https://github.com/artsy/docker-images/blob/master/ruby/2.5.3-node-chrome/Dockerfile

If it's not important (or could be configured at runtime rather than as part of these build deps) all the better...

ansor4:
Oh interesting. I think that makes the `--headless` part of the config redundant. It looks like the xvfb vs headless chrome debate is pretty lively on the internets, but it _seems_ like best practices are veering towards headless chrome. People cite less resource consumption, simpler setup, and ability to have separate sessions as well as screenshots.

According to google's own headless chrome docs, apparently the `--no-sandbox` thing is not needed if you properly set up a user in the container, so I'll try that on the next iteration. That way I think we'd just be able to use `selenium_chrome_headless` as the spec driver.
https://developers.google.com/web/updates/2017/04/headless-chrome
https://github.com/ebidel/lighthouse-ci/blob/master/builder/Dockerfile#L35-L40

PR #131, opened 2018-08-27
Persist total cents on order model

ashkan18:
# Problem
We are currently calculating `items_total_cents` on the fly when requested, this makes the query for getting list of orders really slow and we see things like:

```shell
D, [2018-08-27T20:18:29.669240 #6] DEBUG -- : [5c9143e0-aa36-11e8-abaa-5146025d2f97]   Order Load (2.1ms)  SELECT  "orders".* FROM "orders" WHERE "orders"."partner_id" = $1 ORDER BY "orders"."updated_at" DESC LIMIT $2  [["partner_id", "554123737261692b198c0300"], ["LIMIT", 20]]
D, [2018-08-27T20:18:29.674095 #6] DEBUG -- : [5c9143e0-aa36-11e8-abaa-5146025d2f97]    (1.3ms)  SELECT SUM("line_items"."price_cents") FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 113]]
D, [2018-08-27T20:18:29.674904 #6] DEBUG -- : [5c9143e0-aa36-11e8-abaa-5146025d2f97]   CACHE  (0.0ms)  SELECT SUM("line_items"."price_cents") FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 113]]
D, [2018-08-27T20:18:29.675388 #6] DEBUG -- : [5c9143e0-aa36-11e8-abaa-5146025d2f97]   CACHE  (0.0ms)  SELECT SUM("line_items"."price_cents") FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 113]]
D, [2018-08-27T20:18:29.677528 #6] DEBUG -- : [5c9143e0-aa36-11e8-abaa-5146025d2f97]   LineItem Load (1.1ms)  SELECT  "line_items".* FROM "line_items" WHERE "line_items"."order_id" = $1 LIMIT $2  [["order_id", 113], ["LIMIT", 20]]
D, [2018-08-27T20:18:29.680943 #6] DEBUG -- : [5c9143e0-aa36-11e8-abaa-5146025d2f97]    (1.1ms)  SELECT SUM("line_items"."price_cents") FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 163]]
[...]

PR #261, opened 2018-10-30
Initial Offer

ashkan18:
# What is this PR about?

Add Support For creating initial offer

# What is not in this PR?
- Counter offers are not in this PR.
- Offer Order submission, it will be different than buy now order submission for things like, inventory deduction and charge hold and etc.
 
# Changes
This is the first pass in supporting Offers. As initial step, we've:
## Added `Offer` model 
which has
- `id`
- belongs to an `order`
- integer "amount_cents"
- string "from_id": along with `from_type` defines which side's offer is this
- string "from_type": similar to `seller_type` and `buyer_type`
- string "creator_id": id of the person created this offer
- uuid "responds_to_id": id of the offer this offer is respond to (aka. counter offer)
- datetime "created_at", null: false
- datetime "updated_at", null: false

## Updated `Order` model
to include
- `offers`: list of this order's offers
- `last_offer`: pointer to current last offer
- `offer_total_cents`: holds current offer total cents 

## Created `Offers::InitialOfferService`
`services` folder is getting long and ended up added new folder under there for `offer` related services. First service is `Offers::InitialOfferService` which is responsible for the initial offer of an order.

Few notes here:
- Initial offer can only be created when order is `pending` this is mainly based on current UI design
- Counter offers (or responds) will most likely be a separate service since the logic for first and subsequent offers are going to be different.

When we create initial offer, we set the `last_offer`.

## Calculating Fees
Things got little complicated when it comes to calculating fees. 
We now calculate buyer/seller/commission/transaction/tax based on either items total cents or offer total cents based on the order mode. For offer order its all calculated based on `offer_total_cents`.
We still populate `items_total_cents` for offer orders but mostly not used.
It gets little tricky for commission fee cents, we don't store `commission_fee_cents` on `line_items` anymore for Offer Orders. This feels little strange to me, but don't have any way around it.

## GraphQL Changes
We've updated our schema to support these new changes.

### Added `OfferType`
Pretty much matches the `offer` model, for `from` field we followed the same pattern we used for `seller` and `buyer` on the order.

Along with offer type we added `OfferStateEnum` to match offer states. 

### Added `intialOffer` Mutation
**Name**:
`initialOffer`

**Arguments**:
- `orderId`: required
- `amountCents`: required

**Validations**:
- makes sure the creator is the owner of this order

**Returns**:
- Similar to other mutations, `OrderOrFailureUnion`

**Errors**:
- `ValidationError` -> `cant_offer`: When order is not `offer` order or its not in `pending` state
- `ValidationError` -> `invalid_amount_cents`

#### Sample Query
```graphql
mutation($input: InitialOfferInput!) {
  initialOffer(input: $input) {
    orderOrError {
      ... on OrderWithMutationSuccess {
        order {
          id
          buyer {
            ... on Partner {
              id
            }
          }
          seller {
            ... on User {
              id
            }
          }
          offers {
            edges {
              node {
                id
                amountCents
              }
            }
          }
          lastOffer {
            id
            amountCents
            from {
              ... on User {
                id
              }
            }
            creatorId
            respondsTo {
              id
            }
          }
        }
      }
      ... on OrderWithMutationFailure {
        error {
          code
          data
          type
        }
      }
    }
  }
}
```

### Updated OrderType
- added `offers` which is a connection 
  - added arguments to this connection so you can now ask for offers by specific side if needed using `from_id`, `from_type` argument
- added `last_offer`
- added `offer_total_cents`


app/graphql/types/order_type.rb
```diff
@@ -5,30 +5,36 @@ class Types::OrderType < Types::BaseObject
...
   field :shipping_total_cents, Integer, null: true
+  field :state_expires_at, Types::DateTimeType, null: true
+  field :state_reason, String, null: true
+  field :state_updated_at, Types::DateTimeType, null: true
+  field :state, Types::OrderStateEnum, null: false
+  field :offers, Types::OfferType.connection_type, null: true do
+    argument :from_id, String, required: false
+    argument :from_type, String, required: false
+    argument :state, Types::OfferStateEnum, required: false
+  end
```
orta: Cool, this ended up real elegant 👍

app/models/offer.rb
```diff
@@ -0,0 +1,11 @@
+class Offer < ApplicationRecord
+  STATES = [
```
sweir27: It seems confusing to have a `nil` state as opposed to something that could represent what that means (i.e. `PENDING`?) and make it so we don't have to test for existence.

app/graphql/types/order_type.rb
```diff
@@ -5,30 +5,36 @@ class Types::OrderType < Types::BaseObject
...
+  field :last_offer, Types::OfferType, null: true
```
sweir27: In order for us to use this field in the UI (assuming that's what it was for, possible it's also necessary on its own), I think we'll need it to be `last_buyer_offer` and/or `last_seller_offer`. 

I'm thinking of the case where a buyer and seller are both adding offers at the same time. If you're in the buyer's UI, you only care about the one that you'd most recently added.

If this ☝️ seems too niche, we could drop it and just filter on `offers`.

db/migrate/20181029122615_create_offers.rb
```diff
@@ -0,0 +1,21 @@
+class CreateOffers < ActiveRecord::Migration[5.2]
+  def change
+    create_table :offers, id: :uuid do |t|
+      t.references :order, foreign_key: true, type: :uuid
+      t.integer :amount_cents
+      t.string :from_id
+      t.string :from_type
```
sweir27: What will `from_type` be in this case?

I think we'll need a notion of "buyer" and "seller" given that `User` and `Partner` may be on either side.

app/models/offer.rb
ashkan18: yes, we brought this up in platform practice too, the issue with having `pending` is when a new offer comes in we need to go back to existing one and change the state, the alternative that was talked about was to move this state up to the order so we don't have `offer.state` basically.

db/migrate/20181029122615_create_offers.rb
ashkan18: its same as `seller_type` and `buyer_type`, does it matter if the offer was from buyer or seller? if that matters we may need a separate field to capture that but i think that can be figured out by current data if needed.

sweir27: I  think it does matter.

In the checkout flow, we'll scope information by `Buyer` and on the CMS-side, by `Seller`.

For instance, if I want to know the most recent offer that the buyer has made, I'll need to be able to filter by that.

app/models/offer.rb
ds300: I think I prefer moving the accepted/rejected state to the order type, because it's less confusing. It might be like PENDING => (submit first order) => NEGOTIATION => (partner accepts or rejects) => ACCEPTED/REJECTED

ashkan18: for now going to remove the state and `resolved_*` fields from this PR and we can add them to proper place on PR for adding support for accept/reject

app/graphql/types/order_type.rb
ds300: > I'm thinking of the case where a buyer and seller are both adding offers at the same time. If you're in the buyer's UI, you only care about the one that you'd most recently added.

I don't understand this. Can you explain a little more?

db/migrate/20181029122615_create_offers.rb
ashkan18: i think we can do a most recent offer by buyer by querying `offers(from_id: <user_id>)`. I'd say for purpose of this PR, i'd say knowing `last_offer` is something that we'll access often and makes sense to de-normalize on the order model and for any other specific queries we can add them as needed if its not supported by current `offers` arguments. what do you think?

app/graphql/types/order_type.rb
ashkan18: one idea is emerging right now is to restrict the offers and counter offers, meaning if you already have an active offer, you can't update it unless you get a counter offer, i think that may make things little simpler for now and you wouldn't not be able to get into weird situations like what you mentioned @sweir27. 
The idea being we only provide 2 mutations, `initialOffer` that can be used to start offers and `respondToOffer` which gets an offer id of what you are responding to, if what you are responding to is not the latest that should result to an error. 
I think for this PR we should just focus on initial offer though and we can look into counter offers as its own task, what do you think?

jonallured: I might be off base, but I think what we're exploring here is how both the purchase and sell side are going to reflect the state of the order while buyer and seller are negotiating. For CMS, we have our order detail page that will either show a form where a seller can accept/counter/decline an offer or a message saying that your counter has been sent. You can see those screens in Zeplin here:

https://zpl.io/bznyLpE

So, here's some pseudo code that might be a nice way of doing this:

```javascript
const OrderPage = props => {
  const lastOffer = props.order.lastOffer
  const showForm = lastOffer.from_type === "buyer"
  return (
    {!showForm && <OfferSentMessage />
    <OtherStuff />
    {showForm && <OfferForm>}
  )
}
```

Does this seem like the right way to approach this? I'm guessing that on the purchase side, there will be inverted logic to do about the same.

db/migrate/20181029122615_create_offers.rb
jonallured: Agreed, I expect to use a field like this to determine which version of the order detail page to show a seller. Open to other approaches, but I will need something to tell me which end of the deal has last sent in an offer. See pseudo code above. 😝

app/graphql/types/order_type.rb
ashkan18: so current `from_id` , `from_type` are basically same thing as `seller_id` , `seller_type` combination, with current design your pseudo code should more look like:
```
const OrderPage = props => {
  const lastOffer = props.order.lastOffer
  const showForm = lastOffer.from_type === order.buyer_type
  return (
    {!showForm && <OfferSentMessage />
    <OtherStuff />
    {showForm && <OfferForm>}
  )
}
```
where we check offer's `from_type` with order's `buyer_type`.

If we are switching to `buyer`, `seller` we can remove `from_id` since it's not meaningful anymore. 
In that case the question will be who defines `buyer` or `seller`? is that derived based on the mutation?
with graphql we also always have the option of exposing fields that don't map to any database fields, so we could do things like `waiting_buyers_action` and etc.
@joeyAghion since we talked about this before, do you have any preference?

sweir27: Right now, only buyers can make an initial offer, correct? So from that perspective we could figure it out when the offer is created.

For counteroffers, we may need to pass that in or have separate mutations!

joeyAghion: Not sure I follow the "both adding offers at the same time" scenario. At any given point, it's only valid for one or the other party to be creating a new [counter-]offer. I _think_ this can all be determined pretty simply like `lastOffer.from_type == 'User'`, unless I'm misunderstanding. Are we discussing some further denormalization or status to capture at the order level to make this even easier?

app/models/offer.rb
joeyAghion: 👍 to skipping this for now. If you want to avoid a "state" that might be confusing in relation to the order's state, you could have a "resolution" field equal to accepted or rejected (or simply 2 booleans/timestamps).

spec/services/create_offer_order_service_spec.rb
```diff
@@ -25,7 +25,7 @@
             expect(order.line_items.first.artwork_version_id).to eq 'current-version-id'
             expect(order.line_items.first.edition_set_id).to be_nil
             expect(order.line_items.first.quantity).to eq 2
-            expect(order.items_total_cents).to eq 1080024
+            expect(order.items_total_cents).to eq 0
```
williardx: Why did this change?

ashkan18: for offer orders, `items_total_cents` is basically calculated based on offer not line items, now one thing we could do is to have it as a separate field, basically have 
- `items_total_cents` which will be populated the same for both buy and offer orders
- `offer_total_cents` which is basically the amount of current offer, only populated for `offer` orders

db/schema.rb
```diff
@@ -74,6 +74,19 @@
     t.index ["order_id"], name: "index_line_items_on_order_id"
   end
 
+  create_table "offers", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "order_id"
+    t.integer "amount_cents"
+    t.string "from_id"
```
williardx: I'd love to get some documentation for these different attributes since I think it's a little easy to get tripped up on `from_id`, `from_type`, `responds_to`, etc.

spec/services/create_offer_order_service_spec.rb
williardx: Hmm, so `items_total_cents` in that case would be the list price? I think that will make things confusing when we try to figure out what price we need to use for other calculations, e.g. sales tax. I would expect `items_total_cents` to be whatever the latest offer is.

I see now, though, that this should be 0 because it's a newly-created offer without a price attached yet, is that right?

app/services/order_total_updater_service.rb
```diff
@@ -25,6 +25,13 @@ def update_totals!
 
   private
 
+  def calculate_items_total_cents
+    case @order.mode
+    when Order::BUY then @order.line_items.map(&:total_amount_cents).sum
+    when Order::OFFER then @order.last_offer&.amount_cents || 0
```
williardx: This is minor, but `nil` feels more correct than 0 for the `amount_cents` of an offer that hasn't been set yet, mostly since someone could make an offer for $0. What do you think?

app/graphql/types/order_type.rb
```diff
@@ -62,4 +67,12 @@ def display_commission_rate
       strip_insignificant_zeros: true
     )
   end
+
+  def offers(**args)
+    if args.keys.any? { |ar| %i[from_id from_type].include? ar }
+      object.offers.where(args.slice(:from_id, :from_type, :state))
```
starsirius: We want to remove the `:state` from here, too?

app/services/order_total_updater_service.rb
williardx: Nevermind! This makes okay sense to me, I got confused with `items_total_cents` and `last_offer.amount_cents`.

app/graphql/types/order_type.rb
ashkan18: good point! just removed!

app/services/order_total_updater_service.rb
ashkan18: what do you think about that other proposal? basically add a separate field `offers_total_cents` and keep `items_total_cents` to be the actual items total cents and then have a helper method to retrun `order_total_cents` (or something more descriptive) that returns either `items_total_cents` or `offers_total_cents` based on the `order.mode`

app/services/offers/initial_offer_service.rb
```diff
@@ -0,0 +1,27 @@
...
+        @offer = @order.offers.create!(amount_cents: @amount_cents, from_id: @user_id, from_type: Order::USER, creator_id: @user_id)
+        @order.update!(last_offer: @offer)
+        OrderTotalUpdaterService.new(@order).update_totals!
+      end
+    end
+
+    private
+
+    def assert_offer!
+      raise Errors::ValidationError, :cant_offer unless @order.mode == Order::OFFER && @order.state == Order::PENDING
```
starsirius: Just a thought, and I wonder if this validation can be handled by the Order state machine in a similar way.

app/services/offers/initial_offer_service.rb
```diff
@@ -0,0 +1,27 @@
...
+      @amount_cents = amount_cents
+      @user_id = user_id
+      @offer = nil
+    end
+
+    def process!
+      assert_offer!
+      @order.with_lock do
+        @offer = @order.offers.create!(amount_cents: @amount_cents, from_id: @user_id, from_type: Order::USER, creator_id: @user_id)
+        @order.update!(last_offer: @offer)
```
starsirius: Looks like we only allow submitting an initial offer on a pending order. Do we also need to advance the order to the "confirmed" state?

app/services/order_total_updater_service.rb
williardx: If we have something like `order_total_cents` that gives the expected price for any order then I'm 👍

app/services/offers/initial_offer_service.rb
ashkan18: i actually had it originally in the state machine, had `negotiate` as an _action_ but from state machine's perspective it was not doing anything and was little confusing ended up removing it here https://github.com/artsy/exchange/pull/261/commits/409798e5b1a0a2b5289123501d78002b0ea28524#diff-1c5adff4d5826555962bc3f7fcb4faf5L186
if we started doing these checks in few other places we can add a helper for something like `offerable?` one other argument for keeping this here is, initial offer happens on a _pending_ order while counter offers will all happen on _submitted_ order.

spec/services/create_offer_order_service_spec.rb
ds300: We should keep the list price on the order object somehow. We need to show it in the UI during both the 'first offer' and 'counter offer' flows. I suppose it could be calculated from the line items array, but would prefer to have that done in the backend.

ashkan18: yes, makes total sense. working on adding `offer_total_cents` and keep items total cents as it was. Things are actually little more involved since we need to handle commission based on offer price to and etc.

app/services/offers/initial_offer_service.rb
ashkan18: no, based on the current UI flow, an order can go only to submitted state after getting payment and shipping info, so partner's wont see this offer till user goes through full checkout for the initial offer. So till they submit their offer order, its still stays in pending state.

db/schema.rb
ashkan18: good point, not sure where is the best place for these to add description, `schema` file is auto generated and migrations are hard to read and not a good place for documenting things, we do some documenting on graphql side of things but maybe we could also look into using something like [annotate](https://github.com/brentgreeff/annotate)

ashkan18:
going to merge after some discussion https://artsy.slack.com/archives/CB110C6LE/p1541082192040200 and will followup on possible model changes

app/models/order.rb
```diff
@@ -140,6 +146,10 @@ def last_admin_note
     admin_notes.order(:created_at).last
   end
 
+  def current_total_cents
+    mode == OFFER ? offer_total_cents : items_total_cents
```
joeyAghion: (Sorry for the late comment.)

I think it will be confusing that there are so many different totals: `current_...`, `offer_...`, `items_...`, `buyer_...` and `seller_...`. We already have the original list-price of the work on the `LineItem`, so what use is the `items_total_cents` calculated at the order level (since it doesn't accurately reflect what anyone is willing to pay on an offer-mode submitted order)?

Specifically, I fear that many clients will have to repeat exactly this logic in order to render an order and its most relevant amount. It would be easier if the existing total field simply respected the offer. That, combined with constraints enforcing that offer-mode orders only support single line items and quantities, makes these definitions easier to understand.

app/services/offers/initial_offer_service.rb
starsirius: Got it. So offers can be pending as well. The initial offer state can be inferred by the order state, but I'm not exactly sure for the following offers. Feels like it's driven by UI design, but we can iterate! 👍

app/models/order.rb
ashkan18: `items_total_cents` was added as part of https://github.com/artsy/exchange/pull/131 to avoid calculating the total cents every time it was needed since it's a very popular field :) I agree that it's getting little complicated and i had hard time coming of with proper name for this new helper method sine everything was taken!
I think clients want to know the listed price and current offer though.
As for adding constraint for single line item on the order, i think that would simplify everything but at that point you could argue why even have `LineItem`?

ashkan18:
had more discussions about this offline with @joeyAghion and i think we are going to keep `from_id`, `from_type` on the `Offer` model. For clients knowing which side of the offer this last offer came from is useful and we can totally provide that in our Graphql schema and have the database modeling store more informational data.
Having `from_id`, `from_type` on the offer helps in things like being able to authorize the offer model easy without looking at other models and etc.
In terms of Graphql fields exposed maybe we could do something like:

# Offer
add to type:
- `from_buyer` 

add to query:
- `from_buyer` 

# Order
add to type:
- `need_buyer_respond`?

PR #263, opened 2018-10-31
Save user agent and ip address at order creation time

ds300:
Final part of https://artsyproduct.atlassian.net/browse/PURCHASE-574

Follows on from artsy/metaphysics#1360

I'm assuming we don't want to expose these fields through GraphQL.

Slightly unrelated: it's not obvious to me how certain things are injected into the lexical context like the `context` object itself for graphql resolvers, and `request` in the controllers. Is there somewhere I can read about how this kind of ruby magic works?

### DIFF
--- a/db/migrate/20181030164042_add_order_analytics_fields.rb
+++ b/db/migrate/20181030164042_add_order_analytics_fields.rb
@@ -1,0 +1,6 @@
+class AddOrderAnalyticsFields < ActiveRecord::Migration[5.2]
+  def change
+    add_column :orders, :original_user_agent, :string
+    add_column :orders, :original_user_ip, :string
+  end
+end
--- a/app/controllers/api/graphql_controller.rb
+++ b/app/controllers/api/graphql_controller.rb
@@ -6,7 +6,9 @@ module Api
       operation_name = params[:operationName]
       context = {
         # Query context goes here, for example:
-        current_user: current_user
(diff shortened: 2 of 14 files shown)


ashkan18:
following up on discussion in https://artsyproduct.atlassian.net/browse/PURCHASE-574, i think for now we can only store it on order creation and remove the update from submission 🙌

ashkan18:
sorry missed that comment, yes context is the object passed to query/mutation resolver mostly used for passing things like authentication tokens and etc. there are some docs [here](http://graphql-ruby.org/authorization/overview.html)

ashkan18:
Nice! thanks!

db/migrate/20181030164042_add_order_analytics_fields.rb
```diff
@@ -0,0 +1,6 @@
+class AddOrderAnalyticsFields < ActiveRecord::Migration[5.2]
+  def change
+    add_column :orders, :original_user_agent, :string
+    add_column :orders, :original_user_ip, :string
```
starsirius: Do we need to care about PII?

alloy: Good point! GDPR does consider it to be personal data https://ec.europa.eu/info/law/law-topic/data-protection/reform/what-personal-data_en

PR #265, opened 2018-10-31
Confirm order confirm shipment

ansor4:


app/admin/orders.rb
```diff
@@ -43,11 +43,31 @@
     OrderCancellationService.new(resource).refund!
     redirect_to resource_path, notice: "Refunded!"
   end
+  
+  member_action :approve_order, method: :post do
+    OrderApproveService.new(resource).process!
```
ashkan18: nice! lets add `current_user[:id]` to to approval call too.

app/admin/orders.rb
```diff
@@ -43,11 +43,31 @@
...
 
   action_item :refund, only: :show do
     link_to 'Refund', refund_admin_order_path(order), method: :post, data: {confirm: 'Are you sure you want to refund this order?'} if [Order::APPROVED, Order::FULFILLED].include? order.state
   end
 
+  action_item :approve_order, only: :show do
+    link_to 'Approve Order', approve_order_admin_order_path(order), method: :post, data: {confirm: 'Approve this order?'} if order.state == Order::SUBMITTED
+  end
+
+  action_item :confirm_fulfillment, only: :show do
```
ashkan18: if this action only works for `pickup` maybe we can just not show this unless order is `pickup` and its in `Order::APPROVED` state

app/admin/orders.rb
```diff
@@ -43,11 +43,31 @@
...
+      OrderService.confirm_pickup!(resource, current_user[:id])
+    end
+    redirect_to resource_path, notice: "Fulfillment confirmed!"
+  end
 
   action_item :refund, only: :show do
     link_to 'Refund', refund_admin_order_path(order), method: :post, data: {confirm: 'Are you sure you want to refund this order?'} if [Order::APPROVED, Order::FULFILLED].include? order.state
   end
 
+  action_item :approve_order, only: :show do
```
ashkan18: i think we want to only show this action item if order is approvable, so maybe check if `order.state == Order::SUBMITTED`

ansor4: I followed the pattern I saw on the refund button above and tacked that check at the end of the `link_to`. I agree that it's hard to read and think it should be checked before the `action_item` line. Just wanted to make sure that's what you meant?

app/admin/orders.rb
ansor4: Same here, I ended up tacking it on at the end of the `link_to` line below, but I can just as easily move it here.

ashkan18: oh my bad! i totally missed it! 🙌

app/admin/orders.rb
ashkan18: sorry i totally missed it! this should work!

ansor4: Cool just making sure I wasn't messing up! I might try to find a way to move the check a line above anyway since it's easier to see and reason with.

PR #267, opened 2018-11-01
Filter admin actions, add ability to sort

ansor4:
Changes:

- Add two scopes, one for orders with "Case Closed" note, and another for orders with notes on them, but not "Case Closed"
- Add sorting ability based on Last Admin Action. This is how to join against a `has_many` association apparently, I'm not entirely sure how it works.
- Expand the `by_last_admin_note` scope to take an array of note types
- tests for `by_last_admin_note`

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -21,13 +21,19 @@ ActiveAdmin.register Order do
   filter :state, as: :check_boxes, collection: proc { Order::STATES }
   filter :state_reason, as: :check_boxes, collection: proc { Order::REASONS.values.map(&:values).flatten.uniq.map!(&:humanize) }
 
+  controller do
+    def scoped_collection
+      Order.includes :admin_notes
+    end
+  end
+
   index do
     column :code do |order|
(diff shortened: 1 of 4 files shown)


ansor4:
So, there's one thing that I actually want to bring up. I think it might be time to denormalize the last admin action stuff. The code that adds sorting to the last admin action column, needs to call `order.last_admin_note` on every row in the table in order to sort it. Which means an individual call to the `admin_notes` table for every row (it somehow doesn't join this properly).

At this point, I think denormalizing is easier to deal with, it would simplify a lot of the active admin code. Plus it's a bit easier to understand/read from a coding perspective.

ashkan18:
🙌 this makes sense to me! we ended up doing the same thing in Impulse and ended up de-normalizing it. Makes total sense to me, do you still want to merge this PR? looks good to me, we can follow up on denormalizing the data in separate PR

ansor4:
Yeah I think I'll remove the sorting from this PR and merge it. The scopes up top are still good to go and provide a decent stopgap in the meantime.

PR #268, opened 2018-11-01
Always display times in Eastern time, and with the timezone.

pepopowitz:
Pairing with @dleve123, we found a couple of configuration changes we could make so that the Exchange app **always** displayed times in the Eastern time zone, and with the time zone explicitly identified.

The previous iteration of this PR required us to manually convert to the eastern time zone any time we were displaying a date, which was cumbersome and easy to mess up. With the approach in this PR, we can be certain that every date rendered by Exchange will be in Eastern time, so we won't accidentally miss any.

In addition, we added the time zone abbreviation to times when they are displayed. This helps give clarity when some times are in Daylight time and some are in Standard time.

Once we changed to this approach, we decided that writing system tests didn't make sense, as they would only be testing that Active Admin & Rails were respecting our configuration changes.

Screenshots of the new time format:

##### Orders index

Orders index

##### Order details

Order details

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,4 +11,8 @@ class ApplicationController < ActionController::Base
   def set_current_user_for_error_reporting
     Raven.user_context(current_user: current_user) if current_user.present?
   end
+
+  def admin_display_in_eastern_timezone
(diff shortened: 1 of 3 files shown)


ansor4:
Wow, I love this approach! Much cleaner and we don't need to maintain tests!

PR #269, opened 2018-11-05
Add order.mode to posted events

ashkan18:
Fixes https://artsyproduct.atlassian.net/browse/PURCHASE-624

### DIFF
--- a/app/events/order_event.rb
+++ b/app/events/order_event.rb
@@ -1,5 +1,34 @@
 class OrderEvent < Events::BaseEvent
   TOPIC = 'commerce'.freeze
+  PROPERTIES_ATTRS = %i[
+    mode
+    buyer_id
+    buyer_type
+    buyer_phone_number
+    buyer_total_cents
+    code
+    commission_fee_cents
+    created_at
+    currency_code
+    fulfillment_type
+    items_total_cents
+    seller_id
+    seller_total_cents
+    seller_type
+    shipping_address_line1
+    shipping_address_line2
+    shipping_city
+    shipping_country
+    shipping_name
+    shipping_postal_code
+    shipping_region
+    shipping_total_cents state
+    state_reason
+    state_expires_at
+    tax_total_cents
+    transaction_fee_cents
+    updated_at
+  ].freeze
 
   def self.post(order, action, user_id)
     event = new(user: user_id, action: action, model: order)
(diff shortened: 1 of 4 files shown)


app/events/order_event.rb
```diff
@@ -13,36 +13,24 @@ def subject
...
+      buyer_total_cents code commission_fee_cents
+      created_at currency_code fulfillment_type
+      items_total_cents line_items seller_id seller_total_cents
+      seller_type shipping_address_line1 shipping_address_line2
+      shipping_city shipping_country shipping_name
+      shipping_postal_code shipping_region shipping_total_cents state
+      state_reason state_expires_at tax_total_cents
+      transaction_fee_cents updated_at
+    ]
+    Hash[attrs.map { |att| [att, @object.send(att)] }].merge(line_items: line_items_details)
```
yuki24: Very minor but `Hash[attrs.map { ... }]` could be replaced with `attrs.map { ... }.to_h`.

app/events/order_event.rb
```diff
@@ -13,36 +13,24 @@ def subject
...
-      state_expires_at: @object.state_expires_at,
-      tax_total_cents: @object.tax_total_cents,
-      transaction_fee_cents: @object.transaction_fee_cents,
-      updated_at: @object.updated_at
-    }
+    attrs = %i[
+      mode buyer_id buyer_type buyer_phone_number
+      buyer_total_cents code commission_fee_cents
+      created_at currency_code fulfillment_type
+      items_total_cents line_items seller_id seller_total_cents
```
pepopowitz: This is not a significant issue; I'm asking about it more for my understanding of what's happening - 

We probably don't need to specify `line_items` here, right? Because we just replace it in line 27 with `.merge(line_items: line_items_details)`?

app/events/order_event.rb
```diff
@@ -13,36 +13,24 @@ def subject
...
+      seller_type shipping_address_line1 shipping_address_line2
+      shipping_city shipping_country shipping_name
+      shipping_postal_code shipping_region shipping_total_cents state
+      state_reason state_expires_at tax_total_cents
+      transaction_fee_cents updated_at
```
ds300: Another #superminor thing: would prefer to see one element per line in here. It's super hard to read when they're all jumbled up like this. Maybe that's just my prettier-addled brain 😅

app/events/order_event.rb
ashkan18: good catch! yes, we actually set it on 27, `line_items` in that list is misleading 🙌

app/events/order_event.rb
ashkan18: haha, i actually originally had it per line, but our friend the robotic cop didn't like that and complained about the number of lines in the method, let me move `attrs` to its own method maybe?

ds300:
Yay, thanks! :tada:

PR #271, opened 2018-11-05
Accept Offer Mutation

dleve123:
Jira: https://artsyproduct.atlassian.net/browse/SELL-1145

CR Action Items:

- [x] Handle when the offer / order doesn't exist in Exchange
~~- [ ] Capture the user that's accepting an offer~~
- [x] Add TODOs / pending specs for other offer accepting validations
- [x] Rename id to offer_id in `Mutations::SellerAcceptOffer`
- [x] Move offer accept service to `Offers` namespace


dleve123:
@ashkan18 Any early feedback on this would be great - notably we opted to instrument to data dog with `order.approved` as it seemed logical.

app/graphql/mutations/accept_offer.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  argument :id, ID, required: true
+
+  field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure', null: false
+
+  def resolve(id:)
+    offer = Offer.find(id)
+    order = offer.order
+
+    validate_seller_request!(order)
```
ashkan18: _accepting an offer_ could also happen from a buyer side when buyer gets a counter offer and wants to accept it. Here we may have few options:
1) update existing mutation to support accepting by both end and do something like
```ruby
validate_selle_or_buyer_request!(order)
```
2) Have this mutation get a required input asking caller to tell which side of the deal this request is for
```ruby
argument :side, BuyerSellerEnum, required: true
```
3) we can leave this mutation as is, and as part of validation check we are waiting for which side's response and off of that decide if we need to do a `validate_seller_request` or `validate_buyer_request` something like
```ruby
order.need_buyer_response? ? validate_buyer_request!(order) : validate_seller_request(order)
```
4) Have two separate mutations for `buyerSellerAccept` and `sellerOfferAccept`

Knowing that clients for buyer and seller are currently different and each represent one side, i may vote for 4, we can share lot of code on services layer but i think have clean exact mutations have been pretty useful so far and make things easier to read/maintain.

app/graphql/mutations/base_mutation.rb
```diff
@@ -1,9 +1,20 @@
 class Mutations::BaseMutation < GraphQL::Schema::RelayClassicMutation
   def validate_seller_request!(order)
-    raise Errors::ValidationError, :not_found unless order.seller_type != Order::USER && context[:current_user]['partner_ids'].include?(order.seller_id)
+    raise Errors::ValidationError, :not_found unless authorized_to_sell?(order)
```
ashkan18: nice 👍

app/services/offer_accept_service.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def process!
+    @offer.order.approve!
+    instrument_order_approved
+  end
+
+  private
+
+  def instrument_order_approved
+    Exchange.dogstatsd.increment 'order.approve'
```
ashkan18: Awesome 🙌

app/graphql/mutations/accept_offer.rb
dleve123: We'll rename offer accept to seller offer accept to keep things tight.

dleve123:
@ashkan18 Ready for review -- let us know what you think. Once we're aligned, we'll open up a PR against MP.

app/graphql/mutations/seller_accept_offer.rb
```diff
@@ -0,0 +1,20 @@
+class Mutations::SellerAcceptOffer < Mutations::BaseMutation
+  null true
+
+  argument :id, ID, required: true
```
ashkan18: i don't have strong feeling here but maybe we should make this `offer_id`? what do you think?

app/services/offer_accept_service.rb
```diff
@@ -0,0 +1,16 @@
+class OfferAcceptService
+  def initialize(offer)
+    @offer = offer
+  end
+
+  def process!
+    @offer.order.approve!
```
ashkan18: nice! we could add some TODO/stubs for more validations so we don't forget. That could come in separate PR too, for example eventually we want to make sure that the offer you are about to accept is `order.last_offer` and its on your side to accept it and etc.

app/services/offer_accept_service.rb
```diff
@@ -0,0 +1,16 @@
+class OfferAcceptService
```
ashkan18: should we call this `sellerOfferAccept`? Also I started `offers` folder under `services` for _offer_ related services, not sure yet about it but this could go there, so it'll end up being something like `Offers::SellerAcceptService`

app/services/offer_accept_service.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize(offer)
```
ashkan18: i would add `by` as an argument here to capture the user who is accepting this offer and store it on the offer, might be out of scope of this PR so feel free to ignore

app/graphql/mutations/base_mutation.rb
```diff
@@ -1,9 +1,20 @@
...
   def validate_buyer_request!(order)
-    raise Errors::ValidationError, :not_found unless context[:current_user]['id'] == order.buyer_id
+    raise Errors::ValidationError, :not_found unless authorized_to_buy?(order)
+  end
+
+  private
+
+  def authorized_to_sell?(order)
+    current_user_is_on_seller = context[:current_user]['partner_ids'].include?(order.seller_id)
+    order.seller_type != Order::USER && current_user_is_on_seller
```
starsirius: Just to point out the auth difference here in mutation vs. query: we allow admins with `sales_admin` role to [query](https://github.com/artsy/exchange/blob/a612ed632b1b516321f227ee5a034bf1ee81e8e1/app/graphql/types/query_type.rb#L67) order(s), but don't allow them to mutate an order. This feels like the API is read-only for sales admins. I think it's fine to keep things as is, and there were discussion about [admin use cases](https://artsy.slack.com/archives/C9ZJYFDNF/p1540408700000100) that admins should use Exchange admin interface directly, instead of through graphql API.

ashkan18: we have https://artsyproduct.atlassian.net/browse/PURCHASE-627 to remove `sales_admin` access via GraphQL completely. That should cover this issue.

app/services/offer_accept_service.rb
starsirius: 👍 Remembered we talked about this initially and looks like the field was not included in the original PR to simplify things.

app/graphql/mutations/accept_offer.rb
```diff
@@ -0,0 +1,20 @@
+class Mutations::AcceptOffer < Mutations::BaseMutation
+  null true
+
+  argument :id, ID, required: true
+
+  field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure', null: false
+
+  def resolve(id:)
+    offer = Offer.find(id)
+    order = offer.order
```
starsirius: "offer" and "order" look sooooo similar! 😵

starsirius: #minor. Do we need to catch offer/order not found cases? Seems to be reasonable to handle before validating seller request.

app/graphql/mutations/base_mutation.rb
starsirius: Nice! We will need a corresponding ticket on the UI/CMS side to remove support for sales admins. That's probably come earlier than the purchase ticket?

app/graphql/mutations/seller_accept_offer.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure', null: false
+
+  def resolve(id:)
+    offer = Offer.find(id)
+    order = offer.order
```
starsirius: Commented on an outdated commit and not visible here: Pasted again https://github.com/artsy/exchange/pull/271#discussion_r231264461

app/graphql/mutations/base_mutation.rb
starsirius: Ticketed: https://artsyproduct.atlassian.net/browse/SELL-1151

app/services/offer_accept_service.rb
dleve123: Absolutely, I'll def add some TODO and/or pending specs to document this.

app/services/offer_accept_service.rb
dleve123: Def down for moving into `offers/` -- Considering how generic the code is ATM, I'd recommend renaming to `Offers::Accept`, and then renaming / refactoring once we realize how seller accepting and offer accepting differ. Notably, we can rename without changing the GraphQL API.

What do you think? Do you care strongly?

app/graphql/mutations/accept_offer.rb
dleve123: We should def this! Nice call. I'll see if we need to change any implementation.

app/graphql/mutations/base_mutation.rb
dleve123: To confirm, I'm operating off of the understanding that no change needs to be implemented in this initial release WRT this conversation. Someone please correct me if that isn't the case!

app/services/offer_accept_service.rb
dleve123: No, it seems logical and MVP to store this information. We wouldn't want to have any production data that doesn't store the acceptor. I'll investigate and implement if possible.

app/graphql/mutations/accept_offer.rb
dleve123: @ashkan18 How does Exchange's error handling code handle? The spec `xit'`ed out 8fa1393 throws a `Graphlient::Errors::FaradayServerError (the server responded with status 404)` error, which makes sense, but I don't think we're coercing this type of error to one of type `Types::ApplicationErrorType`? Can you point me in the right direction here?

ashkan18: sure, `ActiveRecord::NotFound` is handled in https://github.com/artsy/exchange/blob/master/app/controllers/concerns/error_handler.rb#L15

app/services/offer_accept_service.rb
dleve123: @ashkan18 I see that the current schema for `offers` has `from_id`, `from_type`, `creator_id` and `responds_to_id`, but none of these seem to reflect a "state changer" / "resolver" / someone that could mark this as accepted.

Assuming that this is the case, and considering that I don't _think_ we need this data to facilitate the UI, I think we should add this column as needed, and defer for now.

Does that make sense? What do you think?

app/services/offer_accept_service.rb
dleve123: I opted to implement the `order.last_offer` validation. Gonna pass on the pending specs for "side to accept it and etc" b/c I think it would be a good idea to allow UI flow to drive out these requirements.

dleve123:
@ashkan18 @starsirius I know that y'all already approved the PR, but if either of you want to give some more feedback, please do so / let me know you plan on doing so ~5 PM. 

cc @sepans and @jonallured on this PR as well -- totally forgot to ping you in on the latest changes until now 🤦‍♂️

app/services/offers/accept_service.rb
```diff
@@ -0,0 +1,27 @@
...
+      order.approve!
+      instrument_order_approved
+    end
+
+    private
+
+    attr_reader :order, :offer
+
+    def validate_is_last_offer!
+      raise Errors::ValidationError.new(:not_last_offer, offer) if order.last_offer != offer
```
jonallured: Another way to approach this is to leave the concept of "is this the last offer" up to the offer to answer. I'd expect to be able to do something like this:

```ruby
raise Errors::ValidationError.new(:not_last_offer, offer) unless offer.last_offer?
```

Which I think would just be:

```ruby
# app/models/offer.rb
class Offer < ApplicationRecord
  def last_offer?
    order.last_offer == self
  end
end
```

I like this because it isolates this concept in a model and could possibly persuade us to remove `order` as an argument to this service. If we already have the offer, I'd like to see us not pass in both the offer and the order unless there's a compelling reason.

dleve123: I like the push to a more [Tell, Don't Ask](https://pragprog.com/articles/tell-dont-ask) approach! Will implement this refactoring.

dleve123:
@ashkan18 This should be good to merge when you got a sec!

PR #272, opened 2018-11-06
Add Rake task so that `bundle exec rake` fails when there's schema diff

dleve123:
If a developer forgets to run `rake graphql:schema:idl` and commit the
schema diff after changing GraphQL types, then `bundle exec rake` fails.

* Adds new `rake graphql:schema:diff_check`
* Adds new task as the first step of default Rake task

### DIFF
--- a/lib/tasks/graphql.rake
+++ b/lib/tasks/graphql.rake
@@ -1,19 +1,18 @@
 namespace 'graphql' do
   namespace 'schema' do
+    SCHEMA_FILE_PATH = '_schema.graphql'.freeze
+    ORIG_SCHEMA_PATH = '_schema.graphql.orig'.freeze
+
     desc 'fail if there is ungenerated diff in _schema.graphql'
     task :diff_check do
       puts 'Checking for GraphQL schema diffs...'
-      no_schema_diff_exists = system(
-        'cp _schema.graphql _schema.graphql.orig &&' \
-        'rake graphql:schema:idl > /dev/null && ' \
-        'diff _schema.graphql.orig _schema.graphql'
-      )
 
-      if no_schema_diff_exists
-        system('rm _schema.graphql.orig')
-      else
-        raise failure_message
-      end
+      cp SCHEMA_FILE_PATH, ORIG_SCHEMA_PATH, verbose: false
+      Rake::Task['graphql:schema:idl'].invoke
+
+      raise failure_message unless identical?(SCHEMA_FILE_PATH, ORIG_SCHEMA_PATH)
+    ensure
+      rm(ORIG_SCHEMA_PATH, verbose: false) if File.exist?(ORIG_SCHEMA_PATH)
     end
 
     def failure_message
--- a/lib/tasks/graphql.rake
+++ b/lib/tasks/graphql.rake
@@ -1,13 +1,19 @@
 namespace 'graphql' do
   namespace 'schema' do
-    desc 'fail if there is uncommitted diff in _schema.graphql'
+    desc 'fail if there is ungenerated diff in _schema.graphql'
     task :diff_check do
-      is_successful = system(
+      puts 'Checking for GraphQL schema diffs...'
+      no_schema_diff_exists = system(
+        'cp _schema.graphql{,.orig} &&' \
         'rake graphql:schema:idl > /dev/null && ' \
-        'git diff --exit-code _schema.graphql > /dev/null'
+        'diff _schema.graphql.orig _schema.graphql'
       )
 
-      raise(failure_message) unless is_successful
+      if no_schema_diff_exists
+        system('rm _schema.graphql.orig')
+      else
+        raise failure_message
+      end
     end
 
     def failure_message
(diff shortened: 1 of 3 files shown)


dleve123:
I expect the first commit to fail CI as there's schema diff that's not committed. The new commit should result in a passing test suite.

dleve123:
In the [latest commit][0], we finally see the desirable failure in [this CircleCI build][1]:

```
exchange_1        | Checking for GraphQL schema diffs...
exchange_1        | 297a298
exchange_1        | >   offerTotalCents: Int
exchange_1        | rake aborted!
exchange_1        | 
exchange_1        | You changed your GraphQL types without updating the schema. 
exchange_1        | Run a `rake graphql:schema:idl` and commit to perform the schema update.
exchange_1        | 
exchange_1        | /app/lib/tasks/graphql.rake:15:in `block (3 levels) in <top (required)>'
exchange_1        | /usr/local/bundle/gems/rake-12.3.1/exe/rake:27:in `<top (required)>'
exchange_1        | /usr/local/bundle/bin/bundle:23:in `load'
exchange_1        | /usr/local/bundle/bin/bundle:23:in `<main>'
exchange_1        | Tasks: TOP => default => graphql:schema:diff_check
exchange_1        | (See full trace by running task with --trace)
```

[0]:https://github.com/artsy/exchange/pull/272/commits/878bf65561eb4bc42605b96761e8063ec842e58a
[1]:https://circleci.com/gh/artsy/exchange/1383?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

🍾

dleve123:
This rake task approach could be a good candidate for:

1. A patch to the [graphql-ruby library](https://github.com/rmosolgo/graphql-ruby), if they are interested in a CI-helper like this
2. A common Rake task that we across our GraphQL-exposing services

lib/tasks/graphql.rake
```diff
@@ -0,0 +1,24 @@
+namespace 'graphql' do
+  namespace 'schema' do
+    desc 'fail if there is ungenerated diff in _schema.graphql'
+    task :diff_check do
+      puts 'Checking for GraphQL schema diffs...'
+      no_schema_diff_exists = system(
+        'cp _schema.graphql _schema.graphql.orig &&' \
+        'rake graphql:schema:idl > /dev/null && ' \
```
jonallured: It's sort of a shame that we have to boot up the Rails environment to dump the schema for this task while we're in the default rake task that also has to boot up the Rails environment, but I can't think of a way around it that I like. Unless you're able to get it merged upstream, that is!

yuki24: I'm not sure if it is safe to assume `git` is always available in any env, but  

```ruby
task diff_check: 'graphql:schema:idl' do
  sh 'git diff --exit-code _schema.graphql > /dev/null'
rescue
  raise "\nYou changed your GraphQL types without updating the schema. \n" \
        "Run a `rake graphql:schema:idl` and commit to perform the schema update.\n\n"
end
```

Or even without the extra `before ... rescue ... end` as the `sh` method fails on non-zero status code:

```ruby
task diff_check: 'graphql:schema:idl' do
  sh 'git diff --exit-code _schema.graphql > /dev/null'
end
```

dleve123: @yuki24 

Great points in all respect.

In 7b3595870bcf7c1f42dfd9bfb8dac28571eae9e8 , I attempted to use a `git` based approached, but because we run tests via `hokusai` and `hokusai` uses Docker under the hood, and we remove the `git/` dir when building Docker images, it's necessary to use an approach that doesn't depend on `git`.

WRT to the speed increase, I don't think we can use Rake's pre-req task strategy as my strategy relies upon a `diff` against the orig file, but perhaps there's a more performant way to call a Rake task from a Rake task.

yuki24: You could use the `#invoke` method on the rake task object so you don't have to load the entire app twice. There are also useful file utils methods available in Rake, so we could clean up the task a bit more:

```ruby
SCHEMA_FILE_PATH = "_schema.graphql"
TEMP_SCHEMA_PATH = "_schema.graphql.orig"

desc '...'
task :diff_check do
  puts 'Checking for GraphQL schema diffs...'

  cp SCHEMA_FILE_PATH, TEMP_SCHEMA_PATH
  Rake::Task["graphql:schema:idl"].invoke

  raise failure_message unless identical?(SCHEMA_FILE_PATH, TEMP_SCHEMA_PATH)
ensure
  rm TEMP_SCHEMA_PATH if File.exists?(TEMP_SCHEMA_PATH)
end

def failure_message
  "\nYou changed your GraphQL types without updating the schema. \n" \
    "Run a `rake graphql:schema:idl` and commit to perform the schema update.\n\n"
end
```

dleve123: @yuki24 Very cool, thanks for the suggestions. Implemented your changes.

dleve123:
@ashkan18 this is ready for a final review whenever you are ready

ashkan18:
💯 Thanks for adding this 🥇

PR #273, opened 2018-11-06
Line Item List Price

ashkan18:
# Problem(s)
We introduced `offer` in #261 , with that change, calculating some total amounts got complicated. Generally the relation between an `offer` and `lineItem` is little strange. Offers are made on `order`, they end up defining the total amounts being paid to seller and buyer and should be used to calculate things like shipping and tax as well.

# Solution
It seems the easiest/cleanest solution was to limit the number of line items on offer orders to 1. This means offer orders can't have more than one line item.
The `total_amount_cents` of that line item is calculated by the last offer of the order.
We still want to know the original `list_price` for that order, for that to work, we now have `total_list_price_cents` on `orderType` which basically calculates that by sum of all `list_price * quantity` of `line_items` of the order.

# Deprecated Field
we renamed `price_cents` to `list_price_cents` on `LineItem` to be more accurate about what that field represents. 
From GraphQL perspective we kept `price_cents` around but deprecated it and for now it returns `list_price_cents` so no change needed in clients immediately.

cc: @artsy/analytics  

this addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-578

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -299,6 +299,7 @@ type Order {
     last: Int
   ): LineItemConnection
   mode: OrderModeEnum
+  offerTotalCents: Int! @deprecated(reason: "itemsTotalCents reflects offer total for offer orders.")
   offers(
     # Returns the elements in the list that come after the specified cursor.
     after: String
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -279,6 +279,8 @@ type Order {
   currencyCode: String!
   displayCommissionRate: String
   id: ID!
+
+  # Item total in cents, for Offer Orders this field reflects current offer
   itemsTotalCents: Int!
   lastApprovedAt: DateTime
   lastOffer: Offer
@@ -297,7 +299,6 @@ type Order {
     last: Int
   ): LineItemConnection
   mode: OrderModeEnum
-  offerTotalCents: Int
   offers(
     # Returns the elements in the list that come after the specified cursor.
     after: String
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -191,7 +191,8 @@ type LineItem {
     last: Int
   ): FulfillmentConnection
   id: ID!
-  priceCents: Int!
+  listPriceCents: Int!
+  priceCents: Int! @deprecated(reason: "switch to use listPrice")
   quantity: Int!
   updatedAt: DateTime!
 }
@@ -321,6 +322,7 @@ type Order {
   stateReason: String
   stateUpdatedAt: DateTime
   taxTotalCents: Int
+  totalListPriceCents: Int!
   transactionFeeCents: Int
   updatedAt: DateTime!
 }
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -192,7 +192,7 @@ type LineItem {
   ): FulfillmentConnection
   id: ID!
   listPriceCents: Int!
-  priceCents: Int! @deprecated(reason: "switch to use listPrice")
+  priceCents: Int! @deprecated(reason: "switch to use listPriceCents")
   quantity: Int!
   updatedAt: DateTime!
 }
--- a/app/events/order_event.rb
+++ b/app/events/order_event.rb
@@ -53,7 +53,7 @@ class OrderEvent < Events::BaseEvent
 
   def line_item_detail(line_item)
     {
-      price_cents: line_item.price_cents,
+      price_cents: line_item.list_price_cents,
       artwork_id: line_item.artwork_id,
       edition_set_id: line_item.edition_set_id,
       quantity: line_item.quantity,
--- a/app/events/order_event.rb
+++ b/app/events/order_event.rb
@@ -54,6 +54,7 @@ class OrderEvent < Events::BaseEvent
   def line_item_detail(line_item)
     {
       price_cents: line_item.list_price_cents,
+      list_price_cents: line_item.list_price_cents,
       artwork_id: line_item.artwork_id,
       edition_set_id: line_item.edition_set_id,
       quantity: line_item.quantity,
--- a/app/events/transaction_event.rb
+++ b/app/events/transaction_event.rb
@@ -47,6 +47,7 @@ class TransactionEvent < Events::BaseEvent
   def line_item_detail(line_item)
     {
       price_cents: line_item.list_price_cents,
+      list_price_cents: line_item.list_price_cents,
       artwork_id: line_item.artwork_id,
       edition_set_id: line_item.edition_set_id,
       quantity: line_item.quantity,
--- a/app/events/transaction_event.rb
+++ b/app/events/transaction_event.rb
@@ -46,7 +46,7 @@ class TransactionEvent < Events::BaseEvent
 
   def line_item_detail(line_item)
     {
-      price_cents: line_item.price_cents,
+      price_cents: line_item.list_price_cents,
       artwork_id: line_item.artwork_id,
       edition_set_id: line_item.edition_set_id,
       quantity: line_item.quantity,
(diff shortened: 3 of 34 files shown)


app/events/order_event.rb
```diff
@@ -47,7 +47,7 @@ def properties
 
   def line_item_detail(line_item)
     {
-      price_cents: line_item.price_cents,
+      price_cents: line_item.list_price_cents,
```
williardx: Would it make sense to update this `price_cents` as well for consistency? That could be a separate PR since I know we'd have to update Pulse as well.

app/models/line_item.rb
```diff
@@ -3,7 +3,20 @@ class LineItem < ApplicationRecord
...
   has_many :fulfillments, through: :line_item_fulfillments
 
+  before_create :validate_creation
+
   def total_amount_cents
-    price_cents * quantity
+    order.mode == Order::BUY ? total_list_price : order.last_offer&.amount_cents
+  end
+
+  def total_list_price
```
williardx: Can we make this `total_list_price_cents` to be consistent with our other `foo_cents` fields?

app/services/order_total_updater_service.rb
```diff
@@ -27,15 +26,18 @@ def update_totals!
...
 
-  def calculate_commission_cents
-    if @order.mode == Order::OFFER
-      @order.offer_total_cents * @commission_rate
-    else
-      set_commission_on_line_items
-      @order.line_items.map(&:commission_fee_cents).sum
+  def can_calculate?
+    case @order.mode
+    when Order::BUY then @order.line_items.all? { |li| li.list_price_cents.present? }
```
williardx: For a buy order would we ever have a case where `list_price_cents` was `nil`?

williardx: Ah, I see now that we were checking for that earlier.

spec/models/line_item_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'rails_helper'
+
+describe LineItem, type: :model do
```
williardx: 🥇

_schema.graphql
```diff
@@ -191,7 +191,8 @@ type LineItem {
     last: Int
   ): FulfillmentConnection
   id: ID!
-  priceCents: Int!
+  listPriceCents: Int!
+  priceCents: Int! @deprecated(reason: "switch to use listPrice")
```
joeyAghion: Do you mean `listPriceCents`?

app/events/order_event.rb
joeyAghion: +1

We could start returning a separate `list_price_cents` field in the event now, then remove the deprecated one only once Pulse has switched to using it.

app/events/transaction_event.rb
```diff
@@ -45,7 +45,7 @@ def properties
 
   def line_item_detail(line_item)
     {
-      price_cents: line_item.price_cents,
+      price_cents: line_item.list_price_cents,
```
joeyAghion: Ditto.

app/graphql/types/line_item_type.rb
```diff
@@ -3,12 +3,17 @@ class Types::LineItemType < Types::BaseObject
   graphql_name 'LineItem'
 
   field :id, ID, null: false
-  field :price_cents, Integer, null: false
+  field :price_cents, Integer, null: false, deprecation_reason: 'switch to use listPrice'
```
joeyAghion: Again... `listPriceCents`? `list_price_cents`?

app/graphql/types/order_type.rb
```diff
@@ -30,6 +30,7 @@ class Types::OrderType < Types::BaseObject
     argument :from_id, String, required: false
     argument :from_type, String, required: false
   end
+  field :total_list_price_cents, Integer, null: false
```
joeyAghion: Can't this be `null` for orders on works without buy-now enabled?

app/models/line_item.rb
```diff
@@ -3,7 +3,20 @@ class LineItem < ApplicationRecord
   has_many :line_item_fulfillments, dependent: :destroy
   has_many :fulfillments, through: :line_item_fulfillments
 
+  before_create :validate_creation
```
joeyAghion: You can make this a true validation with `on: :create` rather than a before hook that raises an exception.

```
validate :offer_order_lacks_line_items, on: :create

...

def offer_order_lacks_line_items
  errors.add(:order, "...") if order.mode == Order::OFFER && order.line_items.any?
end
```

app/services/sales_tax_service.rb
```diff
@@ -75,7 +75,7 @@ def fetch_sales_tax
     @tax_client.tax_for_order(
       construct_tax_params(
         line_items: [{
-          unit_price: UnitConverter.convert_cents_to_dollars(@line_item.price_cents),
+          unit_price: UnitConverter.convert_cents_to_dollars(@line_item.total_amount_cents / @line_item.quantity),
```
joeyAghion: It's weird to be redoing (or _undoing_) this calculation here. Line item should expose this as a method (like `effective_price_cents`).

app/models/line_item.rb
```diff
@@ -3,7 +3,20 @@ class LineItem < ApplicationRecord
...
+
+  def total_list_price
+    list_price_cents * quantity
+  end
+
+  private
+
+  def validate_creation
+    # Offer orders can have only one line item
+    raise Errors::ValidationError, :offer_more_than_one_line_item if order.mode == Order::OFFER && order.line_items.exists?
```
yuki24: Here `LineItem` and `Order` are circulating. Is there any reason this validation can't live in `Order`? Also, the `#exists?` method may query against the DB that may return `0` on creation. I think the `with_options` DSL may be helpful:

```ruby
class Order < ApplicationRecord
  ...

  # Common validations
  validates ...

  with_options if: -> { self.mode == Order::BUY } do
    # validations for Buy Now come here
    ...
  end

  with_options if: -> { self.mode == Order::OFFER } do
    # validations for Make Offer come here
    ...
    validates :line_items, length: { is: 1 }, on: :create # No need for a custom validation method as there's a built-in length validator
  end
end
```

app/graphql/types/line_item_type.rb
ashkan18: hehe the original one above is the autogenerated schema file :) 👍

app/graphql/types/order_type.rb
ashkan18: i know currently for make offer works, they should have list price too, but we can remove this restriction too.

app/models/line_item.rb
ashkan18: interesting! didn't know about `with_options`!! seems really useful! 

In this case, `length` seems to not be useful, it does a `length` on String and doesn't seem to work for arrays.
Also another question, if we create a `line_item` by lets say `Fabricate(:line_item, order: <some order>)` does this actually get invoked?

ashkan18:
going to merge on green this one so I can test the whole path on staging with MP and Reaction PRs coming after

_schema.graphql
```diff
@@ -278,6 +279,8 @@ type Order {
   currencyCode: String!
   displayCommissionRate: String
   id: ID!
+
+  # Item total in cents, for Offer Orders this field reflects current offer
   itemsTotalCents: Int!
```
sweir27: Wondering what the reasoning is for having this field be non-nullable.

Currently, if you create an "offer"-mode `Order`, it will initially have a `null` `itemsTotalCents`, until you add an initial offer to it. This seems valid so I'm wondering if it would be 👍 to relax this requirement.

PR #274, opened 2018-11-07
[wip] Update hokusai test environment variables to include CI values

dblandin:
Circle CI 2.0 does not support docker volume mounting or opening ports via the docker executor:

https://discuss.circleci.com/t/why-circleci-2-0-does-not-support-mounting-folders/11605/15

If your CI build needs more control over docker orchestration, including these features, you should be using the machine executor.

Circle CI executor types: https://circleci.com/docs/2.0/executor-types/

To support test coverage reporting via Coveralls while continuing to use the Circle CI docker executor, we can instead pass environment variables from the host to the container.

### DIFF
--- a/hokusai/test.yml
+++ b/hokusai/test.yml
@@ -12,20 +12,21 @@ services:
     - DATABASE_HOST=exchange-db
     - DATABASE_USER=postgres
     - REDIS_URL=redis://exchange-redis
-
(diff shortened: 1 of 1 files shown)


dblandin:
This, unfortunately, doesn't work because the git repository is cloned in the first container, not on the host, so we can't pass any git author information to the hokusai created container.

PR #275, opened 2018-11-07
Relax address validation requirements

williardx:
### Problem
It turns out TaxJar does not need a postal code for _seller nexus addresses_ -- the state is enough for sales tax calculation purposes. (We still need a postal code for shipping/artwork pickup addresses, though.) Additionally, our logic for validating addresses for sales tax calculations lives in the `Address` model when it should more properly live in `SalesTaxService`.

### Solution
Refactor our address validation logic to reside in `SalesTaxService` and not require seller nexus addresses to have a postal code.

### What changed
- Move sales tax address validation logic to `SalesTaxService`
- Separate out address validation based on role in sales tax calculation (i.e., destination address or seller nexus address)
- Create new `effective_shipping_total_cents` that stores the value of what amount we should use for shipping
- Reorder private methods in `SalesTaxService` so that in-use methods are nearer to the top

### DIFF
--- a/lib/address.rb
+++ b/lib/address.rb
@@ -1,6 +1,6 @@
 class Address
   attr_reader :country, :region, :city, :street_line1, :street_line2, :postal_code
-
+  UNITED_STATES = Carmen::Country.coded('US')
   def initialize(address)
     @address = parse(address)
     @country = @address[:country]
@@ -21,15 +21,14 @@ class Address
   end
 
   def united_states?
-    @country == Carmen::Country.coded('US').code
+    @country == UNITED_STATES.code
   end
 
   # The "continental United States" is defined as any US state that isn't Alaska or Hawaii.
   def continental_us?
-    us = Carmen::Country.coded('US')
     united_states? &&
-      @region != us.subregions.coded('HI').code &&
-      @region != us.subregions.coded('AK').code
+      @region != UNITED_STATES.subregions.coded('HI').code &&
+      @region != UNITED_STATES.subregions.coded('AK').code
   end
 
   private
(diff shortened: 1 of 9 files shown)


lib/address.rb
```diff
@@ -21,10 +21,14 @@ def ==(other)
       @postal_code == other.postal_code
   end
 
+  def united_states?
+    @country == Carmen::Country.coded('US').code
```
joeyAghion: `Carmen::Country.coded('US')` could maybe be a constant instantiated just once, since it's used here and a few times below.

williardx:
AFAICT TaxJar only needs the postal code of the destination address to calculate local sales tax, though we're not using local sales tax for now.

Calculating sales tax occurs during the set shipping step, so I _think_ it's the earliest step possible we can do that, at least for shipping addresses. CMS still needs to enforce some some basic validation logic on the seller side. 

And yep, these are TaxJar-specific validations.

lib/address.rb
```diff
@@ -54,17 +58,5 @@ def parse_region(country, region)
 
   def validate!
     raise Errors::AddressError, :missing_country if @address[:country].nil?
```
ashkan18: do you think we still need this check here? I know https://github.com/artsy/exchange/pull/275/files#diff-8c9ae1109745b69ae157f9587a8b80d9R6 was added so showing addresses from Admin would work, if we could move this one out of initialize we could remove `validate` argument too.
Maybe we could generally change this `Address` class to not `raise` anything and instead provide methods that caller could use to see if they need to raise errors or not.
This ☝️ could be a follow up small PR too.

lib/address.rb
williardx: Good point! I'll change that.

lib/address.rb
williardx: I like that -- I don't _think_ it'll break anything if we just take it out now since in the sales tax code we already filter out addresses that aren't in the US, which will take out addresses with `nil` country, too. As far as I know nothing else depends on `Address` in a serious way.

williardx: Okay, I took this out and moved the validation to `OrderShippingService`.

PR #276, opened 2018-11-08
Verify that credit given in set payment step belongs to buyer

williardx:
### Problem
When setting the buyer's credit card on an order, we don't currently check whether or not the credit card is actually associated with the buyer's user. This creates a vulnerability where an attacker could pass in any credit card ID they choose.

Resolves [PURCHASE-583](https://artsyproduct.atlassian.net/browse/PURCHASE-583).

### Solution
Verify that the user ID on the credit card ID given in `OrderService.set_payment!`  matches the buyer's user ID on the order.

### DIFF
--- a/app/controllers/errors/error_types.rb
+++ b/app/controllers/errors/error_types.rb
@@ -11,6 +11,7 @@ module Errors
       invalid_amount_cents
       invalid_artwork_address
       invalid_commission_rate
+      invalid_credit_card
       invalid_offer
       invalid_order
       invalid_seller_address
@@ -31,7 +32,6 @@ module Errors
       missing_required_info
       missing_required_param
       no_taxable_addresses
-      non_buyer_credit_card
       not_acquireable
       not_found
       not_last_offer
--- a/app/controllers/errors/error_types.rb
+++ b/app/controllers/errors/error_types.rb
@@ -31,6 +31,7 @@ module Errors
       missing_required_info
       missing_required_param
       no_taxable_addresses
+      non_buyer_credit_card
       not_acquireable
       not_found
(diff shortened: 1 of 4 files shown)


app/controllers/errors/error_types.rb
```diff
@@ -29,6 +29,7 @@ module Errors
       missing_required_info
       missing_required_param
       no_taxable_addresses
+      non_buyer_credit_card
```
ashkan18: #minor i wonder if we should make this little more generic, like `invalid_credit_card` or something like that.

spec/services/order_service_spec.rb
```diff
@@ -30,9 +44,6 @@
     end
   end
 
-  describe 'set_shipping!' do
```
ashkan18: 🙌

sweir27:
@ashkan18 It should be okay!

We catch errors from that mutation here: https://github.com/artsy/reaction/blob/master/src/Apps/Order/Routes/Payment/index.tsx#L364

And will show the default error modal.

app/controllers/errors/error_types.rb
ashkan18: i was just thinking if this reveals too much in case someone was trying to use this credit card, but maybe not, since all we are saying is just it doesn't match buyers.

williardx: That's a good point - it feels better from a hygiene perspective to just say `invalid_credit_card`. I'll update that.

williardx: FWIW it's only meaningful if we use it elsewhere, but we can update that later.

PR #277, opened 2018-11-08
Add submitted_at to Offer

ashkan18:
# Problem
Offers can be created without being submitted yet. We want to know about un-submitted offers and also we want to know when they are submitted.

# Solution
Added `submitted_at` to `Offer` and expose it in the GraphQL response.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -243,6 +243,7 @@ type Offer {
   id: ID!
   order: Order!
   respondsTo: Offer
+  submittedAt: DateTime
 }
 
 # The connection type for Offer.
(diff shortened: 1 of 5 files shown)


_schema.graphql
```diff
@@ -243,6 +243,7 @@ type Offer {
   id: ID!
   order: Order!
   respondsTo: Offer
+  submittedAt: DateTime
```
sweir27: Nice and simple!

Wondering if we even need this exposed in graphQL yet. I imagine clients will want to be able to easily retrieve the most relevant offer to them, or on a given offer be able to determine whether or not the offer is submitted through an `isSubmitted` boolean.

PR #281, opened 2018-11-09
Add mutation for submitting an offer-type order

sweir27:
This PR is tracked in: https://artsyproduct.atlassian.net/browse/PURCHASE-355

It adds a mutation for submitting an order with an offer, called: `submitOrderWithOffer`.

image

The mutation accepts an `offerId` only, since we can infer the `orderId` from that.

Submitting an order with an offer doesn't have to hold a transaction, as with a buy-type order. Instead, it merely marks the order as `'submitted'` and updates the `submitted_at` on the offer.


williardx:
This looks good so far! When do we actually charge the user for an offer order though?

app/models/offer.rb
```diff
@@ -5,4 +5,8 @@ class Offer < ApplicationRecord
   def last_offer?
     order.last_offer == self
   end
+
+  def submitted?
+    submitted_at.present?
+  end
```
yuki24: Very minor but you could also add an alias:

```ruby
alias submitted? submitted_at?
```

The `[attribute_name]?` is a method that Rails adds automatically that is occasionally useful.

sweir27:
@williardx we charge them when an offer is accepted (on the buyer or seller's side). For the offer flow we're not holding the charge for any length of time like we do for buy now.

ashkan18:
Merging this, it seems it won't impact our latest changes we are talking about.

PR #282, opened 2018-11-13
Filter out pending offers in order.offers

ashkan18:
# Problem
We want to filter out pending offers when returning `order.offers` in GraphQL by default and we want to allow option to adding them if needed.
https://artsyproduct.atlassian.net/browse/PURCHASE-649

# Solution
Added a new scope for `submitted` offers to `Offer` model.
Updated `order.offers` to by default use `submitted` scope, and in case newly optional `includePending` was passed, add them to the response.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -313,6 +313,7 @@ type Order {
     first: Int
     fromId: String
     fromType: String
+    includePending: Boolean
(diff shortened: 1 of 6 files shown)


sweir27:
Wahoo!

PR #283, opened 2018-11-14
My Last Offer

ashkan18:
# Change That Is Less Dramatic Than How This PR's Subject Sounds
We can have _pending_ offers, yet need to be submitted, by both buyers and sellers. We want to expose current user's last offer in the `OrderType` returned from Mutations and Queries to also return current's users Last Offer so we can see the order from their current perspective.

This is also a side-effect of another decisions we've made where we are going to set `last_offer` on an Order only when an offer is submitted.

We also added `tax_total_cents` and `shipping_total_cents` to `Offer` so we can see the impact of current pending offer while its not submitted yet on the order. Once the offer is submitted, these amounts should be set on the Order.

https://artsyproduct.atlassian.net/browse/PURCHASE-651

# TLDR
- [ ] Added `myLastOffer` to `OrderType`: returns current user's last offer (pending or submitted)
- [x] Set `Order`'s `last_offer` only when submitting an `Offer`
- [ ] Added `tax_total_cents` and `shipping_total_cents` to `Offer`
- [x] Calculate ☝️ when shipping info is set
- [ ] Added `waitingBuyerResponse` and `waitingSellerResponse` to `OrderType`
- [ ] Rename `initialOffer` mutation to `addInitialOfferToOrder` **deprecation**
- [x] Refactored `SalesTaxService` to `Tax::CalculatorService` and `Tax::CollectorService`
  - [x] `Tax::CalculatorService` doesn't know about `LineItem`s and only gets values needed to calculate. This way we can use same service for calculating tax on an `Offer`
  - [x] `Tax::CollectorService` gets `LineItem`
- [x] Moved order shipping calculation to its own service
- [x] Updated `OrderShippingService` to based on order's `mode` decide if we need to update totals on the `Order` or on the `Offer`.
- [x] Updated `Offers::SubmitOrderService` to do some validations on the order, credit card, partner and etc. and also set `lastOffer` on the Order to the current submitted one.
# Anything Got Deprecated?
Yes, we deprecated `initialOffer` mutation to `addInitialOfferToOrder`.

# Selfie


sweir27:
@ashkan18 can we remove the `initialOffer` rename from this PR?

If you want to keep it here, at least let's make it not a breaking change as we've been consistently exercising the offer flow in staging.

ashkan18:
had an offline chat with @sweir27 about the breaking change, we are going to try to see if we can easily keep both mutations around for a while and deprecate the old one to have 0 breaking change. Maybe another alternative here could be to handle this from MP but going to first try it from Exchange.

ashkan18:
Alright our original idea of just doing 
```ruby
field :initial_offer, mutation: Mutations::AddInitialOfferToOrder
```
didn't work, it uses the class name to create all the input and payload types.
BUT! I ended up using the new class-based mutation and add back `initialOffer` but deprecate it and create `Mutations::InitialOffer` that all it does is
```ruby
class Mutations::InitialOffer < Mutations::AddInitialOfferToOrder
end
```

app/services/tax/calculator_service.rb
```diff
@@ -0,0 +1,119 @@
+class Tax::CalculatorService
```
williardx: We should have a `BaseTaxService` class that encapsulates the shared logic between `CalculatorService` and `CollectorService` -- e.g. TaxJar client, `construct_tax_params` (which are common to every call and could probably be better named), etc.

app/services/tax/collector_service.rb
```diff
@@ -1,11 +1,8 @@
-class SalesTaxService
+class Tax::CollectorService
```
williardx: I can't think of a better name for this off the top of my head, but it strikes me that "collector" is an overloaded term in our line of work 😝 Maybe `CollectionService` instead?

app/graphql/mutations/add_initial_offer_to_order.rb
```diff
@@ -0,0 +1,18 @@
+class Mutations::AddInitialOfferToOrder < Mutations::BaseMutation
```
williardx: (Ashkan notes)
1. This is just a renamed mutation for clarity to emphasize that it returns an order.

app/graphql/mutations/initial_offer.rb
```diff
@@ -1,18 +1,2 @@
...
-  def resolve(order_id:, amount_cents:)
-    order = Order.find(order_id)
-    authorize_buyer_request!(order)
-    service = Offers::InitialOfferService.new(order, amount_cents, context[:current_user]['id'])
-    service.process!
-    { order_or_error: { order: service.order } }
-  rescue Errors::ApplicationError => application_error
-    { order_or_error: { error: Types::ApplicationErrorType.from_application(application_error) } }
-  end
+class Mutations::InitialOffer < Mutations::AddInitialOfferToOrder
```
williardx: 2. In order to deprecate this old mutation that does what #1 does, we have to inherit from it.

app/graphql/types/mutation_type.rb
```diff
@@ -9,6 +9,7 @@ class Types::MutationType < Types::BaseObject
   field :reject_order, mutation: Mutations::RejectOrder
   field :fulfill_at_once, mutation: Mutations::FulfillAtOnce
   field :confirm_pickup, mutation: Mutations::ConfirmPickup
-  field :initial_offer, mutation: Mutations::InitialOffer
+  field :add_initial_offer_to_order, mutation: Mutations::AddInitialOfferToOrder
+  field :initial_offer, mutation: Mutations::InitialOffer, deprecation_reason: 'Switch to addInitialOfferToOrder.'
```
williardx: 3. This field is being deprecated in favor of `add_initial_offer_to_order`

app/graphql/types/order_type.rb
```diff
@@ -31,12 +31,15 @@ class Types::OrderType < Types::BaseObject
     argument :from_type, String, required: false
     argument :include_pending, Boolean, required: false
   end
+  field :my_last_offer, Types::OfferType, null: true
```
williardx: 4. `my_last_offer` can be either a submitted or a pending offer, which contrasts with `last_offer` which is always a submitted offer. `last_offer` could be buyer or seller offer, and we need to distinguish between what the user might want to do and what happened last.

app/graphql/types/order_type.rb
```diff
@@ -31,12 +31,15 @@ class Types::OrderType < Types::BaseObject
...
   field :total_list_price_cents, Integer, null: false
   field :offer_total_cents, Integer, null: false, deprecation_reason: 'itemsTotalCents reflects offer total for offer orders.'
-  field :last_offer, Types::OfferType, null: true
+  field :last_offer, Types::OfferType, null: true, description: 'Last submitted offer'
   field :tax_total_cents, Integer, null: true
   field :transaction_fee_cents, Integer, null: true, seller_only: true
   field :updated_at, Types::DateTimeType, null: false
+  field :waiting_buyer_response, Boolean, null: true
```
williardx: 5. These just mark whether or not we're waiting for a response -- there's nothing in the schema that corresponds to this.

app/graphql/mutations/set_shipping.rb
```diff
@@ -10,7 +10,8 @@ class Mutations::SetShipping < Mutations::BaseMutation
   def resolve(id:, fulfillment_type:, shipping: {})
     order = Order.find(id)
     authorize_buyer_request!(order)
-    OrderShippingService.new(order, fulfillment_type: fulfillment_type, shipping: shipping).process!
+    current_pending_offer = order.offers.pending.where(creator_id: context[:current_user][:id]).first if order.mode == Order::OFFER
+    OrderShippingService.new(order, fulfillment_type: fulfillment_type, shipping: shipping, pending_offer: current_pending_offer).process!
```
williardx: 6. `OrderShippingService` now includes an optional `pending_offer` parameter, which is fetched above if it exists. (We shouldn't `set_shipping` on a submitted offer.)

app/services/offers/initial_offer_service.rb
```diff
@@ -12,7 +12,6 @@ def process!
       assert_offer!
       @order.with_lock do
         @offer = @order.offers.create!(amount_cents: @amount_cents, from_id: @user_id, from_type: Order::USER, creator_id: @user_id)
-        @order.update!(last_offer: @offer)
```
williardx: 7. We no longer set the `last_offer` unless the offer is actually submitted.

app/services/order_shipping_service.rb
```diff
@@ -1,41 +1,43 @@
...
+        buyer_phone_number: @buyer_phone_number,
+        shipping_name: @buyer_name,
+        shipping_address_line1: @shipping_address&.street_line1,
+        shipping_address_line2: @shipping_address&.street_line2,
+        shipping_city: @shipping_address&.city,
+        shipping_region: @shipping_address&.region,
+        shipping_country: @shipping_address&.country,
+        shipping_postal_code: @shipping_address&.postal_code
       )
+      @pending_offer.present? ? set_offer_totals! : set_order_totals!
```
williardx: 8. If we have a pending offer then set shipping / tax on the offer. During offer submission we later set the order totals to be the offer totals.

This also currently doesn't set sales tax on the line item or record `should_artsy_remit_taxes`.

app/services/order_shipping_service.rb
```diff
@@ -50,7 +52,7 @@ def tax_total_cents
       @order.line_items.map do |li|
         artwork_address = Address.new(artworks[li.artwork_id][:location])
         begin
-          service = SalesTaxService.new(li, @fulfillment_type, @shipping_address, shipping_total_cents, artwork_address, seller_addresses)
+          service = Tax::CalculatorService.new(li.total_amount_cents, li.effective_price_cents, li.quantity, @fulfillment_type, @shipping_address, shipping_total_cents, artwork_address, seller_addresses)
```
williardx: 9. `CalculatorService` no longer takes a line item and instead takes the eaxct parameters it needs to calculate sales tax.

app/services/order_shipping_service.rb
```diff
@@ -66,49 +68,40 @@ def tax_total_cents
     raise Errors::ValidationError, e.code
   end
 
+  def offer_tax_total_cents
+    seller_addresses = GravityService.fetch_partner_locations(@order.seller_id)
+    artwork_id = @order.line_items.first.artwork_id # this is with assumption of Offer order only having one lineItem
```
williardx: 10. We assume only one line item on an offer order, so we only need to fetch one artwork id to later get the artwork address.

app/services/order_shipping_service.rb
```diff
@@ -66,49 +68,40 @@ def tax_total_cents
...
+    artwork = @artworks[artwork_id]
+    @tax_total_cents ||= begin
+      artwork_address = Address.new(artwork[:location])
+      begin
+        Tax::CalculatorService.new(
+          @pending_offer.amount_cents,
+          @pending_offer.amount_cents / @order.line_items.first.quantity,
```
williardx: 11. This is the offer's effective unit price, needed by TaxJar to be able to calculate tax breakdowns.

app/services/order_shipping_service.rb
```diff
@@ -66,49 +68,40 @@ def tax_total_cents
...
+          @order.fulfillment_type,
+          @order.shipping_address,
+          shipping_total_cents,
+          artwork_address,
+          seller_addresses
+        ).sales_tax
+      rescue Errors::ValidationError => e
+        raise e unless e.code == :no_taxable_addresses
+
+        # If there are no taxable addresses then we set the sales tax to 0.
```
williardx: 12. This doesn't include updating tax attributes on the line item.

app/services/order_shipping_service.rb
```diff
@@ -66,49 +68,40 @@ def tax_total_cents
...
+  rescue Errors::AddressError => e
+    raise Errors::ValidationError, e.code
+  end
+
   def validate_artwork!(artwork)
     raise Errors::ValidationError, :unknown_artwork unless artwork
     raise Errors::ValidationError, :missing_artwork_location if artwork[:location].blank?
   end
 
-  def validate_shipping_location!
```
williardx: 13. All the deleted code here is moved to `ShippingCalculatorService`

app/services/order_shipping_service.rb
```diff
@@ -1,41 +1,43 @@
...
     @shipping_address = Address.new(shipping) if @fulfillment_type == Order::SHIP
     @buyer_name = shipping[:name]
     @buyer_phone_number = shipping[:phone_number]
+    @pending_offer = pending_offer
   end
 
   def process!
     raise Errors::ValidationError.new(:invalid_state, state: @order.state) unless @order.state == Order::PENDING
 
-    validate_shipping_location! if @fulfillment_type == Order::SHIP
```
williardx: 14. This validation was moved into `ShippingCalculatorService` because the related helper methods also all moved into `ShippingCalculatorService` (e.g. determining int'l v. domestic shipping).

app/services/tax/collector_service.rb
```diff
@@ -1,11 +1,8 @@
...
   REMITTING_STATES = [].freeze
```
ashkan18: remove this from here

app/services/offers/submit_order_service.rb
```diff
@@ -8,15 +8,34 @@ def initialize(offer, by: nil)
     end
 
     def process!
-      assert_can_submit!
+      pre_process!
       @order.submit! do
         @offer.update!(submitted_at: Time.now.utc)
+        @order.update!(last_offer: @offer, shipping_total_cents: @offer.shipping_total_cents, tax_total_cents: @offer.tax_total_cents)
+        OrderTotalUpdaterService.new(@order, @partner[:effective_commission_rate]).update_totals!
```
williardx: 15. `OrderTotalUpdaterService` only reflects the state of the order as of the last _submitted_ offer, so it can only happen after the above update to the `order`.

app/services/offers/submit_order_service.rb
```diff
@@ -8,15 +8,34 @@ def initialize(offer, by: nil)
...
       end
       post_process
     end
 
     private
 
+    def pre_process!
```
williardx: 16. These validations all came from elsewhere, mostly `OrderShippingService`.

dleve123:
Is there any way to split this PR into multiple smaller PRs? The diff is huge, and I think it's going to be hard to give this the review it deserves with this large of a diff.

ashkan18:
we could separate the GraphLQ changes but that would be removing small portion of it out. I will be going through this tomorrow in our knowledge share session if that helps. Most of the diff is tests and dividing some services up.

ashfurrow:
Yeah I'm definitely looking forward to the knowledge share on this PR!

app/graphql/mutations/set_shipping.rb
```diff
@@ -10,7 +10,8 @@ class Mutations::SetShipping < Mutations::BaseMutation
   def resolve(id:, fulfillment_type:, shipping: {})
     order = Order.find(id)
     authorize_buyer_request!(order)
-    OrderShippingService.new(order, fulfillment_type: fulfillment_type, shipping: shipping).process!
+    current_pending_offer = order.offers.pending.where(creator_id: context[:current_user][:id]).first if order.mode == Order::OFFER
```
starsirius: Can the same user have multiple pending offers on the same order?

app/graphql/types/order_type.rb
starsirius: Can probably document this in code with a description.

app/graphql/types/order_type.rb
```diff
@@ -82,4 +85,10 @@ def offer_total_cents
     # This can be removed once reaction is updated to `itemsTotalCents`
     object.items_total_cents
   end
+
+  def my_last_offer
+    return unless context[:current_user][:id]
+
+    object.offers.where(creator_id: context[:current_user][:id]).order(created_at: :desc).first
```
starsirius: I'm not sure the exact use case on the partner side, but for partners, this will only return the last offer made by specific user logged in to CMS, right?

app/services/offers/submit_order_service.rb
```diff
@@ -8,15 +8,35 @@ def initialize(offer, by: nil)
...
+      assert_can_submit!
+      raise Errors::ValidationError, :missing_required_info unless @order.can_submit?
```
starsirius: This may not be needed with the `assert_can_submit!` above?

ashkan18:
closing in favor of #286

ashkan18:
bringing this guy back now that [GraphQL part](#286 ) is merged

app/graphql/mutations/set_shipping.rb
sweir27: Yes!

i.e. if a user is going through the checkout flow and then get to the last page, they may `Change` their offer. This creates a new pending offer (which becomes `myLastOffer` so we make sure to show the correct totals)

app/services/order_shipping_service.rb
sweir27: Maybe we should add a comment to explain why we do this.

app/services/order_shipping_service.rb
```diff
@@ -1,41 +1,44 @@
 class OrderShippingService
-  def initialize(order, fulfillment_type:, shipping:)
+  def initialize(order, fulfillment_type:, shipping:, pending_offer: nil)
     @order = order
     @fulfillment_type = fulfillment_type
     @shipping_address = Address.new(shipping) if @fulfillment_type == Order::SHIP
     @buyer_name = shipping[:name]
     @buyer_phone_number = shipping[:phone_number]
+    @pending_offer = pending_offer
```
sweir27: Should we be verifying that the offer is pending? 

Alternatively, can we assume based on whether or not the offer has a `submitted_at` that we should treat it as a pending offer vs. updating the order? If we did that, this could just accept an `@offer`.

app/services/shipping_calculator_service.rb
```diff
@@ -0,0 +1,42 @@
...
+  def domestic_shipping?
+    @artwork[:location][:country].casecmp(@shipping_address.country).zero? &&
+      (@shipping_address.country != Carmen::Country.coded('US').code || @shipping_address.continental_us?)
+  end
+
+  def international_shipping?
+    !domestic_shipping?
+  end
+
+  def calculate_domestic_shipping_fee
```
sweir27: I am trying to think of a better name for this method but am having a hard time 😅 but as it is, it doesn't appear to be calculating anything. Maybe `fetch_domestic_shipping_fee!`? Or can leave as-is. This is #superminor 😄

app/services/shipping_calculator_service.rb
```diff
@@ -0,0 +1,42 @@
...
+    @artwork[:domestic_shipping_fee_cents] || raise(Errors::ValidationError, :missing_domestic_shipping_fee)
+  end
+
+  def calculate_international_shipping_fee
```
sweir27: (ditto to above)

app/services/shipping_calculator_service.rb
williardx: Maybe `@domestic_shipping_fee`?

sweir27: The `!` could be nice because it's also `raise`ing which is a little unexpected.

Although... looking at the code, maybe these don't need to be separate methods at all?

williardx: Yeah, combining this and `calculate_international_shipping_fee` into `shipping_cents` method makes sense to me

app/graphql/mutations/set_shipping.rb
starsirius: So in the scenario where a new pending offer is created, is it ok to get the `current_pending_offer` by `order.offers.pending.first` above in the `set_shipping` mutation?

I wonder if we need to explicitly provide something like offer id to avoid using the wrong offer.

ashkan18: i _think_ it's ok to assume user is setting shipping on their latest offer for now. but  its a good point, since when submitting offer we ask for offer id but for set shipping we don't it may be little confusing when someone has two tabs open on the order, one in review stage another in set shipping. Don't know if we need to tackle that scenario now, if we want we can do a follow up PR, what do you think?

app/graphql/types/order_type.rb
ashkan18: yes, think of the case that two people from the same partner are looking at their pending offers on the same order, only one of them can submit the offer but when looking at pending offers they see only their own current one. It seems this is less of an issue in CMS since looks like we don't create pending offers and then submit, and instead we submit counter offer in one shot.

app/services/offers/submit_order_service.rb
ashkan18: good catch! 👍

app/services/order_shipping_service.rb
ashkan18: makes sense to add checking for _pending_. If the offer is submitted already i don't think we can update it anymore, right?

app/graphql/mutations/set_shipping.rb
dleve123: I would recommend `explicitly` order-ing the ActiveRecord::Relation being created, effectively acknowledging that there could be many offers and defining how we're going to handle them

starsirius: 👍 to iterate in a followup!

It's hard to predict where we will introduce race condition down the road, so I was thinking making it explicit. Also, does `order.offers.pending.first` return the latest offer? I'm always confused by the ordering and it's based on uuid. 😅

ashkan18:
thanks all for reviews! lets do this!

PR #284, opened 2018-11-16
Use total sales tax amount when calculating sales tax

williardx:
### Problem
Partially reverts https://github.com/artsy/exchange/pull/239

Per legal's request, we're now going to be charging the _total_ collectable sales tax inclusive of state, city, county and other special jurisdiction taxes when calculating sales tax. Previously, we were only charging state sales tax.

### Solution
Pass along whatever total tax amount we obtain in TaxJar's `amount_to_collect` parameter

### What Changed
- Stop using breakdown object in sales tax parameter to selectively pull out state sales tax and instead use `amount_to_collect` parameter.

### DIFF
--- a/app/services/sales_tax_service.rb
+++ b/app/services/sales_tax_service.rb
@@ -26,11 +26,7 @@ class SalesTaxService
   end
 
   def sales_tax
-    @sales_tax ||= begin
-      tax_response = fetch_sales_tax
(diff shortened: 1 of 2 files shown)


joeyAghion:
This seems simpler and better in every way.

Curious about the policy-thrashing though. Someday you can tell me about that.

ashkan18:
Nice!! I'll reflect this in #283

PR #286, opened 2018-11-19
Add my last offer and waiting for seller/buyer responses

ashkan18:
GraphQL changes part of #283


app/graphql/types/order_type.rb
```diff
@@ -33,10 +33,13 @@ class Types::OrderType < Types::BaseObject
   end
   field :total_list_price_cents, Integer, null: false
   field :offer_total_cents, Integer, null: false, deprecation_reason: 'itemsTotalCents reflects offer total for offer orders.'
-  field :last_offer, Types::OfferType, null: true
+  field :last_offer, Types::OfferType, null: true, description: 'Last submitted offer'
+  field :my_last_offer, Types::OfferType, null: true
   field :tax_total_cents, Integer, null: true
   field :transaction_fee_cents, Integer, null: true, seller_only: true
   field :updated_at, Types::DateTimeType, null: false
+  field :waiting_buyer_response, Boolean, null: true
```
ashkan18: `waitingResponseFrom` -> `Enum`

spec/controllers/api/requests/add_initial_offer_to_order_mutation_request_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+          expect(response.data.add_initial_offer_to_order.order_or_error.error.type).to eq('validation')
+          expect(response.data.add_initial_offer_to_order.order_or_error.error.code).to eq('invalid_amount_cents')
+        end
+
+        it 'returns the order' do
+          response = client.execute(mutation, input: mutation_input)
+          response_order = response.data.add_initial_offer_to_order.order_or_error.order
+          expect(response_order.id).to eq(order_id)
+          expect(response_order.total_list_price_cents).to eq 400
+          expect(response_order.last_offer).not_to be_nil
```
sweir27: Shouldn't `last_offer` be `nil` here since it's not submitted?

ashkan18: this PR still sets the `last_offer` on `initial` the service layer PR would remove it

app/graphql/types/order_type.rb
```diff
@@ -33,10 +33,12 @@ class Types::OrderType < Types::BaseObject
   end
```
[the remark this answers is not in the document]
ashkan18: good question... i'll take it out and we can add it if there is a need, it seems it will more cause leak other than being helpful!

spec/controllers/api/requests/order_my_offer_request_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+        result = client.execute(query, id: order.id)
+        expect(result.data.order.my_last_offer).to be_nil
+      end
+      context 'with offers' do
+        before do
+          buyer_offer1
+          buyer_offer2
+          seller_offer
+        end
+        it 'returns myLastOffer' do
```
dleve123: Thoughts on adding a test case demonstrating that myLastOffer will return pending, in addition to submitted offers? I think doing so would better document this code with tests.

spec/controllers/api/requests/order_my_offer_request_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+      let(:jwt_user_id) { nil }
+      let(:jwt_roles) { 'trusted' }
+      before do
+        buyer_offer1
+        buyer_offer2
+        seller_offer
+      end
+      context "trusted account accessing another account's order" do
+        let(:buyer_id) { 'some-user' }
+        it 'returns expected payload' do
```
dleve123: thoughts on rewording the docstring to `it "it returns nil`" to describe what the expectation is?

spec/controllers/api/requests/order_my_offer_request_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+      context 'with offers' do
+        before do
+          buyer_offer1
+          buyer_offer2
+          seller_offer
+        end
+        it 'returns myLastOffer' do
+          result = client.execute(query, id: order.id)
+          expect(result.data.order.offers.edges.count).to eq 2
+          expect(result.data.order.offers.edges.map(&:node).map(&:id)).to match_array [buyer_offer1.id, seller_offer.id]
```
dleve123: Thoughts on moving L70 and L69 into a separate spec? You are testing different logic than the `myLastOffer` logic, and I think a dedicated docstring with something like "offer returns all (buyer and seller) submitted offers" would be helpful documentation.

spec/controllers/api/requests/order_query_request_spec.rb
```diff
@@ -224,54 +226,62 @@
       end
 
       context 'with offers' do
+        let(:state) { Order::SUBMITTED }
         let(:order_mode) { Order::OFFER }
-        let!(:offer1) { Fabricate(:offer, order: user1_order1, amount_cents: 200, from_id: user_id, from_type: Order::USER, submitted_at: Date.new(2018, 1, 1)) }
-        let!(:offer2) { Fabricate(:offer, order: user1_order1, amount_cents: 300, from_id: partner_id, from_type: 'gallery', responds_to_id: offer1.id) }
+        let!(:buyer_offer) { Fabricate(:offer, order: user1_order1, amount_cents: 200, from_id: user_id, from_type: Order::USER, submitted_at: Date.new(2018, 1, 1)) }
+        let!(:seller_offer) { Fabricate(:offer, order: user1_order1, amount_cents: 300, from_id: partner_id, from_type: 'gallery', responds_to_id: buyer_offer.id, submitted_at: Date.new(2018, 1, 2)) }
```
dleve123: similar question to one earlier in the PR review - should this be referencing a constant?

spec/controllers/api/requests/order_query_request_spec.rb
```diff
@@ -224,54 +226,62 @@
...
         it 'includes last_offer' do
           result = client.execute(query, id: user1_order1.id)
-          expect(result.data.order.last_offer.id).to eq offer2.id
+          expect(result.data.order.last_offer.id).to eq seller_offer.id
           expect(result.data.order.last_offer.from.id).to eq partner_id
           expect(result.data.order.last_offer.from.__typename).to eq 'Partner'
-          expect(result.data.order.last_offer.responds_to.id).to eq offer1.id
+          expect(result.data.order.last_offer.responds_to.id).to eq buyer_offer.id
+        end
+        context 'last offer from seller' do
```
dleve123: thoughts on adding a test cases for (1) when last order is not a submitted and offer-mode and (2) when there isn't a last offer?

spec/controllers/api/requests/submit_order_with_offer_mutation_request_spec.rb
```diff
@@ -62,7 +62,7 @@
     before do
       order.line_items << line_item
       Offers::InitialOfferService.new(order, 800_00, user_id).process!
-      @offer = order.reload.last_offer
+      @offer = order.reload.offers.last
```
dleve123: can you briefly describe why we are making this change?

app/graphql/types/order_type.rb
```diff
@@ -61,6 +62,16 @@ def requested_fulfillment
     object
   end
 
+  def waiting_response_from
+    return unless object.mode == Order::OFFER && object.state == Order::SUBMITTED
+
+    if object&.last_offer&.from_id == object.seller_id && object&.last_offer&.from_type == object.seller_type
```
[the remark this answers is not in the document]
sweir27: This also feels like logic that could be extracted into a method on the `Offer` model.

_schema.graphql
```diff
@@ -330,6 +351,7 @@ type Order {
   totalListPriceCents: Int!
   transactionFeeCents: Int
   updatedAt: DateTime!
+  waitingResponseFrom: OrderParticipantEnum
```
joeyAghion: Should this be `awaitingResponseFrom` instead?

Also, it's unclear to me how this would behave for an offer that's already been countered/expired/accepted/rejected. Would this return a null value? Would it continue to refer to the participant it used to await? I'd find it more clear if we were to expose the kind of participant an offer _was from_ for that reason (because that's stable data).

app/graphql/types/order_type.rb
```diff
@@ -29,14 +29,15 @@ class Types::OrderType < Types::BaseObject
   field :offers, Types::OfferType.connection_type, null: true do
     argument :from_id, String, required: false
     argument :from_type, String, required: false
-    argument :include_pending, Boolean, required: false
   end
   field :total_list_price_cents, Integer, null: false
   field :offer_total_cents, Integer, null: false, deprecation_reason: 'itemsTotalCents reflects offer total for offer orders.'
-  field :last_offer, Types::OfferType, null: true
+  field :last_offer, Types::OfferType, null: true, description: 'Last submitted offer'
+  field :my_last_offer, Types::OfferType, null: true
```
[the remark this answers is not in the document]
joeyAghion: +1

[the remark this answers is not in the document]
starsirius: maybe worth calling out it's regardless of state.

spec/controllers/api/requests/order_my_offer_request_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+        commission_rate: 0.10,
+        seller_total_cents: 50_00,
+        buyer_total_cents: 100_00,
+        items_total_cents: 0,
+        state: Order::SUBMITTED
+      )
+    end
+    let(:buyer_offer1) { Fabricate(:offer, order: order, amount_cents: 200, from_id: buyer_id, from_type: Order::USER, creator_id: buyer_id, submitted_at: 2.days.ago, created_at: 2.days.ago) }
+    let(:buyer_offer2) { Fabricate(:offer, order: order, amount_cents: 300, from_id: buyer_id, from_type: Order::USER, creator_id: buyer_id, created_at: 1.day.ago) }
+    let(:seller_offer) { Fabricate(:offer, order: order, amount_cents: 200, from_id: buyer_id, from_type: 'gallery', creator_id: seller_id, submitted_at: Date.new(2018, 1, 2)) }
```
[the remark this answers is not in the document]
ashkan18: we can remove `Order::PARTNER` we used to limit `from_type`, `seller_type`, `buyer_type` to be either `user` or `partner` but now `partner` is expanded to the actual partner type, meaning it can be `gallery` , `institution` and all cases from https://github.com/artsy/gravity/blob/master/app/models/domain/partner.rb#L14

spec/controllers/api/requests/order_my_offer_request_spec.rb
ashkan18: its tested on L71, maybe not that obvious, made it its own spec

spec/controllers/api/requests/submit_order_with_offer_mutation_request_spec.rb
ashkan18: this was copied over from upcoming PR, should not matter in this case and will be switched to this in next PR anyways.

_schema.graphql
ashkan18: i _think_ this field should only return the type if we are actually waiting on someone, on accepted/rejected order we should return `nil` meaning we are not waiting for anyone. would that make sense?

app/graphql/types/order_type.rb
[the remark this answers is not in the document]
ashkan18: i originally had a separate check for `last_offer` existence` and our cop friend wasn't happy about the complexity. I'll add spec about `nil` case to clarify it, but basically we want to return `nil` in any other cases than those two if statements above since in any other cases we are not waiting for anyone's response.

Re offer model, current `waiting_response_from` is purely a graphQL helper thingy and on model layer we don't have that notion anywhere, yet. I'd suggest when this logic was needed in service layer and etc, we can refactor and move it there.

williardx:
@ashkan18 This all looks good to me, feel free to merge whenever you're ready 🙌

ashkan18:
cool! merging so we can try on staging during 🦃 time

spec/controllers/api/requests/add_initial_offer_to_order_mutation_request_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+        context "order in #{state}" do
+          let(:state) { state }
+          it 'returns error' do
+            response = client.execute(mutation, input: { orderId: order_id, amountCents: 100 })
+            expect(response.data.add_initial_offer_to_order.order_or_error.error.type).to eq('validation')
+            expect(response.data.add_initial_offer_to_order.order_or_error.error.code).to eq('cant_offer')
+          end
+        end
+      end
+      context 'Canceled order' do
```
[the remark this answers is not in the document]
dleve123: Got it - I misread the code -- didn't see the `reject`!

PR #288, opened 2018-11-21
WIP reject offer

sepans:
reject offer


app/graphql/mutations/seller_reject_offer.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure', null: false
+
+  def resolve(offer_id:)
+    offer = Offer.find(offer_id)
+    order = offer.order
+
+    authorize_seller_request!(order)
+
+    Offers::RejectOfferService.new(offer: offer, order: order).process!
```
ashkan18: i think we can just pass `offer` to this service, I agree with your earlier point about a helper for authorizing the request based on the offer. In the end we may do one extra fetch of order but i think its lazy loaded and should be ok. what do you think?

sepans:
Made the 👆 change and tested it locally. Should be ready to merge 🙏

app/services/offers/reject_offer_service.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+
+    def process!
+      @offer.order.reject!
+      instrument_offer_reject
+    end
+
+    private
+
+    attr_reader :offer
```
starsirius: #minor. Looks like this is not used?

spec/services/offers/reject_offer_service_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        dd_statsd = stub_ddstatsd_instance
+        allow(dd_statsd).to receive(:increment).with('offer.reject')
+
+        service.process!
+
+        expect(dd_statsd).to have_received(:increment).with('offer.reject')
+      end
+    end
+
+    context "when we can't approve the order" do
```
starsirius: #minor typo: "reject"

app/services/offers/reject_offer_service.rb
```diff
@@ -0,0 +1,20 @@
+module Offers
+  class RejectOfferService
+    def initialize(offer:)
+      @offer = offer
+    end
+
+    def process!
+      @offer.order.reject!
```
starsirius: Chatted offline with @sepans, and we might need some validation to avoid blinding accepting an offer id and reject the order.

PR #289, opened 2018-11-22
Order interface

ashkan18:
This is the Exchange follow up of the same thing we did in https://github.com/artsy/metaphysics/pull/1386, its not breaking change (and added spec to verify that) since all the current offer related fields are still in `OderInterface` but deprecated.


ashkan18:
I did few tests locally and things look ok, @orta going to merge this so i can try it without MP changes on staging.

PR #290, opened 2018-11-25
Expose transactionTotalCents and shippingTotalCents in offerType

ashkan18:
Follow up on #286 to add `shippingTotalCents` and `taxTotalCents` on `offer` type.


ashkan18:
merge on green

PR #291, opened 2018-11-26
Remove _deprecated_ offerTotal

ashkan18:
This field is not used in Reaction anymore, in MP we [already switched it to use `itemsTotalCents` ](https://github.com/artsy/metaphysics/blob/master/src/schema/ecommerce/types/order.ts#L75), we are now good to remove it from Exchange.


sweir27:
👍 cool! Thanks for following up here.

PR #292, opened 2018-11-26
Howto add MP operation guide

sepans:
Links to an annotated guide on how to add MP operations. I didn't add it to the doc since this is temporary until we implement stitching.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,9 @@ rake graphql:schema:idl
 ```shell
 mv schema.graphql exchange.graphql
 ```
-3) copy file above to your local update Metaphysics under `src/data` and make a PR to Metaphysics with this change
+3) copy file above to your local update Metaphysics under `src/data` and make a PR to Metaphysics with this change. 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,7 +67,9 @@ rake graphql:schema:idl
 ```shell
 mv schema.graphql exchange.graphql
 ```
-3) copy file above to your local update Metaphysics under `src/data` and make a PR to Metaphysics with this change
+3) copy file above to your local update Metaphysics under `src/data` and make a PR to Metaphysics with this change. 
+
+There is a guide on how to add operations to Metaphysics [here](https://gist.github.com/sepans/56e13c4084dcdd38fd64812d61b89bdb)
```
ashkan18: Nice! this makes lot of sense! do you think we can have the gist somewhere in MP's docs?

sepans: Yeah I thought this doc is going to be short lived and gets removed after stitching is turned on (🤞). Open to move it anywhere else if it makes more sense to be there 👍

ashkan18: sorry for late response, how about adding it here -> https://github.com/artsy/metaphysics/tree/master/docs as something like `update_exchange_schema`

sepans: done!

PR #293, opened 2018-11-27
Only run coveralls when on CI

dleve123:
This commit means that coveralls is only run on CI and no other
environment.

* Circle CI [sets the CI ENV variable][0]
* Cleans up stdout when running `rspec`
  + also, loading less code is almost always a good thing!

[0]:https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables


ashkan18:
👍

PR #294, opened 2018-11-27
Sellers provide offer rejection reason

sepans:
Fixes [SELL-1188](https://artsyproduct.atlassian.net/browse/SELL-1188)

In CMS partners can provide rejection reason for offers:

screen shot 2018-11-27 at 3 52 15 pm

This PR follows up https://github.com/artsy/exchange/pull/288 to add `rejection_reason` to `seller_reject_offer_mutation`


app/services/offers/reject_offer_service.rb
```diff
@@ -1,11 +1,14 @@
...
-    def initialize(offer:)
+  class RejectOfferService < BaseOfferService
+    def initialize(offer:, reject_reason:)
       @offer = offer
+      @reject_reason = reject_reason
     end
 
     def process!
-      @offer.order.reject!
+      validate_is_last_offer!
```
starsirius: We can also validate if the offer is from the right participant, e.g. in this case, buyer.

app/services/offers/reject_offer_service.rb
```diff
@@ -1,11 +1,14 @@
...
+
+      @offer.order.reject!(@reject_reason)
```
starsirius: #minor. We might want to revisit the [default reject reason](https://github.com/artsy/exchange/blob/ae0f9b4dabb9998ecb1c1dbba9ffef57ccc465d7/app/models/order.rb#L51), which is "seller rejected" now, given both seller and buyer can reject an offer.

ashkan18: yes I think we can migrate existing `seller_rejected` ones to be `seller_reject_other` what do you think?

sepans:
Also discussing it offline in the meeting we also need `buyer_lapsed` and `buyer_rejected` for offers

app/services/offers/reject_offer_service.rb
sepans: running ` Order.where(state_reason: 'seller_rejected').count` there are 2 rejected orders on staging and 3 on production. I can migrate them manually.

sepans: running ` Order.where(state_reason: 'seller_rejected').count` there are 2 rejected orders on staging and 3 on production. I can migrate them manually.

ashkan18: awesome! should be ok! 🙌 we can also ping @artsy/analytics in case they use that word anywhere

app/services/offers/accept_service.rb
```diff
@@ -1,12 +1,13 @@
...
-  class AcceptService
+  class AcceptService < BaseOfferService
     def initialize(offer:, order:)
       @offer = offer
       @order = order
     end
 
     def process!
       validate_is_last_offer!
+      validate_offer_is_from_buyer!
```
starsirius: I imagine we will expand this if we want to reuse this `AcceptService` when buyer accepts an offer. 👍

spec/controllers/api/requests/seller_reject_offer_mutation_request_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+      {
+        input: {
+          offerId: offer.id.to_s,
+          rejectReason: reject_reason
+        }
+      }
+    end
+
+    before do
+      order.update!(last_offer: offer)
```
starsirius: #minor. This can probably be moved to the fabrication above?

app/services/offers/accept_service.rb
sepans: Yes for counter-offer I have added an extra parameter (`counter_from_type`) to the service so it could be used by two sides. Will do the same here in the incoming PR.

spec/controllers/api/requests/seller_reject_offer_mutation_request_spec.rb
sepans: hmmm, I copied it from `seller_accept_offer_mutation_request_spec.rb` so it makes sense to move it to fabricator. Will do it for all of them in the counter offer PR. 👍

ashkan18:
lets do this 🙌

PR #295, opened 2018-11-27
Notify when offer is approved

dleve123:
Do not merge as I'm still doing some manual integration testing, but I expect that this PR is complete

TODO:

- [x] Update transaction event to not include line items total
- [x] Update OrdersEvent to include line items total

https://artsyproduct.atlassian.net/browse/PURCHASE-660

### DIFF
--- a/app/graphql/mutations/approve_order.rb
+++ b/app/graphql/mutations/approve_order.rb
@@ -8,11 +8,17 @@ class Mutations::ApproveOrder < Mutations::BaseMutation
   def resolve(id:)
     order = Order.find(id)
     authorize_seller_request!(order)
-    OrderApproveService.new(order, context[:current_user]['id']).process!
+    OrderApproveService.new(order, current_user_id).process!
     {
       order_or_error: { order: order.reload }
     }
   rescue Errors::ApplicationError => e
     { order_or_error: { error: Types::ApplicationErrorType.from_application(e) } }
+  end
+
+  private
+
+  def current_user_id
+    context[:current_user]['id']
   end
 end
--- a/app/graphql/mutations/approve_order.rb
+++ b/app/graphql/mutations/approve_order.rb
@@ -15,10 +15,4 @@ class Mutations::ApproveOrder < Mutations::BaseMutation
   rescue Errors::ApplicationError => e
     { order_or_error: { error: Types::ApplicationErrorType.from_application(e) } }
   end
-
-  private
-
-  def current_user_id
-    context[:current_user]['id']
-  end
 end
--- a/app/graphql/mutations/seller_accept_offer.rb
+++ b/app/graphql/mutations/seller_accept_offer.rb
@@ -21,10 +21,4 @@ class Mutations::SellerAcceptOffer < Mutations::BaseMutation
   rescue Errors::ApplicationError => e
     { order_or_error: { error: Types::ApplicationErrorType.from_application(e) } }
   end
-
-  private
-
-  def current_user_id
-    context[:current_user]['id']
-  end
 end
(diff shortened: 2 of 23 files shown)


dleve123:
merge on green

app/graphql/mutations/seller_accept_offer.rb
```diff
@@ -11,10 +11,20 @@ def resolve(offer_id:)
 
     authorize_seller_request!(order)
 
-    Offers::AcceptService.new(offer: offer, order: order).process!
+    Offers::AcceptService.new(
+      offer: offer,
+      order: order,
+      user_id: current_user_id
```
ashkan18: #minor to be consistent with params passed in other services, we can rename `user_id` to `by` here.

app/jobs/post_order_notification_job.rb
```diff
@@ -1,4 +1,4 @@
-class PostNotificationJob < ApplicationJob
+class PostOrderNotificationJob < ApplicationJob
```
ashkan18: 👍 BUT we can't just rename this like this since there are already jobs in the queue pointing to old name, the safe way to do this is to create new job, keep the old one around (by possibly an empty class just inheriting this new one), deploy this PR, make sure all existing jobs pointing to old job are processed, in follow up PR remove this job.

app/services/offers/accept_service.rb
```diff
@@ -1,21 +1,28 @@
 module Offers
   class AcceptService < BaseOfferService
-    def initialize(offer:, order:)
+    def initialize(offer:, order:, user_id:)
```
ashkan18: as mentioned above, `by` would be more consistent with other services.

ashkan18:
make sure to fix the `PostNotification` rename issue before merge 👍

app/graphql/mutations/seller_accept_offer.rb
dleve123: I was thinking that `user_id` is a bit more clear than `by` – what do you think about renaming `by` -> `user_id` across the board? As a newish developer on the codebase, I found `by` a bit confusing. Would you be OK with that?

If not, I agree that consistency is more important, and will rename back to `by`

ashkan18: `user_id` works for me too, just need to update other services to use it and make sure those services are not directly queued in `sidekiq` (im pretty sure they are not, but worth checking)

app/jobs/post_order_notification_job.rb
dleve123: Oh interesting, nice catch. Yeah, I'll implement the approach you suggested before merging.

app/graphql/mutations/approve_order.rb
```diff
@@ -8,11 +8,17 @@ class Mutations::ApproveOrder < Mutations::BaseMutation
...
     {
       order_or_error: { order: order.reload }
     }
   rescue Errors::ApplicationError => e
     { order_or_error: { error: Types::ApplicationErrorType.from_application(e) } }
   end
+
+  private
+
+  def current_user_id
```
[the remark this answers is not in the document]
dleve123: Interestingly, the test suite passes when all methods in BaseMutation are private, so I opted for that approach! See https://github.com/artsy/exchange/pull/295/commits/aa4676f2547366c81353d71fe2fa48a685df99d4 for more info.

app/jobs/post_order_notification_job.rb
dleve123: @ashkan18 I've implemented this alias a rebased commit: https://github.com/artsy/exchange/pull/295/commits/aa470960628e0cc46636b3f947cd2c9f3152fe06

Can you confirm that this we're aligned here?

app/services/offers/accept_service.rb
dleve123: See https://github.com/artsy/exchange/pull/304

ashkan18:
👍 lets do this

PR #296, opened 2018-11-28
Remove deprecated initialOffer

ashkan18:
# Change
After https://github.com/artsy/metaphysics/pull/1392 we are not using `initialOffer` mutation anymore.


sweir27:
Looks like this is still being used in reaction, so we'll need to remove it from [there](https://github.com/artsy/reaction/blob/master/src/Apps/Order/Routes/Offer/index.tsx#L60) first.

ashkan18:
oh i missed this, i thought we updated MP's `initialOffer` to `addInitialOffer` here https://github.com/artsy/metaphysics/pull/1392/files#diff-64c859093a3dc429f7f188f6ea906411R33

ashkan18:
closing for now, till we update reaction.

ashkan18:
waiting for https://github.com/artsy/reaction/pull/1605

PR #298, opened 2018-11-29
Add Procfile.dev to ease local development

dleve123:
Sarah and I were QA'ing an event sent from a not-yet-deployed version of
exchange to pulse. At some point, we realized that an event wasn't sent
to RabbitMQ because my local sidekiq wasn't running.

* Add a Procfile.dev to make it easier to remember the
system-level dependencies needed to this service.
* Document usage of Procfile.dev in README
* Add `gem install foreman` to bin/setup

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,0 +1,2 @@
+web: bundle exec rails server -p 3000
+worker: bundler exec sidekiq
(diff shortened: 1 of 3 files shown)


sweir27:
🙌 thank you!

sweir27:
🙌 thank you!

PR #299, opened 2018-11-29
Fix Papertrail versions issue

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PLATFORM-1058

# TLDR;
Since we switched from integer primary keys for models to `uuid`, papertrail has been silently failing storing correct data by trying to insert our `uuid` string field into `integer` field and basically resulting in wrong data in `item_id` field.

# Quick Solution
Papertrail was mainly added as an audit trail to Exchange since we are storing some important data and we do want to track changes on our most important models. Having said that, no one is currently actively using this data and there hasn't been a need to look at them yet. The quickest solution here is to:
- [ ] fix the issue by changing `item_id` to be `string`, this should fix the issue for all new changes to models
- [ ] Archive the table and store in S3 possibly in `artsy-backup/db`, the archived data is still valuable if we ever wanted to go back to papertrail data for older records since they have a the actual `jsonb` data
- [ ] Delete older records with wrong `item_id`

### DIFF
--- a/db/migrate/20181129132032_fix_versions.rb
+++ b/db/migrate/20181129132032_fix_versions.rb
@@ -1,0 +1,5 @@
+class FixVersions < ActiveRecord::Migration[5.2]
+  def change
+    change_column :versions, :item_id, :string
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_11_13_234322) do
+ActiveRecord::Schema.define(version: 2018_11_29_132032) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
@@ -160,7 +160,7 @@ ActiveRecord::Schema.define(version: 2018_11_13_234322) do
 
   create_table "versions", force: :cascade do |t|
     t.string "item_type", null: false
-    t.integer "item_id", null: false
+    t.string "item_id", null: false
     t.string "event", null: false
     t.string "whodunnit"
     t.jsonb "object"


db/migrate/20181129132032_fix_versions.rb
```diff
@@ -0,0 +1,5 @@
+class FixVersions < ActiveRecord::Migration[5.2]
+  def change
+    change_column :versions, :item_id, :string
```
dleve123: While we’re here migrating the versions table, I think we should add the object_versions column to the versions table - in that column paper trail will directly persist the diff between 2 versions, easing developer UX.

dleve123: Also, I feel a bit uneasy about allowing improper versions (pre migration) to stay in this table.

Considering we’re on RDS, which includes nightly backups, thoughts on first deleting all existing versions, and then migrating?

If we ever decide to expose the versions table in the UI or to Looker via Fulcrum, I think we’ll have fewer edge cases to protect against if we perform this data cleanup now.

ashkan18: RE `object_versions`, i would make that a separate ticket and track it separately, i like to keep this PR scoped to the original problem and then in separate PR handle new feature. On a more generic point, i don't think we need UX in front of this, at least at the moment. We can just use sql to verify data if needed, if we want we can always do these things on Looker and handle the diff in sql if needed.

RE delete older records, i do have it as part of migrations in the steps above, but im also fine with keeping the data there too since it doesn't impact the existing data anyways. as for backups, it will be backed up by RDS nightly job but when a migration involves deleting data and we think there is a good reason to keep the backup as a separate thing we also keep that snapshot in S3 for possible later use.

williardx:
Looks good to me!  Feel free to merge when you're ready.

db/migrate/20181129132032_fix_versions.rb
dleve123: > On a more generic point, i don't think we need UX in front of this, at least at the moment. We can just use sql to verify data if needed, if we want we can always do these things on Looker and handle the diff in sql if needed.

I don't think I was very clear with UX. If we added the `object_changes` column, answering the question of "what was the most recent change to this record" is as simple as an `order.versions.first.changeset`. Without the column, we would have to manually diff in the console. In my experience, having the ability to call `#changeset` gives us the most value out of the gem.

I'm all for isolating this into a separate ticket though. I created it here: https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=34&projectKey=BUGS&view=detail&selectedIssue=BUGS-639

ashkan18:
i actually thought about switching to `uuid` but thought having it as `string` would be more generic. I'm happy to change it but also could live with `string`, from _papertrail_ perspective that totally worked.
As I have it in the PR description, for migration, im planning to:
- [ ] Run db migration
- [ ] Copy the snapshot of the database to S3 for any possible future references
- [ ] Delete older data

PR #300, opened 2018-11-29
Add seller counter offer (and services)

sepans:
Fixes [SELL-1176](https://artsyproduct.atlassian.net/browse/SELL-1176)

* Adds seller side mutation to counter an offer ~(`SellerCounterOffer(offer_id, amount_cents`)~ (`SellerCounterOfferOnOrder(offer_id, amount_cents)`).
* Adds services needed to create counter offer from both seller and buyer side:
  - ~`InitialCounterOfferService`~`AddPendingCounterOfferService` to create a pending offer.
  - `SubmitCounterOfferService` to submit the pending offer.
     Seller side mutation uses both to create and submit an order but buyer side does that in two separate steps
* Moves the methods being used to calculate offer tax and shipping from `OrderShippingService` to `BaseTotalCalculatorService` and `OrderShippingService` inherits those methods from that.
* Adds `OfferTotalUpdaterService` (also subclass of `BaseTotalCalculatorService`) which being used by `InitialCounterOfferService` and to calculate offer shipping and tax.

With these services setup, the buyer side counter offer mutations should be very straight forward. Two mutations one for creating pending offer and one for submitting it using `InitialCounterOfferService` and `SubmitCounterOfferService`. (follow up PR)

### Selfie:
#300th PR special edition counter offer

counter-offer


app/graphql/mutations/seller_counter_offer.rb
```diff
@@ -0,0 +1,23 @@
+class Mutations::SellerCounterOffer < Mutations::BaseMutation
```
ashkan18: one thing we've started doing is to add `Order` somewhere in the name of mutation to make it more obvious that the result of this mutation will be an `OrderOrFailure...` what about renaming this to `SellerCounterOfferOnOrder` ?

app/graphql/mutations/seller_counter_offer.rb
```diff
@@ -0,0 +1,23 @@
...
+  argument :offer_id, ID, required: true
+  argument :amount_cents, Integer, required: true
+
+  field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure', null: false
+
+  def resolve(offer_id:, amount_cents:)
+    offer = Offer.find(offer_id)
+    order = offer.order
+    authorize_seller_request!(order)
+    init_service = Offers::InitialCounterOfferService.new(offer: offer, amount_cents: amount_cents)
```
ashkan18: should it just be `Offers::CounterOfferService`?

ashkan18: I see below, I think we can call this one `Offers::AddPendingCounterOffer` which returns offer. Then maybe we can rename the service below to be `Offers::SubmitOffer` since it can be generic and used for any types of offer submission, it will be a lite version of `Offers::SubmitOrderService`.

app/services/offers/initial_counter_offer_service.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      @pending_offer
+    end
+
+    private
+
+    attr_reader :offer
+
+    def instrument_offer_counter
+      Exchange.dogstatsd.increment 'offer.counter'
```
ashkan18: we may want to wait for offer submission to send this, this may be too early if offer never gets submitted.

app/services/offers/initial_counter_offer_service.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    def process!
+      validate_is_last_offer!
+      validate_offer_is_from_buyer! # TODO: generalize to offer_from_other_side
+      validate_order_is_submitted!
+
+      @pending_offer = @order.offers.create!(
+        amount_cents: @amount_cents,
+        from_id: @user_id,
```
ashkan18: where do we set `@from_id` and `@from_type`? should we add these to initializer?

app/services/offers/initial_counter_offer_service.rb
sepans: Makes sense. I thought that data might be useful to see the ratio between pending_counter_offers/submitted_counter_offers (e.g. how many times buyers go back adjust their counter offer) but that data could be extracted from exchange db as well in case it was needed.

app/services/offers/initial_counter_offer_service.rb
sepans: Good catch. Will also add specs for it.

app/services/order_shipping_service.rb
```diff
@@ -71,42 +66,4 @@ def tax_total_cents
   rescue Errors::AddressError => e
     raise Errors::ValidationError, e.code
   end
-
-  def set_offer_tax_total_cents
```
jonallured: Nice extraction! ✂️

app/services/offers/offer_validation_service.rb
```diff
@@ -7,5 +7,17 @@ def validate_is_last_offer!(offer)
...
+
+    def validate_offer_order_is_submitted!(offer)
+      raise Errors::ValidationError.new(:invalid_state, offer) unless offer.order.state == Order::SUBMITTED
+    end
+
+    def validate_offer_not_submitted!(offer)
+      raise Errors::ValidationError.new(:invalid_state, offer) unless offer.submitted_at.nil?
+    end
+
+    def validate_offer_is_mine(offer, from_id)
```
williardx: Looks like you're missing a `!` here?

app/services/offers/add_pending_counter_offer_service.rb
```diff
@@ -0,0 +1,34 @@
...
+      )
+      totals_service = OfferTotalUpdaterService.new(offer: @pending_offer)
+      totals_service.process!
+
+      @pending_offer
+    end
+
+    private
+
+    attr_reader :offer
```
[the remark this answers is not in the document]
sepans: That line I copied from the PR we did with @dleve123 so the credit goes to you guys 🥇

app/graphql/mutations/offer/seller_counter_offer_on_order.rb
```diff
@@ -0,0 +1,24 @@
...
+  argument :amount_cents, Integer, required: true
+
+  field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure', null: false
+
+  def resolve(offer_id:, amount_cents:)
+    offer = Offer.find(offer_id)
+    order = offer.order
+    from_id = context[:current_user][:id]
+    authorize_seller_request!(order)
+    init_service = Offers::AddPendingCounterOfferService.new(offer: offer, amount_cents: amount_cents, from_type: Order::PARTNER, from_id: from_id)
```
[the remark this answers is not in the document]
sepans: 👍  The variable name is from when the service was called `InitialCounterOfferService`

app/services/order_shipping_service.rb
```diff
@@ -1,4 +1,4 @@
-class OrderShippingService
+class OrderShippingService < BaseTotalCalculatorService
```
williardx: It's not super intuitive to me that `OrderShippingService` inherits from `BaseTotalCalculatorService` since `OrderShippingService` is also updating shipping details. What if we shifted the order tax and shipping calculations to a separate service? Or maybe used `OrderTotalUpdaterService` instead, which could inherit from `BaseTotalCalculatorService`?

app/services/order_shipping_service.rb
```diff
@@ -37,11 +37,6 @@ def set_order_totals!
     @order.update!(shipping_total_cents: shipping_total_cents, tax_total_cents: tax_total_cents)
   end
 
-  def set_offer_totals!
-    @pending_offer.update!(shipping_total_cents: shipping_total_cents)
-    set_offer_tax_total_cents
-  end
-
   def artworks
```
williardx: This method looks to be the same as the `artworks` method in the parent class

app/services/order_shipping_service.rb
williardx: Or `BaseTotalCalculatorService` could be a module instead

sepans: 👍 Thats a great idea. Actually I thought about doing something like that and started a spike on it but the changes propagated and I wasn't familiar enough with the logic to be confident to go down that path. Definitely needed as a follow up PR.

app/services/order_shipping_service.rb
sepans: ty ty

app/controllers/errors/error_types.rb
```diff
@@ -38,6 +38,7 @@ module Errors
       not_offerable
       offer_more_than_one_line_item
       offer_not_from_buyer
+      offer_not_mine
```
ashkan18: i wonder if we should use `not_offerable` here, this error seems very specific.

_schema.graphql
```diff
@@ -343,6 +343,7 @@ type Mutation {
   initialOffer(input: InitialOfferInput!): InitialOfferPayload @deprecated(reason: "Switch to addInitialOfferToOrder.")
   rejectOrder(input: RejectOrderInput!): RejectOrderPayload
   sellerAcceptOffer(input: SellerAcceptOfferInput!): SellerAcceptOfferPayload
+  sellerCounterOfferOnOrder(input: SellerCounterOfferOnOrderInput!): SellerCounterOfferOnOrderPayload
```
ashkan18: sorry I know i suggested to add `OnOrder` here but talked to @sweir27 a while ago and after looking at other mutations looks like we are not as consistent as i thought we are. Looking at it more i think your original `sellerCounterOffer` makes more sense here. Not sure if there is a better name here since other mutations seem to have some verb in them `setShipping` and etc. What do you think about `sellerRespondToOffer`? this may make more sense since we also store this offer id as `respond_to_id` on `Offer`
Sorry for late comment on my original bad suggestion :(

app/services/base_total_calculator_service.rb
```diff
@@ -0,0 +1,52 @@
...
+      [artwork[:_id], artwork]
+    end]
+  end
+
+  def validate_artwork!(artwork)
+    raise Errors::ValidationError, :unknown_artwork unless artwork
+    raise Errors::ValidationError, :missing_artwork_location if artwork[:location].blank?
+  end
+
+  def set_offer_totals!
```
ashkan18: if this is the base class, maybe we should move this method and `set_offer_tax_total_cents` to `offer_total_updater_service.rb` and have this one only deal with generic stuff, what do you think?

_schema.graphql
sweir27: I guess `Counter` is a verb too so `sellerCounterOffer` actually sounds OK to me!

app/services/offers/submit_counter_offer_service.rb
```diff
@@ -0,0 +1,32 @@
...
+  class SubmitCounterOfferService
+    include OfferValidationService
+    def initialize(pending_offer:, from_id:)
+      @offer = pending_offer
+      @order = offer.order
+      @from_id = from_id
+    end
+
+    def process!
+      validate_offer_is_mine!(offer, @from_id)
```
sweir27: #minor The `_is_mine` naming seems a little odd in this context. Maybe just `validate_offer_owner!`?

sepans:
🍏

ashkan18:
awesome! thanks for your patience and dealing with me asking for renaming the thing i earlier asked to rename 🙌

ashkan18:
lets try BN and MO on staging once this gets deployed there.

app/services/base_total_calculator_service.rb
```diff
@@ -0,0 +1,52 @@
+class BaseTotalCalculatorService
+  def artworks
+    @artworks ||= Hash[@order.line_items.pluck(:artwork_id).uniq.map do |artwork_id|
+      artwork = GravityService.get_artwork(artwork_id)
+      validate_artwork!(artwork)
+      [artwork[:_id], artwork]
+    end]
```
yuki24: Late to the party, but you could also call the `#to_h` method as an alternative to `Hash[...]`:

```diff
-Hash[@order.line_items.pluck(:artwork_id).uniq.map do |artwork_id|
+@order.line_items.pluck(:artwork_id).uniq.map do |artwork_id|
   artwork = GravityService.get_artwork(artwork_id)
   validate_artwork!(artwork)
   [artwork[:_id], artwork]
-end]
+end.to_h
```

PR #301, opened 2018-11-29
Add from_participant to Offer and expose it in OfferType

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-664

We want to be able to show if an offer is from `Buyer` or `Seller`.

# Solution
We already had a logic to detect this so we can return `awaiting_response_from` on `OrderType`. We added `from_participant` helper to `Offer`, we updated `awaiting_response_from` to also use this newly added method and exposed the new field in `OfferType`.

# GraphQL Changes
- Added `from_participant` to `Offer`.

### DIFF
--- a/app/graphql/types/offer_order_type.rb
+++ b/app/graphql/types/offer_order_type.rb
@@ -12,7 +12,7 @@ class Types::OfferOrderType < Types::BaseObject
   def offers(**args)
     offers = object.offers.submitted
     offers = offers.where(args.slice(:from_id, :from_type)) if args.keys.any? { |ar| %i[from_id from_type].include? ar }
-    offers
+    offers.order(created_at: :desc)
   end
 
   def my_last_offer
@@ -24,10 +24,9 @@ class Types::OfferOrderType < Types::BaseObject
   def awaiting_response_from
     return unless object.mode == Order::OFFER && object.state == Order::SUBMITTED
 
-    if object&.last_offer&.from_id == object.seller_id && object&.last_offer&.from_type == object.seller_type
-      'buyer'
-    elsif object&.last_offer&.from_id == object.buyer_id && object&.last_offer&.from_type == object.buyer_type
-      'seller'
+    case object&.last_offer&.from_participant
+    when Order::BUYER then Order::SELLER
+    when Order::SELLER then Order::BUYER
     end
   end
 end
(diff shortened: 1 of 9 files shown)


app/graphql/types/offer_order_type.rb
```diff
@@ -12,7 +12,7 @@ class Types::OfferOrderType < Types::BaseObject
   def offers(**args)
     offers = object.offers.submitted
     offers = offers.where(args.slice(:from_id, :from_type)) if args.keys.any? { |ar| %i[from_id from_type].include? ar }
-    offers
+    offers.order(created_at: :desc)
```
ds300: 👍

app/models/offer.rb
```diff
@@ -12,4 +12,12 @@ def last_offer?
   def submitted?
     submitted_at.present?
   end
+
+  def from_participant
+    if from_id == order.seller_id && from_type == order.seller_type
```
ds300: Does this cover the case where two different users from the same partner are involved?

sweir27:
This looks great to meeee! Thank you!

spec/models/offer_spec.rb
```diff
@@ -31,4 +31,20 @@
...
+    let(:buyer_offer) { Fabricate(:offer, order: order, from_id: 'buyer1', from_type: 'user') }
+    let(:ufo_offer) { Fabricate(:offer, order: order, from_id: 'marse', from_type: 'ufo') }
+    it 'returns buyer for buyer_offer' do
+      expect(buyer_offer.from_participant).to eq Order::BUYER
+    end
+    it 'returns seller for seller_offer' do
+      expect(seller_offer.from_participant).to eq Order::SELLER
+    end
+    it 'returns nil for ufo_offer' do
+      expect(ufo_offer.from_participant).to be_nil
```
ds300: is this a valid use-case? When might from_type not be the same as order.seller_type or order.buyer_type?

If it's not a valid use-case, would it make sense to report an error instead of just returning nil?

app/models/offer.rb
ashkan18: yes, it should since in this case `buyer_id` and `buyer_type` are actually partner id and partner type and not the specific user.

spec/models/offer_spec.rb
ashkan18: yes, i wasn't sure actually about `nil` but i thought assumption here would be, we don't know the participant.

ds300: But how would that happen? Seems like it should be an error.

PR #303, opened 2018-11-29
AcceptOfferService deducts inventory and charges the buyer

williardx:
Addresses [PURCHASE-668](https://artsyproduct.atlassian.net/browse/PURCHASE-668).

This PR is for early feedback on creating our accept offer service layer.

 **Tests will be reorganized / added / fixed in future commits.**

### What Changed

- Submitting an order and accepting an offer are fairly similar to each other -- in both cases, we need to deduct inventory, charge the buyer and update the state of the order. So, I've created a new class `CommitOrderService` that represents this sequence of events. Child classes can specify the next state of the order and implement their own payment method. I chose the name "commit" because once the inventory is deducted nobody else can claim that artwork unless the order is rejected or refunded. I'm definitely open to other names, though!
- Right now `PaymentService` can only authorize charges and capture previously authorized charges. For an offer, though, we want to create a charge and capture it right away. This renames `PaymentService.authorize_charge` to `PaymentService.create_charge` and adds a `capture` parameter that specifies whether or not to immediately capture the charge. I also added two other methods `create_and_capture_charge` and `create_and_authorize_charge` to make the state of the funds very clear when we create a charge.
- We previously had a `BaseOfferService` containing some validation methods that `RejectOfferService` and `AcceptOfferService` both inherited from. I renamed this to `OfferValidationService` and converted it to a module so that we can use it as a class mixin and allow `AcceptOfferService` to inherit from `CommitOfferService` instead.
- `AcceptOfferService` now inherits from `CommitOfferService`.
- `OrderSubmitService` now inherits from `CommitOfferService`.

### DIFF
--- a/app/graphql/mutations/offer/base_accept_offer.rb
+++ b/app/graphql/mutations/offer/base_accept_offer.rb
@@ -12,7 +12,7 @@ class Mutations::Offer::BaseAcceptOffer < Mutations::BaseMutation
     authorize!(order)
     raise Errors::ValidationError, :cannot_accept_offer unless waiting_for_accept?(offer)
 
-    Offers::AcceptService.new(
+    Offers::AcceptOfferService.new(
       offer: offer,
       order: order,
       user_id: current_user_id
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -159,7 +159,7 @@ class Order < ApplicationRecord
     line_items.map(&:total_list_price_cents).sum
   end
 
-  def can_submit?
+  def can_commit?
     shipping_info? && payment_info?
   end
 
--- a/app/services/commit_order_service.rb
+++ b/app/services/commit_order_service.rb
@@ -4,10 +4,11 @@ class CommitOrderService
   end
 
   attr_accessor :order, :credit_card, :merchant_account, :partner
-  def initialize(order, order_state_action, by)
+
+  def initialize(order, order_state_action, user_id)
     @order = order
     @order_state_action = order_state_action
-    @by = by
+    @user_id = user_id
     @credit_card = nil
     @merchant_account = nil
     @partner = nil
@@ -24,13 +25,16 @@ class CommitOrderService
     undeduct_inventory
     raise e
   ensure
-    if @transaction.present?
-      @order.transactions << @transaction
-      notify_failed_charge if @transaction.failed?
-    end
+    handle_transaction
   end
   
   protected
+
+  def handle_transaction
+    return unless @transaction.present?
+    @order.transactions << @transaction
+    notify_failed_charge if @transaction.failed?
+  end
 
   def commit_order!
     @order.send(@order_state_action) do
@@ -53,6 +57,7 @@ class CommitOrderService
   end
   
   def process_payment
+    raise UnimplementedError
   end
 
   def pre_process!
@@ -75,7 +80,7 @@ class CommitOrderService
   end
 
   def notify_failed_charge
-    PostTransactionNotificationJob.perform_later(@transaction.id, TransactionEvent::CREATED, @by)
+    PostTransactionNotificationJob.perform_later(@transaction.id, TransactionEvent::CREATED, @user_id)
   end
 
   def construct_charge_params
@@ -91,13 +96,16 @@ class CommitOrderService
   end
 
   def assert_credit_card!
-    raise Errors::ValidationError.new(:credit_card_missing_external_id, credit_card_id: @credit_card[:id]) if @credit_card[:external_id].blank?
-    raise Errors::ValidationError.new(:credit_card_missing_customer, credit_card_id: @credit_card[:id]) if @credit_card.dig(:customer_account, :external_id).blank?
-    raise Errors::ValidationError.new(:credit_card_deactivated, credit_card_id: @credit_card[:id]) unless @credit_card[:deactivated_at].nil?
+    error_type = nil
+    error_type = :credit_card_missing_external_id if @credit_card[:external_id].blank?
+    error_type = :credit_card_missing_customer if @credit_card.dig(:customer_account, :external_id).blank?
+    error_type = :credit_card_deactivated unless @credit_card[:deactivated_at].nil?
+    raise Errors::ValidationError.new(error_type, credit_card_id: @credit_card[:id]) if error_type
   end
 
   def charge_description
-    "#{(@partner[:name] || '').parameterize[0...12].upcase} via Artsy"
+    partner_name = (@partner[:name] || '').parameterize[0...12].upcase}
+    "#{partner_name} via Artsy"
   end
 
   def charge_metadata
--- a/app/services/commit_order_service.rb
+++ b/app/services/commit_order_service.rb
@@ -1,8 +1,7 @@
 class CommitOrderService
+  attr_accessor :order
 
-  attr_accessor :order, :credit_card, :merchant_account, :partner
-
-  COMMITTABLE_ACTIONS = %i[ approve submit ]
+  COMMITTABLE_ACTIONS = %i[approve submit].freeze
 
   def initialize(order, action, user_id)
     @order = order
@@ -23,11 +22,12 @@ class CommitOrderService
   ensure
     handle_transaction
   end
-  
+
   protected
 
   def handle_transaction
-    return unless @transaction.present?
+    return if @transaction.blank?
+
     @order.transactions << @transaction
     notify_failed_charge if @transaction.failed?
   end
@@ -45,13 +45,13 @@ class CommitOrderService
   end
 
   def deduct_inventory
-      # Try holding artwork and deduct inventory
-      @order.line_items.each do |li|
-        GravityService.deduct_inventory(li)
-        @deducted_inventory << li
-      end
+    # Try holding artwork and deduct inventory
+    @order.line_items.each do |li|
+      GravityService.deduct_inventory(li)
+      @deducted_inventory << li
+    end
   end
-  
+
   def process_payment
     raise NotImplementedError
   end
@@ -63,10 +63,10 @@ class CommitOrderService
     validate_artwork_versions!
     validate_credit_card!
     validate_commission_rate!
-    
+
     OrderTotalUpdaterService.new(@order, partner[:effective_commission_rate]).update_totals!
   end
-  
+
   def validate_commission_rate!
     raise Errors::ValidationError.new(:missing_commission_rate, partner_id: partner[:id]) if partner[:effective_commission_rate].blank?
   end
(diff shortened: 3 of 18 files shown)


app/services/offers/accept_offer_service.rb
```diff
@@ -0,0 +1,33 @@
...
+    def process_payment
+      super
+      @transaction = PaymentService.create_and_capture_charge(construct_charge_params)
+      raise Errors::ProcessingError.new(:capture_failed, @transaction.failure_data) if @transaction.failed?
+    end
+
+    def pre_process!
+      super
+      validate_is_last_offer!(@offer)
+      validate_offer_is_from_buyer!(@offer)
```
williardx: Is there a reason to keep this validation in this service? We want to use the same service for both buyers and sellers accepting offers, right?

sepans: No that was my naive implementation and naming of the function. I just wanted to verify that offer to accept/reject is from the other side (in this case buyer) and not seller's own offer. So maybe one way to generalize it is to have something like `validate_offer_is_from_other_side` or something.

app/services/commit_order_service.rb
```diff
@@ -0,0 +1,113 @@
...
+      currency_code: @order.currency_code,
+      metadata: charge_metadata,
+      description: charge_description,
+    }
+  end
+
+  def assert_credit_card!
+    raise Errors::ValidationError.new(:credit_card_missing_external_id, credit_card_id: @credit_card[:id]) if @credit_card[:external_id].blank?
+    raise Errors::ValidationError.new(:credit_card_missing_customer, credit_card_id: @credit_card[:id]) if @credit_card.dig(:customer_account, :external_id).blank?
+    raise Errors::ValidationError.new(:credit_card_deactivated, credit_card_id: @credit_card[:id]) unless @credit_card[:deactivated_at].nil?
```
jonallured: I keep looking at this and wondering if there's a refactor - I'll take a shot here but if you don't think it's worth it, feel free to ignore!

```ruby
def assert_credit_card!
  error_type = nil
  error_type = :credit_card_missing_external_id if @credit_card[:external_id].blank?
  error_type = :credit_card_missing_customer if @credit_card.dig(:customer_account, :external_id).blank?
  error_type = :credit_card_deactivated unless @credit_card[:deactivated_at].nil?
  raise Errors::ValidationError.new(error_type, credit_card_id: @credit_card[:id]) if error_type
end
```

Just trying to boost the signal from the noise, ya know?

app/services/commit_order_service.rb
```diff
@@ -0,0 +1,113 @@
+class CommitOrderService
+  def self.call!(order, order_state_action:, by: nil)
+    new(order, order_state_action, by).process!
+  end
+
+  attr_accessor :order, :credit_card, :merchant_account, :partner
```
sweir27: #superminor can we add a space after this line?

app/services/commit_order_service.rb
```diff
@@ -0,0 +1,113 @@
...
+  def initialize(order, order_state_action, by)
+    @order = order
+    @order_state_action = order_state_action
```
sweir27: This is #minor but I find it slightly confusing that we're passing a method in here as opposed to `@order_new_state`, which could then call the correct method on `order` accordingly.

It makes it seem like you could pass _any_ method on order here, which is perhaps desired but I think in this case we're only dealing with state changes so it might be best to be explicit.

app/services/offers/accept_offer_service.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    def post_process!
+      super
+      instrument_order_approved
+    end
+
+    def instrument_order_approved
+      Exchange.dogstatsd.increment 'order.approve'
```
sweir27: Given that we're doing this same thing in the other case (instrumenting based on the state change), this could be moved into the parent class and removed from here.

app/services/commit_order_service.rb
jonallured: Excellent point! I'd rather not see the `send` but if we're going to, we should at least consider using a safelist of method names.

ashkan18: agreed, passing new state would  make things easier. Since this method currently only works for `approve` and `submit` the alternative is to have `submit!` and `approve!` instead of one `process!` and let the caller decide which action they want to take.

app/services/commit_order_service.rb
```diff
@@ -0,0 +1,113 @@
...
+    @by = by
```
ashkan18: https://github.com/artsy/exchange/pull/304 is renaming `by` to `user_id`, you may want to do it in your PR too.

app/services/commit_order_service.rb
```diff
@@ -0,0 +1,113 @@
...
+  end
+
+  def assert_credit_card!
+    raise Errors::ValidationError.new(:credit_card_missing_external_id, credit_card_id: @credit_card[:id]) if @credit_card[:external_id].blank?
+    raise Errors::ValidationError.new(:credit_card_missing_customer, credit_card_id: @credit_card[:id]) if @credit_card.dig(:customer_account, :external_id).blank?
+    raise Errors::ValidationError.new(:credit_card_deactivated, credit_card_id: @credit_card[:id]) unless @credit_card[:deactivated_at].nil?
+  end
+
+  def charge_description
+    "#{(@partner[:name] || '').parameterize[0...12].upcase} via Artsy"
```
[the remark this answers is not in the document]
ashkan18: worth a separate discussion, but I can't find the documentation on Stripe about this limitation anymore, we may want to just have whole partner name in there if there is no limit anymore.

app/services/commit_order_service.rb
williardx: Yep, that makes sense!

app/services/commit_order_service.rb
```diff
@@ -0,0 +1,113 @@
...
+    undeduct_inventory
+    raise e
+  ensure
+    if @transaction.present?
+      @order.transactions << @transaction
+      notify_failed_charge if @transaction.failed?
+    end
+  end
+  
+  protected
```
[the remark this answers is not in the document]
williardx: My understanding is that `private` methods won't be inherited in child classes but that `protected` methods will and still function somewhat like a private method. Is that right?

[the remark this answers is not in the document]
jonallured: Nope, private methods are inherited too.

app/services/commit_order_service.rb
```diff
@@ -0,0 +1,113 @@
...
+      end
+  end
+  
+  def process_payment
+  end
+
+  def pre_process!
+    @order.line_items.map do |li|
+      artwork = GravityService.get_artwork(li[:artwork_id])
+      Exchange.dogstatsd.increment 'submit.artwork_version_mismatch'
```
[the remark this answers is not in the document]
ashkan18: oh good catch! no, we need to do that only if there was a version mismatch.

app/models/order.rb
```diff
@@ -159,7 +159,7 @@ def total_list_price_cents
     line_items.map(&:total_list_price_cents).sum
   end
 
-  def can_submit?
+  def can_commit?
```
williardx: Renaming `can_submit?` to `can_commit?` is a new change in this most recent round of edits, mostly inspired by the fact that this is shared across services that involve `CommitOrderService`.

app/services/commit_order_service.rb
```diff
@@ -0,0 +1,140 @@
...
+
+  def process_payment
+    raise NotImplementedError
+  end
+
+  def pre_process!
+    raise Errors::ValidationError, :uncommittable_action unless COMMITTABLE_ACTIONS.include? @action
+    raise Errors::ValidationError, :missing_required_info unless @order.can_commit?
+
+    validate_artwork_versions!
```
williardx: I refactored `pre_process!` into a series of smaller methods for readability and testing.

spec/controllers/api/requests/offers/buyer_accept_offer_mutation_request_spec.rb
```diff
@@ -2,13 +2,41 @@
 require 'support/use_stripe_mock'
 
 describe Api::GraphqlController, type: :request do
+  include_context 'use stripe mock'
```
williardx: `buyer_accept_offer_mutation_request_spec` and `seller_accept_offer_mutation_request_spec` are virtually identical, save for a few details. This is a good candidate for future refactoring.

app/services/offers/accept_offer_service.rb
williardx: I wonder if we need to validate this here, since we validate it in the mutation?

app/graphql/mutations/offer/base_accept_offer.rb
```diff
@@ -12,7 +12,7 @@ def resolve(offer_id:)
     authorize!(order)
     raise Errors::ValidationError, :cannot_accept_offer unless waiting_for_accept?(offer)
 
-    Offers::AcceptService.new(
+    Offers::AcceptOfferService.new(
```
ashkan18: we can do this in follow up smaller PR, but since we already namespace this service with `Offers` module maybe we can rename this back to `AcceptService`.

app/services/commit_order_service.rb
```diff
@@ -0,0 +1,140 @@
...
+      @deducted_inventory << li
+    end
+  end
+
+  def process_payment
+    raise NotImplementedError
+  end
+
+  def pre_process!
+    raise Errors::ValidationError, :uncommittable_action unless COMMITTABLE_ACTIONS.include? @action
```
ashkan18: 👍

app/services/commit_order_service.rb
```diff
@@ -0,0 +1,140 @@
...
+  def process!
+    pre_process!
+    commit_order!
+    post_process!
+    @order
+  rescue Errors::ValidationError, Errors::ProcessingError => e
+    undeduct_inventory
+    raise e
+  ensure
+    handle_transaction
```
ashkan18: how about renaming this to `store_transaction`?

app/services/offers/accept_offer_service.rb
```diff
@@ -0,0 +1,27 @@
...
+    private
+
+    def process_payment
+      @transaction = PaymentService.create_and_capture_charge(construct_charge_params)
+      raise Errors::ProcessingError.new(:capture_failed, @transaction.failure_data) if @transaction.failed?
+    end
+
+    def pre_process!
+      super
+      validate_is_last_offer!(@offer)
```
ashkan18: i wonder if it makes more sense to first do `last_offer` validation and then do more involved validations in the `super`, what do you think?

app/services/offers/offer_validation_service.rb
```diff
@@ -1,10 +1,10 @@
 module Offers
-  class BaseOfferService
-    def validate_is_last_offer!
+  module OfferValidationService
```
ashkan18: nice!

ashkan18:
Epic PR! 🎉

PR #304, opened 2018-11-29
Rename `by` argument to `user_id`

dleve123:
Follow up and waiting on https://github.com/artsy/exchange/pull/295


ashkan18:
💯 thanks for follow up PR

PR #305, opened 2018-11-30
Remove reference to old PostNotificationJob

dleve123:
https://artsyproduct.atlassian.net/browse/PURCHASE-700

Now that the [commit][0] that refactored name of the job is on staging
and production and there are no scheduled jobs in Sidekiq with the name
PostNotificationJob, we are clear to remove this old alias.

[0]:aa470960628e0cc46636b3f947cd2c9f3152fe06


ashkan18:
nice and safe and easy 👍

PR #307, opened 2018-12-03
Add Papertrail to few other models

ashkan18:
# Change
Add `has_papertrail` to few other models, they will be less useful since these models don't technically get updated and are mostly created but still good to have them.

### DIFF
--- a/app/models/fulfillment.rb
+++ b/app/models/fulfillment.rb
@@ -1,4 +1,6 @@
 class Fulfillment < ApplicationRecord
+  has_paper_trail
+
   has_many :line_item_fulfillments, dependent: :destroy
   has_many :line_items, through: :line_item_fulfillments
 end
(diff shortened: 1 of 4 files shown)


williardx:
Looks good! Thanks @ashkan18

dleve123:
@williardx and @ashkan18 Apologies for the communication hiccup, but I think it would be best to version each table separately, allowing for us to potentially have different primary key types for each table. It's pretty simple and allows us to have the schema change for the primary tables without needing to accommodate a single versions table: the technique is described here: https://github.com/paper-trail-gem/paper_trail#6a-custom-version-classes

@joeyAghion scoped in the general work of wrapping up the paper_trail implementation into the platform sprint: https://artsyproduct.atlassian.net/browse/PLATFORM-1058

Could we revert this commit, so that we don't deploy to production?

PR #309, opened 2018-12-03
Papertrail Versions Followup

ashkan18:
# Changes
This is follow up work on https://github.com/artsy/exchange/pull/299 to delete older invalid records and then update the column to be `uuid`.
This happens in a migration script where we:
- Delete older records not matching `uuid` format. This happens in the migration where we delete all records that don't match Postgres's uuid format.
- Change the `item_id` column to be `uuid`, for this to work we need to change column with 'uuid USING CAST(item_id AS uuid)'

# Where are my older version?!
we delete them from here, you can find old versions in RDS snapshot with name `exchange-before-deleting-old-papertrail-versions`


dleve123:
I tested this locally and this looks good to me! Merging.

PR #311, opened 2018-12-04
Specify artsy/exchange as the `git-remote` for Hokusai

dleve123:
This should ensure that running a `hokusai pipeline promote` pushes Git
tags even if not specified in as a command line argument. See [hokusai
docs][0] for more information.

[0]:
https://github.com/artsy/hokusai/blob/master/docs/Configuration_Options.md#project-configuration


izakp:
LGTM!

PR #314, opened 2018-12-04
PLATFORM-1084: Update papertrail implementation

dleve123:
https://artsyproduct.atlassian.net/browse/PLATFORM-1084

* Rename `versions` -> `orders_versions`
  + Monkey patch PaperTrail::Version to specify as abstract class
    - this approach is documented in the [papertrail_gem’s README][0]
    - disable rubocop Rails/ApplicationRecord check
  + Add PaperTrail::OrderVersion to allow for specifying table_name

* Add `object_changes` to `order_versions`
  + Papertrail will now directly persist the diff
    in paper_trail object_changes

* Follow similar pattern to add versioning to `Fulfillment`, `LineItem`, `Offer` and `Transaction`

[0]:https://github.com/paper-trail-gem/paper_trail#6a-custom-version-classes


db/migrate/20181204182739_add_object_changes_to_order_versions.rb
```diff
@@ -0,0 +1,5 @@
+class AddObjectChangesToOrderVersions < ActiveRecord::Migration[5.2]
```
dleve123: It's important that the timestamp of this file name is in the past of the rename migration -- that ensures that the rename migration is run first.

spec/fabricators/fulfillment_fabricator.rb
```diff
@@ -0,0 +1,2 @@
+Fabricator(:fulfillment) do
```
dleve123: There are no validations on `Fulfillment`, so an empty factory was all I needed here. Happy to make the factory more meaningful if we want to though!

app/models/paper_trail/fulfillment_version.rb
```diff
@@ -0,0 +1,5 @@
+module PaperTrail
```
dleve123: I tried a few times to get all of these highly similar Version subclasses into a single file -- wasn't successful though. I think that would be optimal from a readability perspective. If anyone knows how to do that, please comment!

yuki24: Have you tried adding an alias? Like `app/models/paper_trail/version.rb` could have all the classes and any other files like `app/models/paper_trail/fulfillment_version.rb` could be just an alias to `app/models/paper_trail/version.rb`. Though I'm not confident that this will actually work.

spec/support/shared_examples/shared_examples_for_versioned_models.rb
```diff
@@ -0,0 +1,22 @@
+RSpec.shared_examples 'a papertrail versioned model', versioning: true do |model_symbol|
```
ashkan18: nice shared spec!

dleve123:
> Nice! would be nice to document this somewhere or add a danger rule or something so when someone added a new model and wanted to use papertrail they know that they need to follow this more involved pattern.

Great point @ashkan18. I'll definitely write up some documentation in the README of this project, and see if I can get a danger rule working.

spec/support/shared_examples/shared_examples_for_versioned_models.rb
dleve123: 🙇

app/models/paper_trail/fulfillment_version.rb
dleve123: By alias do you mean something like `FulfillmentVersion = Version`? I don't think that'll work as `self.table_name=` needs to be defined on a version class-by-version class basis. Were you thinking of something else?

dleve123:
@ashkan18 Would it be cool if we defer writing a Danger rule to a separate PR (I'm about to pair with Isac on other work)? I'll track the work now in a Github Issue and maybe try to pair on it with Orta during a Friday.

If that works for you, please review https://github.com/artsy/exchange/pull/314/commits/6d849dce55fa9f7aa40f3adbf813f12c0008eeeb and :shipit: if that looks good to you!

app/models/paper_trail/fulfillment_version.rb
yuki24: Alias as in the file system symbolic link:

```bash
ln -s app/models/paper_trail/version.rb app/models/paper_trail/fulfillment_version.rb
```

dleve123: Interesting. I'm gonna defer this until a later time if that's OK with you. I think introducing symlinks like this is kind of non-standard in a Rails app and probably not worth the complexity

ashkan18:
lets do this

PR #312, opened 2018-12-04
WIP: Initial work on buyer reject offer mutation.

ashfurrow:
This PR is for [PURCHASE-688](https://artsyproduct.atlassian.net/browse/PURCHASE-688). It adds the ability for buyers to reject offers.

So far we have just set up the initial mutation, this is a WIP because:

- [ ] Needs cleanup
- [ ] Specify new default for rejected state that accommodates sellers (and updates through app)
- [ ] Specify reject reason for sellers based on ID (in `base_reject_offer`)
- [ ] Manually test this works locally

@ashkan18 @sepans There are a few TODO comments in here with questions, if you'd like to respond inline, that would be awesome!

PR #316, opened 2018-12-04
Adds ability for a buyers to decline a seller's offer

iskounen:
This PR is for [PURCHASE-688](https://artsyproduct.atlassian.net/browse/PURCHASE-688). It builds on [a feature branch](https://github.com/artsy/exchange/pull/312) that @ashfurrow and I worked on together.

This is a WIP because:

- [x] Needs cleanup
- [x] Specify new default for rejected state that accommodates sellers (and updates through app)
- [x] Specify reject reason for sellers based on ID (in `base_reject_offer`)
- [x] Manually test this works locally


iskounen:
The CircleCI test failed because of the following error:
```
exchange_1        | rake aborted!
exchange_1        | 
exchange_1        | You changed your GraphQL types without updating the schema. 
exchange_1        | Run a `rake graphql:schema:idl` and commit to perform the schema update.
```

app/services/offers/reject_offer_service.rb
```diff
@@ -1,13 +1,15 @@
 module Offers
   class RejectOfferService
     include OfferValidationService
-    def initialize(offer:, reject_reason:)
+    def initialize(offer:, reject_reason:, user_id:)
       @offer = offer
       @reject_reason = reject_reason
+      @user_id = user_id
```
ashfurrow: Nice and simple 😄

app/graphql/mutations/offer/base_reject_offer.rb
```diff
@@ -0,0 +1,37 @@
...
+  field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure', null: false
+
+  def resolve(offer_id:, reject_reason: nil)
+    offer = Offer.find(offer_id)
+    order = offer.order
+
+    authorize!(order)
+    raise Errors::ValidationError, :cannot_reject_offer unless waiting_for_response?(offer)
+
+    if reject_reason.nil?
```
ashkan18: Nice! #minor i think we can also do
```ruby
reject_reason ||= current_user_id == offer.order.buyer_id ? Order::REASONS[Order::CANCELED][:buyer_rejected] : Order::REASONS[Order::CANCELED][:seller_rejected]
```

ashkan18: another alternative is also to have `def reject_reason` in the base class and have both `BuyerReject` and `SellerReject` implement their own.

app/graphql/mutations/offer/buyer_reject_offer.rb
```diff
@@ -0,0 +1,6 @@
+class Mutations::Offer::BuyerRejectOffer < Mutations::Offer::BaseRejectOffer
+  alias authorize! authorize_buyer_request!
+  def waiting_for_response?(offer)
+    offer.from_participant == Order::SELLER # TODO: Where does this constant come from?
```
ashkan18: its here https://github.com/artsy/exchange/blob/master/app/models/order.rb#L12

app/graphql/mutations/offer/seller_reject_offer.rb
```diff
@@ -0,0 +1,6 @@
+class Mutations::Offer::SellerRejectOffer < Mutations::Offer::BaseRejectOffer
+  alias authorize! authorize_seller_request!
+  def waiting_for_response?(offer)
+    offer.from_participant != Order::SELLER
```
ashkan18: we could also check `offer.from_participant == Order::BUYER`

app/graphql/types/mutation_type.rb
```diff
@@ -6,13 +6,14 @@ class Types::MutationType < Types::BaseObject
...
   field :buyer_accept_offer, mutation: Mutations::Offer::BuyerAcceptOffer
+  field :buyer_reject_offer, mutation: Mutations::Offer::BuyerRejectOffer
   field :add_initial_offer_to_order, mutation: Mutations::AddInitialOfferToOrder
   field :submit_order_with_offer, mutation: Mutations::SubmitOrderWithOffer
 
   # Seller
   field :approve_order, mutation: Mutations::ApproveOrder
   field :seller_accept_offer, mutation: Mutations::Offer::SellerAcceptOffer
-  field :seller_reject_offer, mutation: Mutations::SellerRejectOffer
+  field :seller_reject_offer, mutation: Mutations::Offer::SellerRejectOffer
```
ashkan18: 🕺

app/models/order.rb
```diff
@@ -28,7 +28,7 @@ class Order < ApplicationRecord
     # Items have been deemed unavailable and hold is voided.
     CANCELED = 'canceled'.freeze,
     # Order is completely fulfilled by the seller
-    FULFILLED = 'fulfilled'.freeze.freeze,
```
ashkan18: hehe how that happened?! it must have been a really cold day... 🆓zing!

app/models/order.rb
```diff
@@ -39,7 +39,9 @@ class Order < ApplicationRecord
       seller_rejected_offer_too_low: 'seller_rejected_offer_too_low'.freeze,
       seller_rejected_shipping_unavailable: 'seller_rejected_shipping_unavailable'.freeze,
       seller_rejected_artwork_unavailable: 'seller_rejected_artwork_unavailable'.freeze,
-      seller_rejected_other: 'seller_rejected_other'.freeze
+      seller_rejected_other: 'seller_rejected_other'.freeze,
+      seller_rejected: 'seller_rejected'.freeze,
```
ashkan18: i wonder if this is getting too big and we may want to divide it to two columns later, something like `reason` and `reasons_description` or something, having both `seller_rejected` and `seller_rejected_other` might be little confusing. Totally a separate discussion though, not related to this PR, just wanted to see what you think

spec/services/offers/reject_offer_service_spec.rb
```diff
@@ -3,8 +3,8 @@
 describe Offers::RejectOfferService, type: :services do
   describe '#process!' do
     let!(:order) { Fabricate(:order, state: Order::SUBMITTED) }
-    let!(:offer) { Fabricate(:offer, order: order) }
-    let(:service) { Offers::RejectOfferService.new(offer: offer, reject_reason: Order::REASONS[Order::CANCELED][:seller_rejected_offer_too_low]) }
+    let!(:offer) { Fabricate(:offer, order: order, from_id: 'buyer') }
```
ashkan18: #super_minor just not to confuse ourselves later i'd suggest rename the value to `collector_id` maybe here and for the service below have the user id be `partner_admin_id` or something.

app/graphql/mutations/offer/buyer_reject_offer.rb
ashfurrow: That syntax is really weird! It looks like it should only be in the `PARTICIPANTS` array?

app/graphql/mutations/offer/base_reject_offer.rb
iskounen: Amazing! I'm so glad that you caught this because I was _certain_ there was a better way to do it.

ashkan18: one could argue your version is more readable :)

app/graphql/mutations/offer/buyer_reject_offer.rb
ashfurrow: Following up on this, Ashkan explained to me in-person that the syntax above defines both the `PARTICIPANTS` array _and_ its members as constants, which is really cool 👍

iskounen: The `PARTICIPANTS` array contains the `SELLER` constant, so I also thought that this syntax was odd. However, I think that it makes sense because an array doesn't have scope. The scope of the `SELLER` constant and the `PARTICIPANTS` array are both the `Order` class.

app/graphql/mutations/offer/seller_reject_offer.rb
iskounen: I'm going to change that line to just be `offer.from_participant == Order::BUYER` so that this rule (sellers wait for buyers; buyers wait for sellers) remains intact if we add a 3rd participant type to the `Order` class.

sweir27:
#minor but since this PR does more than just prevent buyers from declining their own offers, can we rename it? Will make it easier to search for later 😄. Maybe something like `Adds ability for buyers to decline offers`?

ashkan18:
Nice! thanks for being patient and dealing with conflicts 🙌 merging! Also officially welcome to Exchange :)

PR #317, opened 2018-12-05
Renames partner_id to seller_id through app

ashfurrow:
I did a project-wide rename of `partner_id` to `seller_id` (which was something we did for the external API before the Buy Now launch, but hadn't done internally). After running the tests, I found a few spots to change back.


spec/controllers/api/graphql_controller_spec.rb
```diff
@@ -16,7 +16,7 @@
     end
 
     context 'authorized' do
-      let(:headers) { jwt_headers(user_id: 'user_id', partner_ids: ['partner_id']) }
+      let(:headers) { jwt_headers(user_id: 'user_id', seller_ids: ['partner_id']) }
```
ashkan18: `seller_ids` here should stay `partner_ids` since it's the JWT header.

spec/controllers/api/requests/error_handling_request_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe Api::GraphqlController, type: :request do
   describe 'ErrorType' do
-    let(:auth_headers) { jwt_headers(user_id: 'user-id', partner_ids: ['p1'], roles: nil) }
+    let(:auth_headers) { jwt_headers(user_id: 'user-id', seller_ids: ['p1'], roles: nil) }
```
ashkan18: same here, it should still be `partner_ids`

spec/controllers/api/graphql_controller_spec.rb
ashfurrow: Ah right, that makes sense.

ashkan18:
Thanks @ashfurrow , feel free to merge after conflict

ashkan18:
merge on green

PR #319, opened 2018-12-06
[WIP] Buyer counter offering

ashkan18:


app/graphql/mutations/base_mutation.rb
```diff
@@ -17,12 +17,22 @@ def current_user_id
...
 
-  def authorized_to_buy?(order)
-    context[:current_user]['id'] == order.buyer_id
+  def authorized_buyer?(item)
+    case item
+    when Order then context[:current_user]['id'] == item.buyer_id
+    when Offer then context[:current_user]['id'] == item.order.buyer_id
+    else
+      raise Errors::ValidationError, :not_found
+    end
```
yuki24: You can probably duck-type this:

```ruby
class Order
  ...

  def authorized_buyer?(user)
    user['id'] == buyer_id
  end
end

class Offer
  ...

  def authorized_buyer?(user)
    user['id'] == order.buyer_id
  end
end

raise ... unless item.authorized_buyer?(context[:current_user])
```

app/services/offers/offer_validation_service.rb
```diff
@@ -1,23 +1,19 @@
 module Offers
   module OfferValidationService
     def validate_is_last_offer!(offer)
-      raise Errors::ValidationError.new(:not_last_offer, offer) unless offer.last_offer?
+      raise Errors::ValidationError, :not_last_offer unless offer.last_offer?
     end
 
     def validate_offer_is_from_buyer!(offer)
-      raise Errors::ValidationError.new(:offer_not_from_buyer, offer) unless offer.from_type == Order::USER
+      raise Errors::ValidationError, :offer_not_from_buyer unless offer.from_type == Order::USER
```
ds300: shouldn't this use `from_participant`?

app/services/offers/submit_counter_offer_service.rb
```diff
@@ -1,31 +1,28 @@
...
   class SubmitCounterOfferService
+    attr_reader :offer
     include OfferValidationService
-    def initialize(pending_offer:, from_id:)
-      @offer = pending_offer
+    def initialize(offer, user_id:)
+      @offer = offer
       @order = offer.order
-      @from_id = from_id
+      @user_id = user_id
```
ds300: this `user_id` doesn't appear to be used

app/models/offer.rb
```diff
@@ -20,4 +20,13 @@ def from_participant
       Order::BUYER
     end
   end
+
+  def awaiting_response_from
+    return if submitted?
```
ds300: I'm confused about this line. Shouldn't it be the other way around? i.e. return nil if the offer was not yet submitted?

app/services/offers/offer_validation_service.rb
ds300: oh looks like this file got deleted while I was reviewing

ashkan18:
closing in favor of #324

PR #320, opened 2018-12-06
Add `creator_id` to AddPendingOfferService 

sepans:
and `sellerCounterOffer` mutation

Realized `from_id` of an offer, when submitted from the seller side is the `partner_id` not` user_id`.

In case of a buyer offer they are the same though.


ashkan18:
💯 👍

PR #322, opened 2018-12-07
Add last_offer to order event

ashkan18:
# Problem
We want to know who made `last_offer` in case of offers and also details about it when sending out emails from Pulse.

# Solution
Add `last_offer` to `order_event`'s `properties`.

### DIFF
--- a/app/events/order_event.rb
+++ b/app/events/order_event.rb
@@ -57,6 +57,7 @@ class OrderEvent < Events::BaseEvent
 
     last_offer = @object.last_offer
     {
+      id: last_offer.id,
       amount_cents: last_offer.amount_cents,
       shipping_total_cents: last_offer.shipping_total_cents,
       tax_total_cents: last_offer.tax_total_cents,
--- a/app/events/order_event.rb
+++ b/app/events/order_event.rb
@@ -43,13 +43,27 @@ class OrderEvent < Events::BaseEvent
   end
 
   def properties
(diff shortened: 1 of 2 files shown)


app/events/order_event.rb
```diff
@@ -52,6 +52,20 @@ def line_items_details
...
+    return unless @object.last_offer
+
+    last_offer = @object.last_offer
+    {
+      amount_cents: last_offer.amount_cents,
+      shipping_total_cents: last_offer.shipping_total_cents,
+      tax_total_cents: last_offer.tax_total_cents,
+      from_participant: last_offer.from_participant,
+      creator_id: last_offer.creator_id,
+      responds_to: last_offer.responds_to_id
```
starsirius: Can we have offer id as well? We'd like to log it when a counter offer submitted. 🙏

app/events/order_event.rb
```diff
@@ -52,6 +52,20 @@ def line_items_details
...
 
+  def last_offer
+    return unless @object.last_offer
+
+    last_offer = @object.last_offer
+    {
+      amount_cents: last_offer.amount_cents,
+      shipping_total_cents: last_offer.shipping_total_cents,
+      tax_total_cents: last_offer.tax_total_cents,
+      from_participant: last_offer.from_participant,
```
starsirius: @sepans Looks like we will have `from_participant`, instead of `from_type`.

app/events/order_event.rb
ashkan18: haha good call! adding it now.

PR #324, opened 2018-12-10
Buyer counter Offer

ashkan18:
# Problem
Add new `BuyerCounterOffer` mutation for creating pending counter offer.

# Changes
As part of this ended up doing bunch of other refactoring to DRY things up a little.

- Added `RequestValidator` module which has all request level validations we do in different queries and mutations.
- Updated `authorized_seller` and `authorized_buyer` methods in ☝️ to work with both `Offer` and `Order`, worth try CanCan at somepoint to see if it can make things little easier.
- Moved all service layer `validation` related checks to new `OrderValidator` module which can be `inculde`d in the services trying to validate order's data.
- Moved all order detail related methods, like getting artworks, merchant account, partner and etc to `OrderDetail` module which can be included in services to fetch these data as needed.
- Moved `GravityService` to `lib`
- Fixed an issue with setting totals on an offer with `shipping` fulfillment type
- Fixed setting `itemsTotals` when offer is submitted.

# Followup
fix `::Offers` thing when calling offer services form offers mutations


app/controllers/errors/error_types.rb
```diff
@@ -4,6 +4,7 @@ module Errors
       cant_offer
       cant_submit
       cannot_accept_offer
+      cannot_counter
```
sweir27: `cannot_` vs `cant_` above. We should keep consistent if possible!

app/services/offers/submit_offer_service.rb
```diff
@@ -0,0 +1,31 @@
...
+    attr_reader :offer
+    def initialize(offer)
+      @offer = offer
+      @order = offer.order
+    end
+
+    def process!
+      pre_process!
+
+      offer.update!(submitted_at: Time.now.utc)
```
ashkan18: move these to a `lock`

app/graphql/mutations/offers/buyer_counter_offer.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  argument :offer_id, ID, required: true
+  argument :amount_cents, Integer, required: true
+
+  field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure', null: false
+
+  def resolve(offer_id:, amount_cents:)
+    offer = Offer.find(offer_id)
+
+    validate_request(offer)
```
sweir27: Also #minor but I think elsewhere we're making these `!` functions since they can raise errors.

app/graphql/mutations/offers/buyer_counter_offer.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    add_service = Offers::AddPendingCounterOfferService.new(offer, amount_cents: amount_cents, from_type: Order::USER, from_id: context[:current_user][:id])
```
sweir27: Similar to @williardx 's change in https://github.com/artsy/exchange/pull/325, we should probably do `order.buyer_type` here instead of assuming `Order::USER`.

app/services/commit_order_service.rb
```diff
@@ -1,4 +1,6 @@
 class CommitOrderService
+  include OrderDetails
```
sweir27: Hrm. `OrderDetails` is sort of vague, but I can't think of better name right now. :(

app/services/concerns/order_details.rb
```diff
@@ -0,0 +1,30 @@
+module OrderDetails
+  def artworks
+    @artworks ||= Hash[@order.line_items.pluck(:artwork_id).uniq.map do |artwork_id|
+      artwork = GravityService.get_artwork(artwork_id)
+      validate_artwork!(artwork)
+      [artwork[:_id], artwork]
+    end]
+  end
+
+  def validate_artwork!(artwork)
```
sweir27: Should this live in `OrderValidator` instead?

app/services/offers/add_pending_counter_offer_service.rb
```diff
@@ -1,35 +1,35 @@
 module Offers
   class AddPendingCounterOfferService
-    include OfferValidationService
-    def initialize(offer:, amount_cents:, from_id:, creator_id:, from_type:)
-      @offer = offer
-      @order = offer.order
+    attr_reader :offer
+    include OrderValidator
```
sweir27: #minor but  should we stick the  `include`'s at  the top?

app/services/offers/add_pending_counter_offer_service.rb
```diff
@@ -1,35 +1,35 @@
...
+    def initialize(counter_on, amount_cents:, from_id:, from_type:, creator_id:)
```
sweir27: Should `counter_on` also be a keyed param?

app/services/offers/add_pending_counter_offer_service.rb
```diff
@@ -1,35 +1,35 @@
...
+      validate_action!
 
-      @pending_offer = @order.offers.create!(
+      @offer = @order.offers.create!(
         amount_cents: @amount_cents,
         from_id: @from_id,
         from_type: @from_type,
         creator_id: @creator_id,
-        responds_to: @offer
+        responds_to: @counter_on
```
sweir27: `counter_on` is  also a little confusing. Should we just reuse `responds_to` for this param name?

app/services/offers/submit_offer_service.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    private
+
+    def pre_process!
+      assert_submit!
+      validate_commission_rate!(partner)
+    end
+
+    def assert_submit!
+      raise Errors::ValidationError, :invalid_offer if @offer.submitted?
```
sweir27: If I'm reading these two classes correctly... it seems like:
- You should _not_ be able to submit an order with an offer that is already submitted
- You _should_ be able to submit an order with an offer that is _not_ already submitted, even if the order itself  is submitted.
☝️ this last  one is the one we should verify, since it'll fit our countering case.

app/services/concerns/order_details.rb
ashkan18: it's only used internally in `OrderDetails` as part of fetching artworks for an order, i could move it there but then need to include `OrderValidator` in `OrderDetails` which i think may confuse things a little, we may be good for now to keep it inside `order_details`, what do you think?

app/services/offers/add_pending_counter_offer_service.rb
ashkan18: I tried to make the first arguments of these services not keyed, mainly thinking first argument is what we act on and rest are the attributes of what we are doing. does that make sense?

app/services/offers/submit_offer_service.rb
ashkan18: so this service is only responsible for submitting an Offer so it can be shared between different scenarios where we submit an Offer. Because of that all it checks at the moment is to make sure the offer we are about to submit is not already submitted.
The callers of this service can verify the order status. Currently:
- `SubmitOrderService` checks if order is not submitted yet and then calls this service
- `SubmitCounterOfferService` checks if order is submitted and then calls this service

app/graphql/mutations/offers/buyer_counter_offer.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  private
+
+  def validate_request!(offer)
+    authorize_buyer_request!(offer)
+    raise Errors::ValidationError, :cannot_counter unless offer.awaiting_response_from == Order::BUYER
+  end
+
+  def current_user_id
```
ashkan18: not needed, it's already in `BaseMutation`

app/graphql/mutations/offers/seller_counter_offer.rb
```diff
@@ -8,18 +8,24 @@ class Mutations::Offers::SellerCounterOffer < Mutations::BaseMutation
...
+    validate_request!(offer)
     order = offer.order
-    from_id = order.seller_id
-    creator_id = context[:current_user][:id]
-    authorize_seller_request!(order)
-    add_service = Offers::AddPendingCounterOfferService.new(offer: offer, amount_cents: amount_cents, from_type: Order::PARTNER, from_id: from_id, creator_id: creator_id)
-    pending_offer = add_service.process!
 
-    submit_service = Offers::SubmitCounterOfferService.new(pending_offer: pending_offer, from_id: from_id)
+    add_service = Offers::AddPendingCounterOfferService.new(offer, amount_cents: amount_cents, from_type: order.seller_type, from_id: order.seller_id, creator_id: context[:current_user][:id])
```
ashkan18: switch to `current_user_id`

app/graphql/concerns/request_validator.rb
```diff
@@ -0,0 +1,32 @@
...
+      when Offer then item.order
+      end
+    current_user_is_on_seller = context[:current_user]['partner_ids'].include?(order.seller_id)
+    order.seller_type != Order::USER && current_user_is_on_seller
+  end
+
+  def authorized_buyer?(item)
+    case item
+    when Order then context[:current_user]['id'] == item.buyer_id
+    when Offer then context[:current_user]['id'] == item.order.buyer_id
```
sepans: #minor I guess the `case` part  of the above two functions could also be abstracted out.

spec/controllers/api/requests/offers/seller_counter_offer_mutation_request_spec.rb
```diff
@@ -134,21 +135,36 @@
...
+          expect(last_offer.submitted_at).not_to eq(nil)
           # for partner counte offer creator_id and from_id are different
-          expect(order.reload.last_offer.creator_id).to eq(user_id)
-          expect(order.reload.last_offer.from_id).to eq(partner_id)
-          # shouldn't update order amounts until offer is accepted
-          expect(order.reload.items_total_cents).to eq(0)
+          expect(last_offer.creator_id).to eq(buyer_id)
+          expect(last_offer.from_id).to eq(order_seller_id)
+          # should update order amounts when offer is submitted
+          expect(order.items_total_cents).to eq(400000)
```
sepans: 🙏

sweir27:
Looks good to me! Feel free to merge when ready. 😄

app/graphql/concerns/request_validator.rb
ashkan18: oh good point! ill do it in follow up PR

ashkan18:
merging this to fix `::` module issue in follow up PR

PR #325, opened 2018-12-10
Fix seller counter offer mutation bug and raise error on unknown offer participant type

williardx:
This PR addresses two issues:

1. In the seller counter offer mutation, `Order::PARTNER` is currently passed in as the `from_type` for every new pending offer created by `AddPendingCounterOfferService`. We should be using `order.seller_type` instead since the `seller_type` might not be "partner."
2. `Offer.from_participant` can currently return `nil` if, say, the offer `from_type` does not match the buyer or seller type. Since our UI depends on having this value not be null in some places (e.g., we use `Order.awaiting_response_from` in Reaction, which is defined by `Offer.from_participant`), this should  raise an error if we can't figure out the `from_type`.

### DIFF
--- a/app/graphql/mutations/offers/seller_counter_offer.rb
+++ b/app/graphql/mutations/offers/seller_counter_offer.rb
@@ -12,7 +12,7 @@ class Mutations::Offers::SellerCounterOffer < Mutations::BaseMutation
     from_id = order.seller_id
     creator_id = context[:current_user][:id]
     authorize_seller_request!(order)
-    add_service = Offers::AddPendingCounterOfferService.new(offer: offer, amount_cents: amount_cents, from_type: Order::PARTNER, from_id: from_id, creator_id: creator_id)
+    add_service = Offers::AddPendingCounterOfferService.new(offer: offer, amount_cents: amount_cents, from_type: order.seller_type, from_id: from_id, creator_id: creator_id)
     pending_offer = add_service.process!
 
     submit_service = Offers::SubmitCounterOfferService.new(pending_offer: pending_offer, from_id: from_id)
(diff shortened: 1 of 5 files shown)


app/models/offer.rb
```diff
@@ -20,6 +20,8 @@ def from_participant
       Order::SELLER
     elsif from_id == order.buyer_id && from_type == order.buyer_type
       Order::BUYER
+    else
+      raise Errors::ValidationError, :unknown_participant_type
```
ashkan18: i wasn't 100% about this one, kind of was thinking we decide if this offer is from buyer or seller, and if none we just don't know but the offer is still valid by itself. What do you think?

williardx: I agree with you on a conceptual level, but we don't currently have this use case (i.e. we always expect "buyer" or "seller") and it makes it easier to reason about our code if we constrain this a little. Do you ever see us wanting to create an offer that comes from an unknown participant type?

app/graphql/mutations/offers/seller_counter_offer.rb
```diff
@@ -12,7 +12,7 @@ def resolve(offer_id:, amount_cents:)
     from_id = order.seller_id
     creator_id = context[:current_user][:id]
     authorize_seller_request!(order)
-    add_service = Offers::AddPendingCounterOfferService.new(offer: offer, amount_cents: amount_cents, from_type: Order::PARTNER, from_id: from_id, creator_id: creator_id)
+    add_service = Offers::AddPendingCounterOfferService.new(offer: offer, amount_cents: amount_cents, from_type: order.seller_type, from_id: from_id, creator_id: creator_id)
```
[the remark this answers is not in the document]
williardx: Good point! I'll do that.

app/models/offer.rb
sweir27: I think this validation makes sense. From a user's perspective, it follows that someone who is not a member of the buyer/seller pair should not be allowed to make an offer.

PR #326, opened 2018-12-10
Expose the artwork version id

jonallured:
Once this is exposed, then I can update MP's schema and use it over there in the artwork version resolver.


ashkan18:
awesome thanks!!

PR #329, opened 2018-12-12
Move offers mutation to root

ashkan18:
# Problem
We have two `Offers` module. One under `mutations` another under `services`. Because of this, calling services `Offers::` from mutation `Offers` gives us loading errors. 

# Solution
We mainly added these two modules to group things together and not end up with lot of files in one directory to help finding things easier.
In case of mutations, each mutation name is usually descriptive and uniq so we are safe moving all the Offer ones to root folder and this way the only `Offers` module for now would be under services.


sweir27:
👍 easy enough!

PR #330, opened 2018-12-12
Trim leading and trailing whitespace when parsing region

williardx:
### Problem
State names must exactly match their corresponding name in Carmen in order to successfully be parsed into their abbreviation. A seller had a trailing space in their state name, which initially prevented an order from going through because we couldn't parse the state.

Addresses [PURCHASE-714](https://artsyproduct.atlassian.net/browse/PURCHASE-714).

### Solution
Trim leading and trailing whitespace when parsing the region in `Address`.

### DIFF
--- a/lib/address.rb
+++ b/lib/address.rb
@@ -48,6 +48,7 @@ class Address
   end
 
   def parse_region(country, region)
+    region = region&.strip
     return region unless country&.code == UNITED_STATES.code || country&.code == Carmen::Country.coded('CA').code
 
     parsed_region = country.subregions.named(region) || country.subregions.coded(region)
--- a/spec/lib/address_spec.rb
+++ b/spec/lib/address_spec.rb
@@ -40,6 +40,11 @@ describe Address do
     end
   end
   describe '#parse_region' do
+    let(:united_states) { Carmen::Country.coded('US') }
+    it 'strips trailing and leading whitespace from the region' do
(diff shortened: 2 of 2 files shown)


lib/address.rb
```diff
@@ -48,6 +48,7 @@ def parse(address)
   end
 
   def parse_region(country, region)
+    region = region&.strip
```
ashkan18: im thinking maybe we should do this cleanup on higher level, basically in the Graphql ruby argument coming in, [looks like](https://github.com/rmosolgo/graphql-ruby/wiki/How-To:-Use-prepare-to-modify-or-validate-arguments-or-inputs) we can use `prepare` method on each GraphQL `argument` to do extra validation/cleanup.  
I can imagine we'd need to update https://github.com/artsy/exchange/blob/master/app/graphql/inputs/shipping_attributes.rb#L9 and add `prepare` method which strips the data.
Thinking more about it, we may want to add `strip` to all those arguments anyway.

williardx: I'm less worried about user input than I am about input from CMS, where we've already had some issues. For that case, this seems like the most natural place to cover our bases.

ashkan18: feel free to merge, we can later look into that `prepare` thingy, looks useful in few of our cases. I think generally its good idea for us to make sure the input to our mutations are clean.

PR #331, opened 2018-12-12
Add an admin cancel state reason

jonallured:
I just noticed that we have a state reason called `seller_rejected_other` that can be used in one of two ways:

1. the seller rejects an offer in CMS, but picks “other” as the reason
2. the seller contacts Artsy and an admin canceled the order on their behalf

I believe we should create a new state reason just for this second case so that it's clear when it's being used.

In terms of production data, we have none of the former because MO hasn’t launched, but we have 2 of latter. Once this is merged and deployed, we'd want to do a migration of those 2 over to this new `state_reason` value.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -42,6 +42,7 @@ class Order < ApplicationRecord
       seller_rejected_artwork_unavailable: 'seller_rejected_artwork_unavailable'.freeze,
       seller_rejected_other: 'seller_rejected_other'.freeze,
       seller_rejected: 'seller_rejected'.freeze,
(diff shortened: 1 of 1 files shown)


ashkan18:
when rejecting an order we pass in `user_id` who rejected this order, i think when admin rejects and order (currently only available via CRT contacting engineers), they are rejecting on behalf of sellers. I think it would still be one of `seller_reject` reasons but we track who did it in `state_history` so that should tell us if it was rejected by admins or not.

jonallured:
Ok makes sense, thanks!!

PR #332, opened 2018-12-12
Update `state_expires_at` when counteroffer is submitted

sepans:
Notes:

* ~Wanted to add that `2.days` as a constant [here](https://github.com/artsy/exchange/blob/master/app/models/order.rb#L49-L53) but this is not really a 'state' expiration. If this sounds ok or have better suggestions let me know.~ Ended up reusing `Order::STATE_EXPIRATIONS['submitted']` since it would be easy to split it into two different state expiration values later if business logic changes.

### DIFF
--- a/app/services/offers/submit_counter_offer_service.rb
+++ b/app/services/offers/submit_counter_offer_service.rb
@@ -12,6 +12,7 @@ module Offers
       @order.with_lock do
         SubmitOfferService.new(@offer).process!
       end
+      @order.update!(state_expires_at: @order.state_expires_at + 2.days)
       post_process!
     end
 
--- a/app/services/offers/submit_counter_offer_service.rb
+++ b/app/services/offers/submit_counter_offer_service.rb
@@ -12,7 +12,7 @@ module Offers
       @order.with_lock do
         SubmitOfferService.new(@offer).process!
       end
-      @order.update!(state_expires_at: @order.state_expires_at + Order::STATE_EXPIRATIONS['submitted'])
+      @order.update!(state_expires_at: Time.now.utc + Order::STATE_EXPIRATIONS['submitted'])
       post_process!
     end
 
--- a/app/services/offers/submit_counter_offer_service.rb
+++ b/app/services/offers/submit_counter_offer_service.rb
@@ -12,7 +12,7 @@ module Offers
       @order.with_lock do
         SubmitOfferService.new(@offer).process!
       end
-      @order.update!(state_expires_at: @order.state_expires_at + 2.days)
+      @order.update!(state_expires_at: @order.state_expires_at + Order::STATE_EXPIRATIONS['submitted'])
       post_process!
     end
 
--- a/spec/controllers/api/requests/offers/seller_counter_offer_mutation_request_spec.rb
+++ b/spec/controllers/api/requests/offers/seller_counter_offer_mutation_request_spec.rb
@@ -12,7 +12,7 @@ describe Api::GraphqlController, type: :request do
     let(:artwork) { { _id: 'a-1', current_version_id: '1', location: artwork_location, domestic_shipping_fee_cents: 1000 } }
     let(:order_state) { Order::SUBMITTED }
     let!(:order) { Fabricate(:order, state: order_state, seller_id: order_seller_id, buyer_id: buyer_id, **shipping_info) }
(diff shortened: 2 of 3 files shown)


spec/controllers/api/requests/offers/seller_counter_offer_mutation_request_spec.rb
```diff
@@ -12,7 +12,7 @@
     let(:artwork) { { _id: 'a-1', current_version_id: '1', location: artwork_location, domestic_shipping_fee_cents: 1000 } }
     let(:order_state) { Order::SUBMITTED }
     let!(:order) { Fabricate(:order, state: order_state, seller_id: order_seller_id, buyer_id: buyer_id, **shipping_info) }
-    let!(:offer) { Fabricate(:offer, order: order, amount_cents: 10000, from_id: buyer_id, from_type: Order::USER, submitted_at: Time.now.utc) }
+    let!(:offer) { Fabricate(:offer, order: order, amount_cents: 10000, from_id: buyer_id, from_type: Order::USER, submitted_at: Time.now.utc - 10.hours) }
```
sepans: This change was not really needed but changed it to make it more conceptually correct.

app/services/offers/submit_counter_offer_service.rb
```diff
@@ -12,6 +12,7 @@ def process!
       @order.with_lock do
         SubmitOfferService.new(@offer).process!
       end
+      @order.update!(state_expires_at: @order.state_expires_at + Order::STATE_EXPIRATIONS['submitted'])
```
starsirius: Maybe this should be `Time.now.utc + Order::STATE_EXPIRATIONS['submitted']`? I was wondering if we can reuse [`order.update_state_timestamps`](https://github.com/artsy/exchange/blob/52aeba0997c4ba761837f7745e323c9c5d4f24ef/app/models/order.rb#L194-L197). Looks like we only want to update `state_updated_at` when the order state changes, but submitting a counter is a self transition.

starsirius: Also, maybe this reset `state_expires_at` can be done in [`Offers::SubmitOfferService`](https://github.com/artsy/exchange/blob/52aeba0997c4ba761837f7745e323c9c5d4f24ef/app/services/offers/submit_offer_service.rb), given it's the ultimate place an offer/counter gets submitted.

Right now only `SubmitOrderService` and this `SubmitCounterOfferService` use it, and in `SubmitOrderService`, resetting `state_expires_at` is taken care of by [`@order.submit!`](https://github.com/artsy/exchange/blob/52aeba0997c4ba761837f7745e323c9c5d4f24ef/app/services/offers/submit_order_service.rb#L14) so might be redundant, but that could be another option.

sepans: @starsirius thanks for catching `Time.now.utc` 🙏 

As for reusing `update_state_timestamps` for counteroffer I don't think updating `state_updated_at` without changing the state would cause a harm at the moment but it creates an inconsistency that could create some problems down the road.

To your other point about moving that line to `SubmitOfferService`, I agree that is the more logical place to add that and I thought about that but had no strong opinion which one to do.

So maybe @ashkan18 can help with those decisions.

@sweir27 yes that is coming up in a follow up PR: https://github.com/artsy/exchange/pull/335 (I don't dare to even think about doing a **Big PR**™️)

sweir27: Actually it does seem like there's some movement re: submitting offers (https://github.com/artsy/exchange/pull/336) so syncing up makes sense!

app/services/offers/submit_counter_offer_service.rb
```diff
@@ -12,6 +12,7 @@ def process!
       @order.with_lock do
         SubmitOfferService.new(@offer).process!
       end
+      @order.update!(state_expires_at: Time.now.utc + Order::STATE_EXPIRATIONS['submitted'])
```
ashkan18: Makes sense! We should probably move this up into the block to have everything in one transaction. I wonder if it makes sense to even move this to `SubmitOfferService`.

ashkan18: oh didn't see all the conversations above! yes lets sync up tomorrow 🙌

sepans:
Was taken care of in https://github.com/artsy/exchange/pull/336

PR #333, opened 2018-12-13
Removes RejectOfferService in favor of OrderCancellationService for rejecting offers

sweir27:
### Problem
We're not currently posting an event when an order is `canceled` because a buyer or seller rejects the offer.

### Solution
Our `RejectOfferService` wasn't doing a ton anyways, I moved the validation logic into the mutation and reused `OrderCancellationService` for this case. `reject!`-ing from this service will send a notification, like we want.


app/graphql/mutations/base_reject_offer.rb
```diff
@@ -11,9 +11,12 @@ def resolve(offer_id:, reject_reason: nil)
     order = offer.order
 
     authorize!(order)
+
+    # should check whether or not it's an offer-mode order
```
sweir27: @ashkan18 We talked about this a little last night-- do you think it's worth adding this check?

ashkan18: im currently looking at submit pending offer flow and we are on the border line of checking too many things and some of them seem unnecessary. For example, if we are checking that you can only add offers to `OfferOrder`s then we don't need this check here. I'd vote for not adding this check 🗳

ashkan18:
merge on green

PR #337, opened 2018-12-18
Do not raise error if artwork versions mismatch on accepting offer

williardx:
Addresses [PURCHASE-757](https://artsyproduct.atlassian.net/browse/PURCHASE-757).

Allows artwork mismatches when accepting an offer.


ashkan18:
awesome! thanks!  🙌

PR #338, opened 2018-12-18
Add OfferEvent and post when offer is submitted

ashkan18:
# Problem
When an offer is submitted currently we are posting an `Order::SUBMITTED` event. There are 2 issues with it:
1) Order was not actually submitted :) Order was _submitted_ before and at this stage we are only adding a new offer but submitting a _pending_ Offer. 
2) Currently ☝️ event is used for sending emails when offer is submitted and they rely on order event's `last_offer` to send email for last email. The issue is, we queue jobs for posting these events, and technically speaking its possible that the `last_offer` may not be the offer we wanted to send event for at the time we are processing the job. In this case consumers will receive an event but there is no way for them to find the offer triggered this event originally.

Because of ☝️ reasons, we want to send an offer-based event when an offer is submitted.

# Solution
Add an `OfferEvent` which gets an `Offer` as its main object.
Post ☝️ when an offer is submitted. An `OfferEvent` includes the offer details also order details are accessible in the `properties` section of the event under `properties[:order]`.

# Follow up
For now we are still posting `Order::Submitted` event but eventually once consumers have switched to new event for counter offer, we can stop sending that one.

### DIFF
--- a/app/events/offer_event.rb
+++ b/app/events/offer_event.rb
@@ -1,0 +1,72 @@
+class OfferEvent < Events::BaseEvent
+  TOPIC = 'commerce'.freeze
+  ACTIONS = [
+    CREATED = 'created'.freeze,
+    SUBMITTED = 'submitted'.freeze
+  ].freeze
+
+  def self.post(offer, action, user_id)
+    event = new(user: user_id, action: action, model: offer)
+    Artsy::EventService.post_event(topic: TOPIC, event: event)
+  end
+
+  def subject
+    {
+      id: @subject
+    }
+  end
+
+  def properties
+    {
+      amount_cents: @object.amount_cents,
+      submitted_at: @object.submitted_at,
+      from_participant: @object.from_participant,
+      order: order
+    }
+  end
+
+  private
+
+  def order
+    order = @object.order
+    OrderEvent::PROPERTIES_ATTRS.map { |att| [att, order.send(att)] }.to_h.merge(line_items: line_items_details, last_offer: last_offer)
+  end
+
+  def line_items_details
+    @object.order.line_items.map { |li| line_item_detail(li) }
+  end
+
+  def last_offer
+    return unless @object.order.last_offer
+
+    last_offer = @object.order.last_offer
+    in_response_to = if last_offer.responds_to
+                       {
+                         id: last_offer.responds_to.id,
+                         amount_cents: last_offer.responds_to.amount_cents,
+                         created_at: last_offer.responds_to.created_at,
+                         from_participant: last_offer.responds_to.from_participant
+                       }
+                     end
+    {
+      id: last_offer.id,
+      amount_cents: last_offer.amount_cents,
+      shipping_total_cents: last_offer.shipping_total_cents,
+      tax_total_cents: last_offer.tax_total_cents,
+      from_participant: last_offer.from_participant,
+      creator_id: last_offer.creator_id,
+      in_response_to: in_response_to
+    }
+  end
+
+  def line_item_detail(line_item)
+    {
+      price_cents: line_item.list_price_cents,
+      list_price_cents: line_item.list_price_cents,
+      artwork_id: line_item.artwork_id,
+      edition_set_id: line_item.edition_set_id,
+      quantity: line_item.quantity,
+      commission_fee_cents: line_item.commission_fee_cents
+    }
+  end
+end
--- a/app/events/offer_event.rb
+++ b/app/events/offer_event.rb
@@ -21,6 +21,13 @@ class OfferEvent < Events::BaseEvent
       amount_cents: @object.amount_cents,
       submitted_at: @object.submitted_at,
       from_participant: @object.from_participant,
+      last_offer: @object.last_offer?,
+      from_id: @object.from_id,
+      from_type: @object.from_type,
+      creator_id: @object.creator_id,
+      responds_to: @object.responds_to_id,
+      shipping_total_cents: @object.shipping_total_cents,
+      tax_total_cents: @object.tax_total_cents,
       order: order
     }
   end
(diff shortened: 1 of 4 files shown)


ashkan18:
I _think_ its ok for us to send an `OfferEvent` that says exactly what happened to an specific offer, in this case saying offer id #231231 was submitted, but have `last_offer` not be that same offer.

Basically we are saying someone has submitted this offer but at the time we are sending this, last offer is not this one which by itself is a useful information. I think at that point consumer can decide if they want to proceed with what they wanted to do with the event or not.

I agree though that sending everything based on the time job was queued would be the better solution (as long as we guarantee events are sent/processed in order).

app/services/offer_service.rb
```diff
@@ -40,11 +40,20 @@ def self.submit_pending_offer(offer)
...
   end
+
+  class << self
+    private
+
+    def post_submit_offer(offer)
+      OrderFollowUpJob.set(wait_until: offer.order.state_expires_at).perform_later(offer.order.id, offer.order.state)
+      PostOfferNotificationJob.perform_later(offer.id, OfferEvent::SUBMITTED, offer.creator_id)
+      # We are posting order.submitted event 👇, for now since Pulse (email service) is expecting that in case of submitting pending offer
+      # We need to send OfferEvent eventually in this case to be more accurate
```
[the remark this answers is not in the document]
ashkan18: good point

sweir27:
I think I would rather do the better solution now vs. adding this on top, as it adds a layer of complexity to downstream services that we could _also_ avoid.

ashkan18:
I think the happy path here for us would be to remove `last_offer` from the event (which i did in last commit). It is not useful in offer submitted event and if a consumer wants to know the current last offer they can always call exchange and ask for details.

Generally I think `Offer.submitted` event by itself is useful and correct event to send when an offer is submitted. The data passed in this event is also more focused around the offer and always correct. I can imagine offer submitted email could eventually switch to this and we may not need to call exchange for anything extra since it has the basic order data as well.

ashkan18:
We had some discussions offline, we decided there are 3 concerns:
1) Post `offer.submitted` when an offer is submitted :)
2) Since we queue/delay posting events make sure we use snapshot of the order/offer at the time of queuing for the posted event
3) Decide on if we want to still pass more sensitive things like `last_offer` in the event.

This PR only addresses 1. We have created a ticket for 2 https://artsyproduct.atlassian.net/browse/PURCHASE-759 and once we do 2, item 3 is less concerning issue.

app/events/offer_event.rb
```diff
@@ -0,0 +1,79 @@
...
+  def order
+    order = @object.order
+    OrderEvent::PROPERTIES_ATTRS.map { |att| [att, order.send(att)] }.to_h.merge(line_items: line_items_details)
+  end
+
+  def line_items_details
+    @object.order.line_items.map { |li| line_item_detail(li) }
+  end
+
+  def last_offer
```
sweir27: Seems like this is unused.

app/services/offer_service.rb
```diff
@@ -40,11 +40,19 @@ def self.submit_pending_offer(offer)
...
   end
+
+  class << self
+    private
+
+    def post_submit_offer(offer)
+      OrderFollowUpJob.set(wait_until: offer.order.state_expires_at).perform_later(offer.order.id, offer.order.state)
+      PostOfferNotificationJob.perform_later(offer.id, OfferEvent::SUBMITTED, offer.creator_id)
+      # We are posting order.submitted event 👇, for now since Pulse (email service) is expecting that in case of submitting pending offer
+      PostOrderNotificationJob.perform_later(offer.order.id, Order::SUBMITTED, offer.creator_id)
```
starsirius: So this will be removed once https://github.com/artsy/pulse/pull/323 is deployed! 👍

app/events/offer_event.rb
```diff
@@ -0,0 +1,56 @@
...
+      id: @subject
+    }
+  end
+
+  def properties
+    {
+      amount_cents: @object.amount_cents,
+      submitted_at: @object.submitted_at,
+      from_participant: @object.from_participant,
+      last_offer: @object.last_offer?,
```
starsirius: So this is at the time of the event, if the offer is the last offer of its order, right?

app/events/offer_event.rb
```diff
@@ -0,0 +1,56 @@
+class OfferEvent < Events::BaseEvent
+  TOPIC = 'commerce'.freeze
+  ACTIONS = [
+    CREATED = 'created'.freeze,
```
starsirius: The `CREATED` hasn't been used, right?

app/events/offer_event.rb
```diff
@@ -0,0 +1,56 @@
...
+      submitted_at: @object.submitted_at,
+      from_participant: @object.from_participant,
+      last_offer: @object.last_offer?,
+      from_id: @object.from_id,
+      from_type: @object.from_type,
+      creator_id: @object.creator_id,
+      responds_to: @object.responds_to_id,
+      shipping_total_cents: @object.shipping_total_cents,
+      tax_total_cents: @object.tax_total_cents,
+      order: order
```
starsirius: > 2. Since we queue/delay posting events make sure we use snapshot of the order/offer at the time of queuing for the posted event

👍 Yeah I'm also slightly concerned about the denormalized `order` in this event. Similar to the `last_offer` [above](https://github.com/artsy/exchange/pull/338/files#r244045447), the `order` is not the snapshot when the offer is submitted. It can already be changed (e.g. by following submitted offers) in case of `PostOfferNotificationJob` retries, etc.

app/services/offer_service.rb
ashkan18: yes! it's removed as part of https://github.com/artsy/exchange/pull/347

app/events/offer_event.rb
ashkan18: yes and should be more accurate with https://github.com/artsy/exchange/pull/347

app/events/offer_event.rb
ashkan18: no, i can remove it for now

app/events/offer_event.rb
ashkan18: yes :) https://github.com/artsy/exchange/pull/347 should fix this

PR #339, opened 2018-12-18
add datadog events on deploy

saolsen:
Relies on DD_API_KEY being set in circle which I've already set up.

I know @izakp mentioned doing this in hokusai deploy hooks but to do that we'd need some mechanism where we could know in the hook if we're deploying to staging or prod. That would probably be nice so I'm still open to that path but this is simple and should be a good way to get started.

Events look like this and can be overlaid on metric graphs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,9 @@ jobs:
       - run:
           name: Deploy
           command: hokusai staging deploy $CIRCLE_SHA1
+      - run:
+          name: Deploy Event
+          command: hokusai/deploy_event.sh exchange staging
 workflows:
   version: 2
   default:
--- a/hokusai/deploy_event.sh
+++ b/hokusai/deploy_event.sh
@@ -6,8 +6,8 @@
 
 curl -X POST -H "Content-type: application/json" \
 -d "{
-      \"title\": \"test API key\",
-      \"text\": \"does this work with curl?\",
+      \"title\": \"$1 was deployed to $2\",
+      \"text\": \"version: $CIRCLE_SHA1\",
       \"priority\": \"normal\",
       \"tags\": [\"service:$1\", \"env:$2\"],
       \"alert_type\": \"info\"
--- a/hokusai/deploy_event.sh
+++ b/hokusai/deploy_event.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+# ./deploy_event app env
+# eg:
+# ./deploy_event exchange production
+# Must have DD_API_KEY set to datadog api key.
+
+curl -X POST -H "Content-type: application/json" \
+-d "{
+      \"title\": \"test API key\",
+      \"text\": \"does this work with curl?\",
+      \"priority\": \"normal\",
+      \"tags\": [\"service:$1\", \"env:$2\"],
+      \"alert_type\": \"info\"
+}" \
+"https://api.datadoghq.com/api/v1/events?api_key=$DD_API_KEY"
--- a/hokusai/deploy_event.sh
+++ b/hokusai/deploy_event.sh
@@ -4,6 +4,8 @@
 # ./deploy_event exchange production
 # Must have DD_API_KEY set to datadog api key.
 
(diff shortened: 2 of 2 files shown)


ashkan18:
Nice! so to be sure, we won't yet be able to see prod deploys this way, right?

saolsen:
From my quick look at the hokusai post deploy hook it just runs a given command but doesn't set any vars or inject any info about the context.

ashkan18:
yes, if hokusai's `pre_deploy` was passing context we could just add this logic to https://github.com/artsy/exchange/blob/master/hokusai/config.yml#L3, right?

izakp:
Thinking that we could solve the problem, not by Hokusai injecting and env
car into each run, but by creating a cluster-wide env var that would be
injected into every pod.
https://kubernetes.io/docs/tasks/inject-data-application/podpreset/

saolsen:
I'd rather not use rake and have a solution that works for any project. This still seems like the simplest way to do it to me.

dleve123:
> Thinking that we could solve the problem, not by Hokusai injecting and env
car into each run, but by creating a cluster-wide env var that would be
injected into every pod.
https://kubernetes.io/docs/tasks/inject-data-application/podpreset/

@izakp So are you imaging that we could ensure that `CLUSTER_NAME` exists for every pod, with a value of `staging` or `production` (or something like that)?

If so, then if this deploy instrumentation happened within a pod, then that pod would have access to (1) the service name (that needs to already been in the ENV to have DD instrumentation going on) and (2) the env name (due to the cluster-level env). This should be enough to accomplish the goal here, right? I believe the only work would be tweaking hokusai to include the ENV when running post-deploys - aka tweaking [the `CommandRunner` call for post deploy](https://github.com/artsy/hokusai/blob/e61f78808d96dfacfc5638344fce631c1ff1f0dd/hokusai/services/deployment.py#L86). Is there a reason we couldn't/shouldn't include the ENV vars from the config map @izakp?

I don't think this would work on production ATM as @ashkan18 pointed out because exchange doesn't use Circle to deploy, developers do a `hokusai pipeline promote` on their developer laptops.

joeyAghion:
By my question I just meant doesn't Hokusai already have that context info _inside_ the `promote` or `deploy` operations (i.e., if this were completely integrated with those), not the post-deploy operation.

Either way, I agree that a bash script like this is a lowest common denominator among projects. Should we update Exchange to use a PR-driven release to take advantage of it?

ashkan18:
I agree current language agnostic approach makes sense 👍 I was looking to see if K8S can send webhooks on specific deployment events, looks like you can setup pod-level events but not on deployment level :( https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/

@joeyAghion my only concern with PR-based deploy is we may possibly wait in CircleCi traffic. Other than that 👍

izakp:
@dleve123 ...

> So are you imaging that we could ensure that CLUSTER_NAME exists for every pod, with a value of staging or production (or something like that)?

Yes, using the PodPreset resource to make env vars (more-or-less) available to every pod in the cluster (the PodPreset needs a namespace and pod label targets)

> If so, then if this deploy instrumentation happened within a pod, then that pod would have access to (1) the service name (that needs to already been in the ENV to have DD instrumentation going on) 

This is hitting the Datadog API directly, so does not need to find the agent

> and (2) the env name (due to the cluster-level env). This should be enough to accomplish the goal here, right? I believe the only work would be tweaking hokusai to include the ENV when running post-deploys - aka tweaking the CommandRunner call for post deploy. Is there a reason we couldn't/shouldn't include the ENV vars from the config map

We could also set the env in the configmap - all hokusai run / pre-deploy and post-deploy hooks include the project's env https://github.com/artsy/hokusai/blob/master/hokusai/services/command_runner.py#L32

Though even if we did set the `DD_API_KEY`, service name and environment in the project env, this script would still need access to the `$CIRCLE_SHA1` Sha1 somehow... so it looks like the pre-post-deploy hooks will need to interpolate variables somehow.

In any case I am in favor of using this and going to a PR-release strategy

saolsen:
Added a test for the env var. Gonna merge this and add it to some more projects.

PR #340, opened 2018-12-19
Adds buyer_lapse order action and follow-up-job

sepans:
Addresses [SELL-1244](https://artsyproduct.atlassian.net/browse/SELL-1244)

Adds `buyer_lapse` action to `Order` and `OrderFollowUpService`

### DIFF
--- a/app/services/order_cancellation_service.rb
+++ b/app/services/order_cancellation_service.rb
@@ -17,8 +17,6 @@ class OrderCancellationService
   def buyer_lapse!
     @order.buyer_lapse!
     PostOrderNotificationJob.perform_later(@order.id, Order::CANCELED)
-  ensure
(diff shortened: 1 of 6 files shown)


app/services/order_cancellation_service.rb
```diff
@@ -14,6 +14,13 @@ def seller_lapse!
     @order.transactions << @transaction if @transaction.present?
   end
 
+  def buyer_lapse!
+    @order.buyer_lapse!
+    PostOrderNotificationJob.perform_later(@order.id, Order::CANCELED)
+  ensure
```
ashkan18: you can remove this ensure block. we don't call Stripe at this point yet, that `@transaction` would be the result of calling Stripe.

PR #341, opened 2018-12-20
Separate line item and offer weird relation

ashkan18:
Revisiting `LineItem` and `Offer` weird relation.

# Problem
Currently there is a weird invisible relation between `Offer` and `LineItem`. When an order is an `Offer` order, the `total_amount_cents` on a `LineItem` is calculated based on `lastOffer` otherwise its calculated by `list_price * quantity`. This relates `LineItem` to `Offer` in a hidden way where has caused issues like #342.

# Solution
Have `LineItem` only reflect the listed price. Offers have their own total fields and when setting totals on an `Order` based on the order `mode` we either use `offer` data or `line_item`.

This PR does some refactoring to do ☝️ along with that did:
- updated `SubmitOrderWithOffer` to use `OfferService.submit_order_with_offer(offer)` and removed old style `Offers::SubmitOrderService`.
- renamed `OrderData` to `OrderHelper`.

PR #342, opened 2018-12-20
Quick version of the fix for wrong tax calculation

ashkan18:
# Problem (scary one)
After refactors in #336 where we consolidated the logic of what happens during offer submission to new style, we introduced a pretty complicated bug where after submitting an offer, we calculate tax based on the previous offer which then causes rest of the totals being calculated wrong.

https://artsyproduct.atlassian.net/browse/SELL-1262

# Cause
When consolidating the logic for submitting offer, we ended up having setting `last_offer` on an order in one update with setting other totals.
when calculating totals for an order, we use `order.items_total_cents` which is calculated by the sum of all line item's `total_amount_cents`. In case of `Offer` orders, line item's `total_amount_cents` based on `order.last_offer` and with our current logic, at the time we try to get `items_total_cents`, `last_offer` hasn't been updated to new submitted offer so we calculated using the old `last_offer` 😱 

# Quick Solution
This PR fixes the issue above by first updating `last_offer` on the `order` and then in separate call calculate totals and set it on the order.

# Things are Pretty Complicated Between Line Item and Offer, What Should We Do?
Started working on a longer term solution in #341 where I'm planning to decouple `LineItem` and `Offer` invisible relation. Basically the idea will be `LineItems` will always represent the list price value and offer calculations don't use `line_item` for calculating totals.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,7 +13,7 @@ Metrics/ParameterLists:
   Max: 9
 
 Metrics/AbcSize:
-  Max: 41
+  Max: 42
   Exclude:
     - "db/migrate/*"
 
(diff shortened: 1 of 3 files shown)


sweir27:
Thank you!!

PR #343, opened 2018-12-20
Move more internal services to lib

ashkan18:
This PR moves some of our internal service files (services that are not used directly externally by mutations/queries) to `lib`


ashkan18:
merge on green

PR #344, opened 2018-12-21
Check inventory before offer submit

ashkan18:
# Problem
We want to raise error when buyer/seller submit counter offers.

Fixes https://artsyproduct.atlassian.net/browse/PURCHASE-767

# Solution
In submission logic, check if we have enough inventory for this order's artwork.

# Raised Error
Processing error with `insufficient_inventory` as error code.

### DIFF
--- a/app/controllers/errors/error_types.rb
+++ b/app/controllers/errors/error_types.rb
@@ -55,12 +55,12 @@ module Errors
       capture_failed
       charge_authorization_failed
       insufficient_inventory
+      received_partial_refund
       refund_failed
-      tax_recording_failure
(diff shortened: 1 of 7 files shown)


joeyAghion:
To be clear, this is just a convenience for partners/buyers and not _necessary_, right? As long as we prevent orders from being "accepted" without available inventory, I'd expect things to stay consistent.

ashkan18:
yes this mainly to avoid extra  back and forth and avoid later confusions.

ashkan18:
it will be more accurately checked/deducted on acceptance time.

PR #346, opened 2018-12-24
Initial offer refactor

ashkan18:
Move initial offer to new style

based on https://github.com/artsy/exchange/pull/341


app/graphql/mutations/add_initial_offer_to_order.rb
```diff
@@ -7,9 +7,10 @@ class Mutations::AddInitialOfferToOrder < Mutations::BaseMutation
   def resolve(order_id:, amount_cents:)
     order = Order.find(order_id)
     authorize_buyer_request!(order)
-    service = Offers::InitialOfferService.new(order, amount_cents, context[:current_user]['id'])
-    service.process!
-    { order_or_error: { order: service.order } }
+    raise Errors::ValidationError, :cannot_offer unless order.state == Order::PENDING
```
ashkan18: by moving this validation to the mutation layer, we can reuse `create_pending_offer` for initial offer as well.

app/graphql/mutations/buyer_counter_offer.rb
```diff
@@ -11,7 +11,7 @@ def resolve(offer_id:, amount_cents:)
 
     validate_request!(offer)
 
-    pending_offer = OfferService.create_pending_offer(offer, amount_cents: amount_cents, from_id: offer.order.buyer_id, from_type: offer.order.buyer_type, creator_id: current_user_id)
+    pending_offer = OfferService.counter(offer, amount_cents: amount_cents, from_id: offer.order.buyer_id, from_type: offer.order.buyer_type, creator_id: current_user_id)
```
ashkan18: Separated `create_pending_offer` and `counter` since counter does more validation. `counter` uses `create_pending_offer` underneath.

app/services/order_shipping_service.rb
```diff
@@ -33,6 +33,7 @@ def process!
   def pre_process!
     raise Errors::ValidationError.new(:invalid_state, state: @order.state) unless @order.state == Order::PENDING
     raise Errors::ValidationError, :invalid_state if @pending_offer.present? && @pending_offer.submitted_at?
+    raise Errors::ValidationError, :missing_country if @fulfillment_type == Order::SHIP && @shipping_address&.country.blank?
```
ashkan18: this used to happen in `shipping_calculator_service` 👇 but I think it makes sense to have this validation here.

lib/offer_order_totals.rb
```diff
@@ -9,6 +9,8 @@ class OfferOrderTotals
   delegate :buyer_total_cents, to: :offer
 
   def initialize(offer)
+    raise NotImplementedError unless offer.order.mode == Order::OFFER
```
ashkan18: making sure this `offer_order_total` is not used by `buy` orders.

ashkan18:
im thinking of merging this more _refactoring_ PRs sooner so we can do more tests on staging. But would be great if you review these even after they were merged and i'll address comments later. Sorry for many refactoring PRs but i think this is the best time to do it since they are really conflict-friendly.

app/graphql/mutations/buyer_counter_offer.rb
starsirius: Naming is hard! `counter` here sounds like a combined action but is actually `create_pending_counter_offer`, and we still need to `submit_pending_offer`. I'd suggest making it more explicit. I also like the concise `counter` to be `create_pending_counter_offer` + `submit_pending_offer` for seller's scenario.

```ruby
module OfferService
  def self.create_pending_offer
  end

  def self.create_pending_counter_offer
    # additional validation
    create_pending_offer
  end

  def self.submit_pending_offer
  end

  def self.counter
    create_pending_counter_offer
    submit_pending_offer
  end
end
```

lib/offer_order_totals.rb
starsirius: We are passing an offer already, so this means the order has been in an incorrect state somehow, i.e. with offer(s) but mode buy.

ashkan18: good point! here i wanted to just make sure no one uses this class for wrong type of orders, i can remove it.

app/graphql/mutations/buyer_counter_offer.rb
ashkan18: `create_pending_offer` and `create_pending_counter_offer` makes total sense, right now the actual `counter` (combined call) happens in the mutation, we may be better leave it there to not confuse people, what do you think?

lib/offer_order_totals.rb
ashkan18: addressed https://github.com/artsy/exchange/pull/353

PR #347, opened 2018-12-25
Use order/offer snapshot before queuing the event job

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-761
We queue jobs for posting order and offer events. Depending on whats the current state of the order/offer some fields may be wrong if by the time Sidekiq is processing the job, the order/offer is already updated.

# Solution
Add `delay_post` to both `OrderEvent` and `OfferEvent` which create the event and queue a newly added `PostEventJob` which gets an event json string and also the routing key.

# Follow up
We can remove `PostOrderNotifictionJob` and `PostOfferNotificationJob` since we've moved to `PostEventJob`.

### DIFF
--- a/app/events/offer_event.rb
+++ b/app/events/offer_event.rb
@@ -10,6 +10,11 @@ class OfferEvent < Events::BaseEvent
     Artsy::EventService.post_event(topic: TOPIC, event: event)
   end
 
+  def self.delay_post(offer, action)
+    event = new(user: offer.creator_id, action: action, model: offer)
+    PostEventJob.perform_later(event.to_json, event.routing_key)
+  end
+
   def subject
     {
       id: @subject
--- a/app/events/order_event.rb
+++ b/app/events/order_event.rb
@@ -37,6 +37,11 @@ class OrderEvent < Events::BaseEvent
     Artsy::EventService.post_event(topic: TOPIC, event: event)
   end
 
+  def self.delay_post(order, action, user_id = nil)
+    event = new(user: user_id, action: action, model: order)
+    PostEventJob.perform_later(event.to_json, event.routing_key)
+  end
+
   def subject
     {
       id: @subject
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -64,11 +64,11 @@ GEM
       omniauth-artsy (>= 0.2.2)
       omniauth-oauth2
       rails (>= 4.2.0)
-    artsy-eventservice (1.0.8)
+    artsy-eventservice (1.0.9)
       bunny
     ast (2.4.0)
     builder (3.2.3)
-    bunny (2.12.0)
+    bunny (2.13.0)
       amq-protocol (~> 2.3, >= 2.3.0)
     byebug (10.0.2)
     capybara (3.10.0)
--- a/app/jobs/post_event_job.rb
+++ b/app/jobs/post_event_job.rb
@@ -1,0 +1,7 @@
+class PostEventJob < ApplicationJob
+  queue_as :default
+
+  def perform(event_json, routing_key)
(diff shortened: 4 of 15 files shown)


app/events/order_event.rb
```diff
@@ -36,6 +36,11 @@ def self.post(order, action, user_id)
     Artsy::EventService.post_event(topic: TOPIC, event: event)
   end
 
+  def self.delay_post(order, action, user_id = nil)
```
starsirius: #minor. The signatures of `OrderEvent.delay_post` and `OfferEvent.delay_post` are slightly different:

- `OfferEvent` always assumes the user is the creator of the offer.
- `OrderEvent` takes an optional user for cases where action is not explicitly initiated by a user, e.g. order lapsed.

It seems sufficient for current use cases, but we could also make them consistent.

app/events/offer_event.rb
```diff
@@ -9,6 +9,11 @@ def self.post(offer, action, user_id)
     Artsy::EventService.post_event(topic: TOPIC, event: event)
   end
 
+  def self.delay_post(offer, action)
+    event = new(user: offer.creator_id, action: action, model: offer)
+    PostEventJob.perform_later(event.to_json, event.routing_key)
```
starsirius: 😎

starsirius: Do we need to worry about payload size?

joeyAghion: Can we use `as_json` on these event objects instead of `to_json`? I think it's strange to pass around JSON strings as arguments between ruby methods rather than the structured equivalent.

ashkan18:
@joeyAghion there were 2 issues with our past approach:
1) We were sending `order.submitted` and `OrderEvent` when an offer was submitted, in this case, consumers had to rely on `last_offer` to find out what actually happened and `last_offer` can be wrong since the job is queued. This was taken care of in #338 
2) Now with posting proper `offer.submitted` ☝️ is less scary since consumers know exactly which offer was submitted and dont relay on `last_offer` BUT there are some aggregated fields on the order that may be out of date with the current offer. Think of `tax_total_cents` may not match the `tax_total_cents` of the `Offer`, this is a less scary problem than issue 1 but still possible. I think it would be safer to pass the event as snapshot of the order/lineitem/offer relation.

app/events/offer_event.rb
ashkan18: actually, very interesting point... playing with this locally, i don't think `to_json` does what i expect with current `artsy-eventservice`'s `BaseEvent` https://github.com/artsy/artsy-eventservice/blob/master/lib/artsy-eventservice/presenters/events/base_event.rb
we need to call `.json` to actually get proper event.
`to_json` uses `@properties` which is `nil` while `.json` uses `properties` method in `BaseEvent`. I can update the gem to work with `to_json`

app/events/order_event.rb
ashkan18: good point, makes sense to ask for `user_id`, in case of `submit` it will be the same but i can see other actions having different users.

app/events/offer_event.rb
ashkan18: @starsirius RE payload size, yes we should monitor sidekiq memory usage.

ashkan18:
separated the part of this PR that stops sending `order.submitted` events in https://github.com/artsy/exchange/pull/356 so we can continue discussion on this PR :)

ashkan18:
alright this is rebased and should be 👍

ashkan18:
@joeyAghion @yuki24 if you are ok with this lets merge and i'll try things on staging..

ashkan18:
thanks all! going to merge this and try on staging tomorrow

PR #348, opened 2018-12-26
[WIP] Move shipping payment

ashkan18:
Based on #346


ashkan18:
This is mostly moving internal services to `lib` and change the style of the service from initiated class to class method since there was only one method exposed from `ShippingCalculatorService`

PR #349, opened 2018-12-27
Find or create offer order

ashkan18:
# Change
We want to reuse existing order when a user with existing `pending` or `submitted` MO order clicks on "Make Offer". This only happens for MO and also should consider edition set when checking for existing orders. 

https://artsyproduct.atlassian.net/browse/PURCHASE-772


lib/order_creator.rb
```diff
@@ -0,0 +1,127 @@
+class OrderCreator
```
starsirius: I'm a bit unclear about the difference between services and lib. `OrderCreator` sounds like a single responsibility service to me. Maybe we can have some guidance around the differentiation?

ashkan18: good question, this is a new style i'm trying where the _service_ is something thats only exposed externally to _view_ layer, think of mutations/active admin/webhook calls and etc.
The logic for some of these service calls can get pretty complicated, in that case im thinking of using helpers, or classes that do one thing based on our original objet, like creating order or setting shipping. These kind of helpers can live in `lib` so our main code is easier to traverse and find.
Let me know what you think, once we are good with this style I can document the idea.

lib/order_creator.rb
```diff
@@ -0,0 +1,127 @@
...
+    @user_agent = user_agent
+    @user_ip = user_ip
+  end
+
+  def valid?
+    @valid ||= valid_artwork? && valid_edition_set? && valid_action? && valid_price?
+  end
+
+  def find_or_create!(&post_create_block)
+    existing_order = Order.joins(:line_items).find_by(buyer_id: @buyer_id, buyer_type: @buyer_type, state: [Order::PENDING, Order::SUBMITTED], line_items: { artwork_id: @artwork_id, edition_set_id: edition_set_id })
```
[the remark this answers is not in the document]
ashkan18: Basically from mutation perspective nothing is different, when force (in reality Reaction) gets the order back based on the state it will redirect you to the correct page, so if you have pending offers you can update them but if you are waiting for buyer's response it will take you to status page.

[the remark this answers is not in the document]
ashkan18: `edition_set_id` should be correct. To have Eigen working, we have this logic that if an artwork has only one edition set, we don't require client to tell us the edition set id and we pick based on the artwork. `edition_set_id` method should do that.

app/services/order_service.rb
```diff
@@ -1,4 +1,14 @@
 module OrderService
+  def self.create_with_artwork!(buyer_id:, buyer_type:, mode:, quantity:, artwork_id:, edition_set_id: nil, user_agent:, user_ip:)
+    order_creator = OrderCreator.new(buyer_id: buyer_id, buyer_type: buyer_type, mode: mode, quantity: quantity, artwork_id: artwork_id, edition_set_id: edition_set_id, user_agent: user_agent, user_ip: user_ip)
+    # in case of Offer orders, we want to reuse existing pending/submitted offers
+    create_method = mode == Order::OFFER ? :find_or_create! : :create!
+    order_creator.send(create_method) do |created_order|
+      Exchange.dogstatsd.increment 'order.create'
```
ds300: should this and the follow up job still happen even for find_or_create?

ds300: Oh I see this only happens if the create! method is called. nvm

ashkan18:
Good point @joeyAghion , the refactor part is pretty isolated. The business logic change is technically only https://github.com/artsy/exchange/pull/349/files#diff-ac4b06920a6a91788513095812abc129R5 where we call `find_or_create` in case of Offers.

The main reason behind having Exchange returning the same order is the fact that we didn't want to touch iOS code already in the app. Otherwise we have to build the same logic there. 

From the API perspective, I agree it's not the most visible approach but I wonder if thats ok in our case where we have very specific mutations. Basically saying `createOrderWithOffer` reuses existing orders if there was any.

joeyAghion:
Understood. In retrospect I wonder if we could have held back the Eigen implementation until the possible back-end results were more clear.

A good addition to this PR might be to modify the mutation's return type declaration with some documentation of this behavior.

ashkan18:
makes sense I'll document the behavior in the mutation description. Also thinking of adding an optional attribute to change the behavior to not reuse to make it even more obvious.

ashkan18:
going to merge this so we can have enough time on staging for testing

PR #354, opened 2018-12-28
Move to new style for shipping service

ashkan18:
# Refactoring

This PR follows the same pattern we've been doing where we:
- Move our _external_ facing services to `order_service` as class methods
- Create helpers in `lib` when logic in the ☝️ was complicated

# Changes
- Deleted `OrderShippingService`
- Added `set_shipping` to `OrderService`
- Created `OrderShipping` helper which gets initiated with an `order` and you can call `pickup` or `ship` on it and it handles setting those on the order
- Added `LineItemTotals` which calculate totals on a given line item

- Also moved `OrderHelper` to be included in `Order` with @sweir27's suggestion, this makes getting order details more natural
- Added `LineItemHelper` which provides helper to get some details about the line item like artwork and etc. included that on the `Order` so we can do things like `line_item.artwork` now

# Migration?
As part of this refactoring one thing came up was how we calculate shipping currently. We have the notion of flat rate shipping and the question was if this flat rate is per item? meaning if you get 2 quantity of the same artwork should we double the shipping?
After talking to @joanna20 , it seems thats the case and we are going to double the costs. For that to work we need to add `shipping_total_cents` to `LineItem` to calculate shipping at the time of set shipping, then `shipping_total_cents` on the order ends up being the sum of all line items.
This kind of matches our pattern for other totals.

As part of this migration we copy `shipping_total_cents` to existing line items.

# Selfie
selfie-0
[_GitHub Selfies_](https://github.com/thieman/github-selfies/)


app/services/order_service.rb
```diff
@@ -9,6 +9,17 @@ def self.create_with_artwork!(buyer_id:, buyer_type:, mode:, quantity:, artwork_
     end
   end
 
+  def self.set_shipping!(order, fulfillment_type:, shipping:)
+    raise Errors::ValidationError, :invalid_state unless order.state == Order::PENDING
+
+    order_shipping = OrderShipping.new(order)
```
sweir27: Somehow these names (like `OrderShipping`) still seem a bit _naked_. What is this actually creating? Maybe `OrderShippingCalculator` is more accurate?

lib/line_item_totals.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    @shipping_total_cents ||= begin
+      per_item_shipping_cents = ShippingHelper.calculate(artwork, @order.fulfillment_type, @order.shipping_address)
+      per_item_shipping_cents * @line_item.quantity
+    end
+  end
+
+  private
+
+  def artwork
```
sweir27: This method and `artwork_location` feel like something that could live in a similar `LineItemHelper` module, as I imagine they're accessed elsewhere.

lib/order_shipping.rb
```diff
@@ -0,0 +1,71 @@
...
+    offer_totals = OfferTotals.new(@order, pending_offer.amount_cents)
+    pending_offer.update!(
+      shipping_total_cents: offer_totals.shipping_total_cents,
+      tax_total_cents: offer_totals.tax_total_cents,
+      should_remit_sales_tax: offer_totals.should_remit_sales_tax
+    )
+  end
+
+  def set_order_totals!
+    # update order line items
```
sweir27: Not sure you need this comment, pretty clear what's happening below

lib/order_shipping.rb
```diff
@@ -0,0 +1,71 @@
...
+    @order.line_items.map do |li|
+      line_item_totals = LineItemTotals.new(li, fulfillment_type: @order.fulfillment_type, shipping_address: @order.shipping_address, seller_locations: @order.seller_locations, artsy_collects_sales_tax: @order.artsy_collects_sales_tax?)
```
sweir27: Can we move more of these arguments to new lines? It's pretty hard to read in a long string like this

lib/order_shipping.rb
```diff
@@ -0,0 +1,71 @@
...
+      li.update!(shipping_total_cents: line_item_totals.shipping_total_cents, sales_tax_cents: line_item_totals.tax_total_cents, should_remit_sales_tax: line_item_totals.should_remit_sales_tax)
+    end
+    buy_order_totals = BuyOrderTotals.new(@order)
```
sweir27: #minor naming but `BuyOrderCalculator` might be more explicit here.

spec/lib/line_item_totals_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    allow(Adapters::GravityV1).to receive(:get).with('/artwork/a-1').and_return(artwork)
+  end
+  describe 'shipping_total_cents' do
+    context 'artwork missing location' do
+      let(:artwork) { gravity_v1_artwork(location: nil) }
+      it 'raises error' do
+        expect { line_item_totals.shipping_total_cents }.to raise_error(Errors::ValidationError)
+      end
+    end
+    it 'returns nil if missing shipping info' do
```
sweir27: Can we add some newlines in this file? 😄 like, after `before` and `it` blocks, etc.

spec/lib/order_shipping_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    allow(Adapters::GravityV1).to receive(:get).with("/partner/#{seller_id}/locations?private=true").and_return([{ country: 'US', state: 'NY' }])
+    allow(Adapters::GravityV1).to receive(:get).with("/artwork/#{artwork_id}").and_return(gravity_v1_artwork(_id: artwork_id))
+  end
+  describe '#pickup!' do
+    context 'Buy Order' do
+      before do
+        allow_any_instance_of(Tax::CalculatorService).to receive_messages(sales_tax: 100, artsy_should_remit_taxes?: false)
+        order_shipping.pickup!
+      end
+      it 'sets pickup fulfillment type' do
```
sweir27: ditto to the newlines comment above!

app/services/order_service.rb
ashkan18: so this _helper_ is the one that actually does the bulk of the work by setting shipping info on the order and calculate the totals. `OrderShippingHelper`, I wonder what we can do to make it more obvious that its actually setting fields on the `Order`.

lib/line_item_totals.rb
ashkan18: yes makes sense. going to make that change 🙌

spec/lib/line_item_totals_spec.rb
ashkan18: haha 👍

lib/order_shipping.rb
ashkan18: yes, we can rename all those in a separate follow up since there are few other "...Totals" helpers. what do you think? that can add lot of noise to this PR

sweir27: Sounds good!

lib/line_item_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def inventory?
+    inventory = if edition_set_id.present?
+                  edition_set = artwork[:edition_sets].detect { |a| a[:id] == edition_set_id }
+                  raise Errors::ValidationError, :unknown_edition_set unless edition_set
+
+                  edition_set[:inventory]
+                else
+                  artwork[:inventory]
+                end
```
joeyAghion: (Minor) I prefer the style where lines are indented just 1 step from the initial line (and I recall that rubocop has an option to accept this).

lib/line_item_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+    inventory[:count].positive? || inventory[:unlimited] == true
```
joeyAghion: You might not need `== true` here.

lib/line_item_helper.rb
ashkan18: yes! I agree, let me give it a try :)

PR #360, opened 2019-01-03
Reuse orders only if it's same mode

ashkan18:
# Problem
When a user has a `buy` order, when they try to make offer we return their `buy` order. 

https://artsyproduct.atlassian.net/browse/PURCHASE-778

# Cause
When checking for existing orders, we were not checking `mode`.

# Solution
Check for `mode` when looking for `existing_order `

### DIFF
--- a/spec/controllers/api/requests/offers/create_offer_order_with_artwork_mutation_request_spec.rb
+++ b/spec/controllers/api/requests/offers/create_offer_order_with_artwork_mutation_request_spec.rb
@@ -250,9 +250,25 @@ describe Api::GraphqlController, type: :request do
               end.to change(Order, :count).by(1).and change(LineItem, :count).by(1)
             end
           end
-          context 'with existing pending order for artwork' do
+          context 'with existing pending buy now order for artwork' do
             let!(:order) do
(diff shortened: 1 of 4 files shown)


spec/controllers/api/requests/offers/create_offer_order_with_artwork_mutation_request_spec.rb
```diff
@@ -250,9 +250,25 @@
...
           end
-          context 'with existing pending order for artwork' do
+          context 'with existing pending buy now order for artwork' do
             let!(:order) do
-              order = Fabricate(:order, buyer_id: jwt_user_id, state: Order::PENDING)
+              order = Fabricate(:order, buyer_id: jwt_user_id, state: Order::PENDING, mode: Order::BUY)
+              order.line_items = [Fabricate(:line_item, artwork_id: artwork_id)]
+              order
+            end
+            it 'creates a new offer order' do
```
williardx: Would be good to assert that the mode is `offer` here.

ashkan18:
thanks both! merging 🙌

PR #361, opened 2019-01-03
Update reminder emails to 5hrs before expiration

ashkan18:
# Problem
We want to switch Pulse emails to use Exchange notification instead of queuing their own job in the future for reminder emails. https://github.com/artsy/pulse/pull/325

# Solution
Currently reminder notifications are posted only 2hrs before expiration and only consumer is APRb for these for slack notification. Emails need to be sent 5hrs before expiration. After talking to CRT we are good with getting internal notifications at 5hr too.
We updated reminder notification to 5hrs and this way pulse could switch to these events.

# What happens to existing queued jobs?
For existing ones we still use 2hrs since they were already queued, if we want to not send emails 2hrs before for existing jobs we can follow this:
- Deploy Exchange with this change
- Wait 2 days (so all jobs queued in 2 days from the deploy time are finished)
- Merge https://github.com/artsy/pulse/pull/325 
- Delete current jobs queued in Pulse for reminder emails
- Deploy Pulse

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -3,7 +3,7 @@ class Order < ApplicationRecord
 
   SUPPORTED_CURRENCIES = %w[USD].freeze
 
-  INITIAL_EXPIRATION_REMINDER = 5.hours
+  DEFAULT_EXPIRATION_REMINDER = 5.hours
 
   MODES = [
     BUY = 'buy'.freeze,
@@ -176,8 +176,8 @@ class Order < ApplicationRecord
     last_offer&.awaiting_response_from
   end
 
-  def initial_reminder_time
-    state_expires_at - INITIAL_EXPIRATION_REMINDER
+  def state_expiration_reminder_time(time_to_expiration = DEFAULT_EXPIRATION_REMINDER)
+    state_expires_at - time_to_expiration
   end
 
   private
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -3,6 +3,8 @@ class Order < ApplicationRecord
 
   SUPPORTED_CURRENCIES = %w[USD].freeze
 
+  INITIAL_EXPIRATION_REMINDER = 5.hours
+
   MODES = [
     BUY = 'buy'.freeze,
     OFFER = 'offer'.freeze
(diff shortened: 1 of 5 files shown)


app/models/order.rb
```diff
@@ -174,6 +176,10 @@ def awaiting_response_from
     last_offer&.awaiting_response_from
   end
 
+  def initial_reminder_time
+    state_expires_at - INITIAL_EXPIRATION_REMINDER
+  end
```
starsirius: This could be useful if we want to send multiple reminders for the same expiration event, e.g. 5 hours before a pending order expires and then 2 hours before expiration again if it's still pending.

A different reminder time might be useful for different [state expirations](https://github.com/artsy/exchange/blob/4e33c444b27e750e6d30563112b8c32bada58bbe/app/models/order.rb#L50-L54). For example, a pending order will expire in 2 days and we will remind people 5 hours before that; an approved order will expire in 7 days  and we will remind people 24 hours before that, etc.

Also, we might want to use different reminder time for different audiences, e.g. buyer, seller, Artsy admins, etc.

Looks like none of the ☝️ support is prioritize now, so maybe we can start from something naive, like

```ruby
DEFAULT_EXPIRATION_REMINDER = 5.hours

def state_expiratrion_reminder_time(time_to_expiration = DEFAULT_EXPIRATION_REMINDER)
  state_expires_at - time_to_expiration
end
```

ashkan18: good point, going to update!

app/services/order_submit_service.rb
```diff
@@ -23,6 +23,6 @@ def post_process!
     super
     PostOrderNotificationJob.perform_later(@order.id, Order::SUBMITTED, @user_id)
     OrderFollowUpJob.set(wait_until: @order.state_expires_at).perform_later(@order.id, @order.state)
-    ReminderFollowUpJob.set(wait_until: @order.state_expires_at - 2.hours).perform_later(@order.id, @order.state)
+    ReminderFollowUpJob.set(wait_until: @order.initial_reminder_time).perform_later(@order.id, @order.state)
```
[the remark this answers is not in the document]
ashkan18: oh interesting to basically say `ReminderFollowUpJob.queue`? Im thinking having the time on the order may make things more flexible in the future, but maybe we can revisit later when we added second reminder.

ashkan18:
@starsirius this should be ready

PR #363, opened 2019-01-03
[Active Admin] Order Details Changes

ashkan18:
# Changes
Add offer support to admin order details page
https://artsyproduct.atlassian.net/browse/SELL-1281

# Selfie

image

### DIFF
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -1,0 +1,7 @@
+module AdminHelper
+  def format_money_cents(amount_cents)
+    return 'N/A' if amount_cents.blank?
+
+    number_to_currency(amount_cents.to_f / 100, unit: '$', negative_format: '( %u%n )')
+  end
+end
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -1,7 +1,8 @@
 module AdminHelper
-  def format_money_cents(amount_cents)
+  def format_money_cents(amount_cents, negate: false)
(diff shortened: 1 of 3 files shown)


config/initializers/active_admin.rb
```diff
@@ -294,3 +294,8 @@
 end
 
 ActiveAdmin::ViewHelpers.include UrlHelper
```
joeyAghion: I actually prefer this form over the re-opening style used below. Either way, you don't need both.

app/helpers/admin_helper.rb
```diff
@@ -0,0 +1,7 @@
+module AdminHelper
+  def format_money_cents(amount_cents)
+    return 'N/A' if amount_cents.blank?
+
+    number_to_currency(amount_cents.to_f / 100, unit: '$', negative_format: '( %u%n )')
```
joeyAghion: Hopefully we prevent negative values elsewhere?

config/initializers/active_admin.rb
ashkan18: oh meant to remove this line, i can add the other one this style then

app/helpers/admin_helper.rb
ashkan18: when showing some number like transaction fee and commission fee we want to show them as negative numbers apparently, this is to support that, otherwise all numbers should be positive in db.

ashkan18:
Thanks! Addressed CR, going to merge 🙌

PR #364, opened 2019-01-03
Moves undeduct inventory work into separate Sidekiq job

ashfurrow:
This PR addresses [PURCHASE-718](https://artsyproduct.atlassian.net/browse/PURCHASE-718), which moves the call to undeduct inventory into its own Sidekiq job that will run independently of the Stripe call to refund the charge.

@williardx and I looked for ways that Gravity might fail but couldn't find any. Nevertheless, we have set up a Sentry Slack integration to notify the #bnmo-pde channel if an undeduct inventory job fails.

### DIFF
--- a/app/jobs/undeduct_inventory_job.rb
+++ b/app/jobs/undeduct_inventory_job.rb
@@ -1,8 +1,0 @@
-class UndeductInventoryJob < ApplicationJob
-  queue_as :default
-
-  def perform(order_id)
-    order = Order.find(order_id)
-    order.line_items.each { |li| Gravity.undeduct_inventory(li) }
-  end
-end
--- a/app/jobs/undeduct_inventory_job.rb
+++ b/app/jobs/undeduct_inventory_job.rb
@@ -1,0 +1,8 @@
+class UndeductInventoryJob < ApplicationJob
+  queue_as :default
+
+  def perform(order_id)
+    order = Order.find(order_id)
+    order.line_items.each { |li| Gravity.undeduct_inventory(li) }
+  end
+end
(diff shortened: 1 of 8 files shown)


app/services/order_cancellation_service.rb
```diff
@@ -7,8 +7,9 @@ def initialize(order, user_id = nil)
 
   def seller_lapse!
     @order.seller_lapse! do
-      process_refund if @order.mode == Order::BUY
+      process_stripe_refund if @order.mode == Order::BUY
```
ashkan18: nice! 🙌

app/jobs/undeduct_inventory_job.rb
```diff
@@ -0,0 +1,8 @@
+class UndeductInventoryJob < ApplicationJob
+  queue_as :default
+
+  def perform(order_id)
```
ashkan18: what do you think about have this job only undeduct one line item? The reason for that would be, imagine if an order has multiple line items, if the second undeduct fails, this job is going to be retired by sidekiq and we will end up undeducting the first line item twice. 
I'm thinking of something like
```ruby
def perform(line_item_id)
  line_item = LineItem.find(line_item_id)
  Gravity.undeduct_inventory(li)
end
```
and then when queuing 
```ruby
order.line_items.each { |li| UndeductInventoryJob.perform_later(li.id) }
```
Maybe then we can rename it to `UndeductLineItem` too to make it more obvious what it expects as input.

let me know what you think.

ashkan18:
Nice! thanks! merge on green

PR #365, opened 2019-01-04
Ignores deleted artworks during inventory undeduction

ashfurrow:
[Discussed in Slack](https://artsy.slack.com/archives/CB110C6LE/p1546616701076600?thread_ts=1546616137.073300&cid=CB110C6LE), this PR is a follow-up to #364 and [PURCHASE-718](https://artsyproduct.atlassian.net/browse/PURCHASE-718). This PR changes calls to gravity's `inventory` endpoint to undeduct inventory (typically from a refund). If the API call fails due to a 404 (a deleted artwork) then it won't be retried. I looked for other places we call this undeduction, and I don't see any issues with not retrying.

### DIFF
--- a/app/jobs/undeduct_line_item_inventory_job.rb
+++ b/app/jobs/undeduct_line_item_inventory_job.rb
@@ -1,8 +1,13 @@
 class UndeductLineItemInventoryJob < ApplicationJob
   queue_as :default
+  discard_on(Errors::ValidationError) do |job, _error|
+    Rails.warn.info "Line item #{job.line_item.id} has deleted artwork, skipping inventory undeduction"
+  end
+
+  attr_accessor :line_item
 
   def perform(line_item_id)
-    line_item = LineItem.find(line_item_id)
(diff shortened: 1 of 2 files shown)


ashkan18:
Thanks for quick fix! Another alternative could be using [`discard_on`](https://api.rubyonrails.org/v5.2.1/classes/ActiveJob/Exceptions/ClassMethods.html) in the job itself, we could say 
```ruby
discard_on ErrorsValidationError do |_job, error|
  Logger.warn(...)
end
```
Looking at `undeduct` usages, your current approach should work too but i wonder if there are cases that we want to raise error in the future. We can always follow up if we needed to raise error.

ashfurrow:
Yeah, let's move to `discard_on` to be more idiomatic/explicit. I'll switch to `Logger.warn`, too.

ashfurrow:
I switched to `discard_on`, but let me add a test for it too.

ashkan18:
Thanks! 🙌

PR #368, opened 2019-01-09
New way of doing deploy events

saolsen:
Lets you see events for any deploys, not just circle ones.


hokusai/config.yml
```diff
@@ -2,3 +2,4 @@
 project-name: exchange
 pre-deploy: 'bundle exec rake db:migrate'
 git-remote: git@github.com:artsy/exchange.git
+post-deploy: .circleci/deploy_event.sh
```
ashkan18: 👍

PR #374, opened 2019-01-17
spec for making sure we don't undeduct inventory if validation fails

ashkan18:
# Source of Fear
Currently we rescue from validation errors during commit process, which looking at the code at the first look it seemed we are then undeducting the inventory without deducting it

https://github.com/artsy/exchange/blob/f30f86c4a7cb7605611a43d8495a760aace6ffca/app/services/commit_order_service.rb#L20

# What?
This is not going to happen since on `undeduct_inventory` method we will only try to call Gravity for the line items that we have deducted inventory on them.
https://github.com/artsy/exchange/blob/f30f86c4a7cb7605611a43d8495a760aace6ffca/app/services/commit_order_service.rb#L43
In case of validation error, the list of deducted inventory will be empty so we are safe

# Still!
Added a spec to verify ☝️


sweir27:
Haha, nice!

PR #375, opened 2019-01-17
Fix migration and log weird cases

ashkan18:
follow up on #354 where migration failed updating line items.


ashkan18:
merge on green

PR #377, opened 2019-01-18
Expose PageCursor fields on orders endpoint for artsy pagination component.

erikdstock:
This is a WIP attempt to match the [Pagination relay-connected component interface](https://github.com/artsy/reaction/blob/master/src/Components/v2/Pagination.tsx#L244-L264) we are currently consuming from gravity (via metaphysics) in exchange's orders connections. This is a start, but doesn't support querying by page yet and might contain anti-patterns.

The current approach leans heavily on the graphql-ruby documentation, particularly:
- http://graphql-ruby.org/api-doc/1.8.13/GraphQL/Types/Relay/BaseConnection.html  < This is the class we are extending in `PageableConnection`
- http://graphql-ruby.org/api-doc/1.8.13/GraphQL/Relay/RelationConnection.html  < This is the `object` inside a `PageableConnection`
- http://graphql-ruby.org/relay/connections.html  < This is documentation that may point in the direction of a better way to do it


app/graphql/types/pagination/pageable_connection_interface.rb
```diff
@@ -0,0 +1,69 @@
...
+  field :count_after, Int, 'number of objects after the last cursor in this page', null: false
+
+
+
+  def nodes_before(node)
+    object.nodes.take_while { |n| n.cursor != node.cursor }
+  end
+
+  def nodes_after(node)
+    object.nodes.drop_while { |n| n.cursor != node.cursor }.drop(1)
```
erikdstock: i'm sure there's a better way to check equality- maybe `==` but it's not mentioned in the source.

app/graphql/types/pagination/pageable_connection_interface.rb
```diff
@@ -0,0 +1,69 @@
...
+  ######## WIP ######
+  # Missing implementations/data: items per page, total pages, efficient method for getting the "around" pages
+  ITEMS_PER_PAGE = "??????????"
+  PageCursor = Struct.new(:cursor, :isCurrent?, :page) # temporary, not even sure if this will work, probably should have some kind of service class that does this for us?
+
+  ## Known things
+  # self.object should be a BaseConnection: https://www.rubydoc.info/gems/graphql/GraphQL/Relay/BaseConnection
+  # object.nodes is all nodes, object.edge_nodes is the nodes in edges on this page
+  # being that this is pagination we should have first, last, (Ints) before and after (cursors) available
+
```
erikdstock: I've mostly been working off the graphql-ruby gem docs here: https://www.rubydoc.info/gems/graphql/GraphQL/Relay/BaseConnection#edge_nodes-instance_method

app/graphql/types/order_connection_with_total_count_type.rb
```diff
@@ -2,7 +2,25 @@ class Types::OrderEdgeType < GraphQL::Types::Relay::BaseEdge
...
+class PageCursorType < Types::BaseObject
+  field :cursor, String, 'first cursor on the page', null: false
+  field :isCurrent, Boolean, 'is this the current page?', null: false
+  field :page, Int, 'page number out of totalPages', null: false
+end
+
+class PageCursorsType < Types::BaseObject
+  field :first, PageCursorType, 'optional, may be included in field around', null: true
+  field :last, PageCursorType, 'optional, may be included in field around', null: true
+  field :around, [PageCursorType], null: true
```
mzikherman: This shouldn't be null (based on the MP implementation), since this will always contain the 'current' page (even in the degenerate case of an empty collection).

app/graphql/types/pagination/pageable_connection.rb
```diff
@@ -0,0 +1,109 @@
...
+  #   end
+  # end
+
+  def nodes_before
+    object.send(:offset_from_cursor, object.cursor_from_node(object.edge_nodes.first)) - 1
+  end
+
+  def nodes_after
+    object.nodes.size - object.send(:offset_from_cursor, object.cursor_from_node(object.edge_nodes.last))
+  end
```
erikdstock: new approach to determining page#s by looking at the private `offset_from_cursor` method.

app/graphql/types/pagination/pageable_connection.rb
```diff
@@ -0,0 +1,109 @@
...
+
+  # A cursor for querying a given page number (after: cursor_for_page(8), first: nodes_per_page)
+  def cursor_for_page(page_num)
+    if page_num > 1 ## page 1 has no cursor
+      after_cursor = (page_num - 1) * nodes_per_page
+      object.encode(after_cursor.to_s)
+    else
+      nil
+    end
```
erikdstock: New approach to generating cursors which circumvents using `object.cursor_from_node(node)` which errors for cursors that are not part of the edge (but under the hood, uses the 'offset' number.

app/graphql/types/pagination/pageable_connection_interface.rb
erikdstock: [update] This file is not used for now.

spec/controllers/api/requests/orders_pagination_query_request_spec.rb
```diff
@@ -0,0 +1,224 @@
...
+        expect(current_page.data.orders.page_cursors.first.page).to eq 1
+        expect(current_page.data.orders.page_cursors.last).to be_nil
+        expect(current_page.data.orders.page_cursors.previous.page).to eq 7
+        expect(current_page.data.orders.page_cursors.around.count).to eq 4
+        expect(current_page.data.orders.page_cursors.around.first.is_current).to be false
+        expect(current_page.data.orders.page_cursors.around.first.page).to eq 7
+        expect(current_page.data.orders.page_cursors.around[1].is_current).to be true
+        expect(current_page.data.orders.page_cursors.around[1].page).to eq 8
+        expect(current_page.data.orders.page_cursors.around.last.is_current).to be false
+        expect(current_page.data.orders.page_cursors.around.last.page).to eq 10
```
damassi: looking good @erikdstock

app/graphql/types/order_connection_with_total_count_type.rb
iskounen: @mzikherman can you confirm that your comment refers to the `around` field and not the whole `page_cursor` object described above?

app/graphql/types/pagination/pageable_connection.rb
```diff
@@ -0,0 +1,80 @@
...
+  def around_page_numbers
+    if total_pages <= MAX_CURSOR_COUNT
+      (1..total_pages).to_a
+    elsif current_page <= 3
+      (1..4).to_a
+    elsif current_page >= total_pages - 2
+      ((total_pages - 3)..total_pages).to_a
+    else
+      [current_page - 1, current_page, current_page + 1]
+    end
```
alloy: lol, I love how we made it full-circle by porting (and adjusting) [this old Ruby code of mine](https://github.com/Fingertips/peiji-san/blob/6bd1bc7c152961dcde376a8bcb2ca393b5b45829/lib/peiji_san/view_helper.rb#L111-L120) to TS and then back to Ruby again 🤓

/cc @mzikherman

iskounen: Thanks for the inspiration, Peiji-san!

alloy: 🎌

damassi: Haha that's awesome :)

PR #378, opened 2019-01-22
Query for competing orders

### LINKED ISSUE #372: Expose Competing Orders
In order to work this ticket:

https://artsyproduct.atlassian.net/browse/SELL-1334

I'll need for Exchange to expose when there are better orders for a seller to pick. This only applies to MO offers and the two types of better orders are:

* a BN order at full price
* another MO offer at a higher price

In either of these situations, I'd like to be able to query for the order and get links to the better offers, something like this:

```
query($id: String!) {
  ecommerceOrder(id: $id) {
    id
    betterOrders {
      fullPriceOrder {
        id
      }
      higherPriceOffer {
        id
      }
    }
  }
}
```

So then when I'm consuming this query, I can check to see if those fields are null or not. If they are not null then we'll link to those orders in the header of the page, like so:

https://app.zeplin.io/project/5aabc6f28d835c7b2585fdf4/screen/5b72f58dd5475320522cb1dd

I actually don't even really care about anything other than the id because that's all I need in order to construct the link, but it's probably best to set ourselves up to return full Order types here so that future changes here will be cheap. For instance, I could see us wanting to show the price of higher offers. Something to think about.

Now, there are edges (and maybe others, but this is what I could think of):

* what if there are more than one BN full price orders?
* what if there are more than one MO higher price orders?

In the case of where there is more than one better order, we prefer the order placed earliest. In the case where there are more than one MO higher offers, we take the highest one and break any ties by when they were placed.

Thoughts, questions, ideas?

jonallured:
This PR exposes a new query for competing orders given an order id and seller id:

```gql
query($orderId: ID!) {
  competingOrders(orderId: $orderId) {
    totalCount
  }
}
```

There was some discussion over on #372 as to the approach for the consumer of this query, but I think this primitive lines up with whatever approach we take there.

In terms of the approach here, I've mostly copy/pasted from the orders query and then added a `Order#competing_orders` method to do the database query - open to ideas!

closes #372


app/graphql/types/query_type.rb
```diff
@@ -25,6 +25,12 @@ class Types::QueryType < Types::BaseObject
     argument :order_states, [Types::OrderStateEnum], required: false
   end
 
+  field :competingOrders, Types::OrderConnectionWithTotalCountType, null: true, connection: true do
+    description 'Find list of competing orders'
+    argument :order_id, ID, required: true
+    argument :seller_id, String, required: true
```
sepans: Do we need the `sellerId`?

jonallured:
@sepans whoa, you've blown my mind!! I think you're right - just going to read through that query again with this use case in mind and might explore with a test or two, but really nice catch!! 🏆 ❤️ 💯

app/models/order.rb
```diff
@@ -117,6 +117,12 @@ class Order < ApplicationRecord
     end
   end
 
+  def competing_orders
+    artwork_ids = line_items.pluck(:artwork_id)
+    conditions = 'orders.id != ? AND orders.state = ? AND line_items.artwork_id IN (?)'
+    Order.joins(:line_items).where(conditions, id, SUBMITTED, artwork_ids)
```
sepans: I don't exactly know how the edition sets are handled in exchange but is ☝️ query catch the competing orders on an edition set artwork?

jonallured: Yeah you're right, this should be more like:

```
artwork_ids, edition_set_ids = line_items.pluck(:artwork_id, :edition_set_id)
conditions = 'orders.id != ? AND orders.state = ? AND (line_items.artwork_id IN (?) OR line_items.edition_set_id IN (?))'
Order.joins(:line_items).where(conditions, id, SUBMITTED, artwork_ids, edition_set_ids)
```

jonallured: Ok, I've addressed this with bd32ff4.

app/graphql/types/query_type.rb
jonallured: I think we need it in order to check that we're safe to return the results, but I only copy/pasted from the way that `validate_orders_request!` works. What do you think @ashkan18?

jonallured:
Ok after much discussion with @sepans I'm back to thinking that adding this query makes sense. In my context, I've only got an order id, not an artwork id so the line items query that he mentioned isn't quite right.

app/graphql/types/query_type.rb
```diff
@@ -47,6 +53,12 @@ def line_items(args = {})
     query
   end
 
+  def competing_orders(params)
+    order = Order.find(params[:order_id])
+    validate_competing_orders_request!(params, order)
```
ashkan18: if we remove `seller_id` we can just use `validate_order_request`

app/graphql/types/query_type.rb
```diff
@@ -25,6 +25,12 @@ class Types::QueryType < Types::BaseObject
     argument :order_states, [Types::OrderStateEnum], required: false
   end
 
+  field :competingOrders, Types::OrderConnectionWithTotalCountType, null: true, connection: true do
+    description 'Find list of competing orders'
```
ashkan18: should we give option to filter by `mode`? I can imagine `buy` orders have higher priority over `offer` ones.

app/models/order.rb
```diff
@@ -117,6 +117,17 @@ class Order < ApplicationRecord
     end
   end
 
+  def competing_orders
```
ashkan18: should we use scope for this? then we can also pass in `mode` as a param if needed and it will be more obvious that it can be chained.

jonallured:
@ashkan18 yeah that's what @sepans and I were discussing - in his use case for Gravity hitting Exchange looking for line items he's blocking artwork deletion so he's got artwork/edition set ids, but in my use case, I'm drawing the order detail page and only have the order id. I think this approach makes more sense there.

app/models/order.rb
jonallured: I'm not a fan of scopes, I think they end up getting too hairy and I especially dislike them if they are complicated. I think a nice place for this business logic is a method on an order.

app/graphql/types/query_type.rb
jonallured: I considered this, but I don't currently need it so I left it out. We can always add it later if need be!

app/graphql/types/query_type.rb
jonallured: Cool, good call - I really wasn't sure what to do here!! 😝 

I'll get a commit in that validates without passing in the seller id. 👍

jonallured:
Ok with 2520f5e I've switched to only passing in an order id and using the existing order validation based on either being trusted or the context of the query. I think I've addressed all feedback but lmk if I've missed something!! ❤️

app/models/order.rb
ashkan18: i was just thinking of case someone wanted to chain the result of this and add extra where clauses like `Order.competing_orders.where(mode: ?)` but we can cross that bridge when we get there :)

app/models/order.rb
```diff
@@ -117,6 +117,17 @@ class Order < ApplicationRecord
...
+    artwork_ids = line_items.pluck(:artwork_id)
+    edition_set_ids = line_items.pluck(:edition_set_id)
+    conditions = <<~SQL
+      orders.id != ?
+      AND orders.state = ?
+      AND (line_items.artwork_id IN (?) OR line_items.edition_set_id IN (?))
+    SQL
+    Order.joins(:line_items).where(conditions, id, SUBMITTED, artwork_ids, edition_set_ids)
```
ashkan18: I still think we need some sort of "sort" here, either having the default one (`created_at: :asc` ?)  or give mutation the option of passing sorts, what do you think?

.rubocop.yml
```diff
@@ -29,7 +29,7 @@ Metrics/MethodLength:
     - "db/migrate/*"
 
 Metrics/ClassLength:
-  Max: 190
+  Enabled: false
```
jonallured: idk, maybe this is controversial, but it seems like just bumping this number up and up as I work is a waste of time - are people cool with me just disabling this sucker?

ashkan18: haha 😅 I think so, I generally agree with making classes, methods and etc short but i can't easily put a number on it!  I _think_ disabling this makes sense as long as we still look for opportunities to make shorter classes when possible 👍

jonallured:
Ok, I had a couple wacky commits there that I've force pushed to fix. I have added the default `order` clause in 4ac44e6 and then with 46c694f I regenerated the RuboCop todo file and removed the class length setting. My thinking there is that we should still have RuboCop complain about long classes but the current offenders have been listed in the todo file so that we can come back to them and clean up.

ashkan18:
🙌 thanks! lets do this

app/models/order.rb
```diff
@@ -117,6 +117,21 @@ class Order < ApplicationRecord
...
   end
 
+  def competing_orders
+    artwork_ids = line_items.pluck(:artwork_id)
+    edition_set_ids = line_items.pluck(:edition_set_id)
+    conditions = <<~SQL
+      orders.id != ?
+      AND orders.state = ?
+      AND (line_items.artwork_id IN (?) OR line_items.edition_set_id IN (?))
+    SQL
```
yuki24: Let's see if I can scope-ify this raw query...

PR #379, opened 2019-01-23
Switch to new style services for commit order

ashkan18:
# How it used to be?
We used to have a `CommitOrderService` which was responsible for doing all common logic needed to _commit_ an order, meaning handling:
1) deduct inventory
2) process the charge
3) properly undeduct inventory if any of the above failed

We need to do ☝️ in two different paths, when a buy now order is submitted, and when an offer is accepted by either side. For that to work we had two services inheriting from `CommitOrderService` and each override specific part of the code to handle their own need.

# How it is now?
We are now following our new pattern of service layer where we:
- Expose our external service calls either in `order_service` or `offer_service`. So class methods in these two reflect whats accessible from mutation/sidekiq layer.
- We moved common logic (in this case, think of the logic that used to be in `CommitOrderService`) to helper style classes in `lib`. So now we have `OrderProcessor` which gets an order, can call `charge` or `hold`, so the caller can decide if they want to capture the charge or just hold it.

# Selfie

selfie-0
[_GitHub Selfies_](https://github.com/thieman/github-selfies/)

https://artsyproduct.atlassian.net/browse/PURCHASE-794


app/graphql/mutations/base_accept_offer.rb
```diff
@@ -11,13 +11,8 @@ def resolve(offer_id:)
...
 
-    service = Offers::AcceptOfferService.new(
-      offer,
-      user_id: current_user_id
-    )
-    service.process!
-
-    { order_or_error: { order: service.offer.order.reload } }
+    OfferService.accept_offer(offer, current_user_id)
+    { order_or_error: { order: offer.order.reload } }
```
joeyAghion: I'm curious why the `reload` is necessary. If this reloads data that's been updated by inline jobs in test mode, it might result in tests behaving differently than production (when the jobs may not have processed yet).

app/services/order_service.rb
```diff
@@ -30,6 +30,34 @@ def self.set_payment!(order, credit_card_id)
...
+      )
+      order_processor.hold
+    end
+
+    PostOrderNotificationJob.perform_later(order.id, Order::SUBMITTED, user_id)
+    OrderFollowUpJob.set(wait_until: order.state_expires_at).perform_later(order.id, order.state)
+    ReminderFollowUpJob.set(wait_until: order.state_expiration_reminder_time).perform_later(order.id, order.state)
+    Exchange.dogstatsd.increment 'order.submitted'
+    order
+  end
```
joeyAghion: There's some repetition or similarity between this and `OfferService.accept_offer`. I don't have any great ideas [that wouldn't make it more confusing] though.

One difference I'm confused about is how this one calculates commissions on line items. Does the other require that too? (Both are getting updated totals.)

spec/lib/order_processor_spec.rb
```diff
@@ -0,0 +1,260 @@
...
+      context 'missing shipping info' do
+        before do
+          order.update!(fulfillment_type: nil)
+        end
+        it 'raises validation error' do
+          expect { order_processor.hold }.to raise_error do |e|
+            expect(e.code).to eq :missing_required_info
+          end
+        end
+      end
```
joeyAghion: A bunch of these are very nested (with `context`s, `before`s, etc.) when they could just be a single 2-3 line spec.

app/services/order_service.rb
ashkan18: Yes, we can revisit if we ended up repeating the repeated code :) in general this may make it more obvious on what happens after submission but 👍 on keeping an 👁 on the length of this method.

ashkan18:
merging this to do some testings on staging

PR #382, opened 2019-01-28
Add offer notes field to exchange

ansor4:
Signed-off-by: Sepand Ansari <sepans@sepans.com>

For https://artsyproduct.atlassian.net/browse/GALL-1366

- Added new `offer_note` text field to Offer model
- Altered function signatures in OfferService to pass in new `offer_note` param
- Added optional `offerNote` param in `buyer_counter_offer`, `seller_counter_offer`, and `add_initial_offer_to_order` graphql endpoints
- Updated respective specs to test optional `offer_note` params


db/migrate/20190128182354_add_notes_field_to_offer.rb
```diff
@@ -0,0 +1,5 @@
+class AddNotesFieldToOffer < ActiveRecord::Migration[5.2]
+  def change
+    add_column :offers, :offer_note, :text
```
sweir27: Given that this field is on the `offers` table, can we just name it `note`?

ashkan18: yes to ☝️

joeyAghion:
I don't feel strongly that we need to encrypt these `note`s at the field level. From a product perspective it's actually important that these notes _not_ include contact or payment details, so we should do what we can to discourage that. Other fields (like phone/address--actual PII) strike me as better candidates for special care.

sweir27:
If we're good not encrypting this field, is this good to go? cc @ashkan18 @ansor4 😄

ashkan18:
Yes 🙌 lets do this 🕺

PR #383, opened 2019-01-28
Use SQL subqueries to reduce extra network roundtrips

yuki24:
In ActiveRecord, almost all `#pluck` calls could be replaced with `#select`. This will consolidate the three SQL queries sent individually and reduce extra network roundtrip and memory footprint.

## Before:

```sql
# First query
SELECT "line_items"."artwork_id" FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", "..."]]

 # Second query
SELECT "line_items"."edition_set_id" FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", "..."]]

SELECT *
FROM "orders"
INNER JOIN "line_items" ON "line_items"."order_id" = "orders"."id"
WHERE orders.id != '...'
  AND orders.state = 'submitted'
  AND (
    line_items.artwork_id IN ("...", "...") OR
    line_items.edition_set_id IN ("...", "...")
  )
```

## After:

```sql
SELECT *
FROM "orders" 
INNER JOIN "line_items" ON "line_items"."order_id" = "orders"."id" 
WHERE orders.id != '...'
  AND orders.state = 'submitted'
  AND (
    line_items.artwork_id IN (
      SELECT "line_items"."artwork_id" FROM "line_items" WHERE "line_items"."order_id" = '...'
    ) OR
    line_items.edition_set_id IN (
      SELECT "line_items"."edition_set_id" FROM "line_items" WHERE "line_items"."order_id" = '...'
    )
  )
) 
```

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -118,8 +118,9 @@ class Order < ApplicationRecord
   end
 
   def competing_orders
-    artwork_ids = line_items.pluck(:artwork_id)
-    edition_set_ids = line_items.pluck(:edition_set_id)
+    artwork_ids = line_items.select(:artwork_id)
+    edition_set_ids = line_items.select(:edition_set_id)
+
     conditions = <<~SQL
       orders.id != ?
       AND orders.state = ?


jonallured:
@yuki24 interesting! When I was doing #378, I was trying to reduce the SQL statements down but couldn’t think of a good way. I'm still a little unclear about how this PR reduces the number of requests to the database down to one. Is `select` lazy in some way that allows AR to combine things? Is this documented somewhere?

yuki24:
If the `where` method receives a `Relation` object it will wrap it as a subquery. The `#select` method is defined on the `ActiveRecord::Relation` object, and when it's not given a block it'll return a lazy `Relation` object so it's not going to load any data. If it is given a block, it will load the data and delegate the call to `Enumerable#select`.

It's not [well-documented in the official API doc](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-select), but it's blogged at here: https://medium.com/rubyinside/active-records-queries-tricks-2546181a98dd

PR #384, opened 2019-01-28
Use more relation methods to make it (maybe) a bit easier to understand

yuki24:
I'm not sure if any of you like this but in case you do...

 * `orders.id != ?` could be replaced with `.where.not(id: ...)`
 * `orders.state = ?` could be replaced with `.where(state: SUBMITTED)`

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -121,15 +121,11 @@ class Order < ApplicationRecord
     artwork_ids = line_items.select(:artwork_id)
     edition_set_ids = line_items.select(:edition_set_id)
 
-    conditions = <<~SQL
-      orders.id != ?
-      AND orders.state = ?
-      AND (line_items.artwork_id IN (?) OR line_items.edition_set_id IN (?))
-    SQL
-
     Order
       .joins(:line_items)
-      .where(conditions, id, SUBMITTED, artwork_ids, edition_set_ids)
+      .where.not(id: id)
+      .where(state: SUBMITTED)
+      .where('(line_items.artwork_id IN (?) OR line_items.edition_set_id IN (?))', artwork_ids, edition_set_ids)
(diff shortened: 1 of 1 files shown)


yuki24:
And there's another one that removes the SQL heredoc: https://github.com/artsy/exchange/commit/2221786d1c004bc91858b42e20454ec87467d578

jonallured:
I don't have a strong preference here, but I didn't find the SQL hard to understand so I'd tend to leave it as-is. OTOH I did write it, so I'm biased. 😉

app/models/order.rb
```diff
@@ -122,14 +122,14 @@ def competing_orders
...
-      AND (line_items.artwork_id IN (?) OR line_items.edition_set_id IN (?))
+      (line_items.artwork_id IN (?) OR line_items.edition_set_id IN (?))
     SQL
 
     Order
       .joins(:line_items)
-      .where(conditions, id, SUBMITTED, artwork_ids, edition_set_ids)
+      .where.not(id: id)
+      .where(state: SUBMITTED)
+      .where(conditions, artwork_ids, edition_set_ids)
```
ashkan18: if we go this route, i'd say we can even move `conditions` to its own where and do something like
```ruby
.where.not(id: id)
.where(state: SUBMITTED)
.where('(line_items.artwork_id IN (?) OR line_items.edition_set_id IN (?))', artwork_ids, edition_set_ids)
```

if we want to keep `conditions` in this approach then i would rename it to `artwork_matching_condition` or something more descriptive.

yuki24:
I just applied @ashkan18's suggestion and pushed it. @jonallured @ashkan18 Mind reviewing it again?

ashkan18:
👍 thanks

jonallured:
Thanks!

PR #385, opened 2019-01-29
Switch to bigint for all cents fields

ashkan18:
# Problem
We are using `int` for all of our amount fields, that limits us to max of 2,147,483,647 cents or $21,474,836.47 
We want to use `bigint` just in case that day comes....

https://artsyproduct.atlassian.net/browse/PURCHASE-771

# Solution
Add an irreversible migration for switching all amount fields in different models to bigint.

# Selfie

### DIFF
--- a/db/migrate/20190128180430_switch_to_bigger_int.rb
+++ b/db/migrate/20190128180430_switch_to_bigger_int.rb
@@ -1,0 +1,33 @@
+class SwitchToBiggerInt < ActiveRecord::Migration[5.2]
+  def up
+    change_table :orders, bulk: true do |t|
+      t.change :shipping_total_cents, :bigint
+      t.change :tax_total_cents, :bigint
+      t.change :transaction_fee_cents, :bigint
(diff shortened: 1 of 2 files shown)


sweir27:
It will come, have no fear. Props for discovering this early 👏 👏

db/migrate/20190128180430_switch_to_bigger_int.rb
```diff
@@ -0,0 +1,33 @@
...
+      t.change :shipping_total_cents, :bigint
+      t.change :tax_total_cents, :bigint
+      t.change :transaction_fee_cents, :bigint
+      t.change :commission_fee_cents, :bigint
+      t.change :items_total_cents, :bigint
+      t.change :buyer_total_cents, :bigint
+      t.change :seller_total_cents, :bigint
+    end
+
+    change_table :line_items, bulk: true do |t|
```
sweir27: I'm surprised this `bulk: true` isn't the default

PR #386, opened 2019-01-30
Add Buyer Reject button to admin actions

ashkan18:
Add Buyer Reject action to order admin page 

# Selfie
image


joeyAghion:
1. It's amazing how simple this was.

2. Is there better language than "buyer reject" for this operation? I know that's the language we use internally, but I don't think it's clear to those who might not be familiar.

ashkan18:
Thanks @joeyAghion we just got the second case of this in 1 day, going to merge and deploy and use the opportunity to confirm it on production

PR #387, opened 2019-01-31
Adds note to offer event

sweir27:
This includes the offer note in the event that we send to rabbit.

The length of the note is constrained client-side, so we're not likely to get ones > 200 characters.


ashkan18:
I considered your reasonable offer 👍 so merge on green

PR #389, opened 2019-01-31
Add back sales admin privs

ansor4:
Exact revert of https://github.com/artsy/exchange/pull/280

Tests pass. This is the exchange side of https://artsyproduct.atlassian.net/browse/GALL-1369

See the volt/cms side here: https://github.com/artsy/volt/pull/3427


ashkan18:
nice and easy 🙌

PR #390, opened 2019-02-09
WIP/Experiment ⚠️ Integrates Prettier

ashfurrow:
This PR is an experiment to try running Prettier's new [Ruby plugin](https://github.com/prettier/plugin-ruby) on an Artsy codebase; the goal is to provide feedback to the Prettier and plugin developers on their way to a 1.0 Ruby plugin. I'm going to take a closer look and leave some inline comments, and then send this PR over to them. I encourage others to leave feedback in this PR, too!

---

When running `./node_modules/.bin/prettier --write --plugin=@prettier/plugin-ruby **/*.rb`, I ran into crashes with the following files:

- [`lib/tax/collection_service.rb`](https://github.com/artsy/exchange/blob/fb7b2fca569fc6e5163937240c0e8da81bc1eee3/lib/tax/collection_service.rb)
- [`lib/order_shipping.rb`](https://github.com/artsy/exchange/blob/fb7b2fca569fc6e5163937240c0e8da81bc1eee3/lib/order_shipping.rb)
- [`lib/order_creator.rb`](https://github.com/artsy/exchange/blob/fb7b2fca569fc6e5163937240c0e8da81bc1eee3/lib/order_creator.rb)

The stack traces are all the same, so here's just one:

```
➜  ~/bin/exchange git:(prettier-ruby) ✗ ./node_modules/.bin/prettier --write --plugin=@prettier/plugin-ruby lib/order_shipping.rb
lib/order_shipping.rb
[error] lib/order_shipping.rb: TypeError: Cannot read property '0' of undefined
[error]     at Object.module.exports [as print] (/Users/ashfurrow/bin/exchange/node_modules/@prettier/plugin-ruby/src/print.js:29:11)
[error]     at callPluginPrintFunction (/Users/ashfurrow/bin/exchange/node_modules/prettier/bin-prettier.js:13590:18)
[error]     at /Users/ashfurrow/bin/exchange/node_modules/prettier/bin-prettier.js:13541:16
[error]     at Object.printComments (/Users/ashfurrow/bin/exchange/node_modules/prettier/bin-prettier.js:13264:17)
[error]     at printGenerically (/Users/ashfurrow/bin/exchange/node_modules/prettier/bin-prettier.js:13540:22)
[error]     at FastPath.call (/Users/ashfurrow/bin/exchange/node_modules/prettier/bin-prettier.js:13378:16)
[error]     at makeCall (/Users/ashfurrow/bin/exchange/node_modules/@prettier/plugin-ruby/src/utils.js:12:62)
[error]     at Object.call (/Users/ashfurrow/bin/exchange/node_modules/@prettier/plugin-ruby/src/nodes.js:208:32)
[error]     at Object.module.exports [as print] (/Users/ashfurrow/bin/exchange/node_modules/@prettier/plugin-ruby/src/print.js:8:32)
[error]     at callPluginPrintFunction (/Users/ashfurrow/bin/exchange/node_modules/prettier/bin-prettier.js:13590:18)
```

The crash occurs on [this line](https://github.com/prettier/plugin-ruby/blob/fbb666e4b784130d47e4eb7cc346dab80fddee0a/src/print.js#L29) but it's not immediately obvious to me why. I'm going to open an issue on their repo with details.


app/admin/admin_notes.rb
```diff
@@ -4,9 +4,14 @@
...
-      f.input :admin_id, input_html: {readonly: true, value: current_user[:id]}
-      f.input :note_type, :as => :select, :collection => AdminNote::TYPES.collect { |type| [type[1].humanize, type[0]] }
+      f.input :order_id,
+      as: :string, input_html: { readonly: true, value: order.id }
+      f.input :admin_id,
+      input_html: { readonly: true, value: current_user[:id] }
+      f.input :note_type,
+      as: :select,
+      collection:
+        AdminNote::TYPES.collect { |type| [type[1].humanize, type[0]] }
```
orta: These in particular feel strange - I guess we don't have a line length option?

app/admin/orders.rb
```diff
@@ -4,23 +4,43 @@
...
+  end
+  scope('Case Still Open') do |scope|
+    scope.by_last_admin_note(AdminNote::TYPES.except(:case_closed).values)
+  end
 
   filter :id_eq, label: 'Order Id'
-  filter :mode, as: :check_boxes, collection: proc { Order::MODES }, label: 'Type'
+  filter :mode,
+         as: :check_boxes, collection: proc { Order::MODES }, label: 'Type'
   filter :code_eq, label: 'Order Code'
```
orta: Yeah, I bet this is flipping on 80 chars. Bane of my life. We should try make this 120.

app/admin/orders.rb
```diff
@@ -34,80 +54,97 @@
...
+    OrderCancellationService.new(resource, resource.buyer_id).reject!(
+      Order::REASONS[Order::CANCELED][:buyer_rejected]
+    )
+    redirect_to resource_path, notice: 'Rejected on behalf of buyer!'
   end
 
   member_action :approve_order, method: :post do
     OrderApproveService.new(resource, current_user[:id]).process!
-    redirect_to resource_path, notice: "Order approved!"
+    redirect_to resource_path, notice: 'Order approved!'
```
orta: I wonder if we can prefer double quotes like we do

app/admin/orders.rb
```diff
@@ -170,49 +210,45 @@
...
-            last_admin_note.note_type.humanize
-          end
+          last_admin_note.note_type.humanize if last_admin_note.present?
         end
         row 'Last note' do |order|
-          if last_admin_note.present?
-            last_admin_note.description
-          end
+          last_admin_note.description if last_admin_note.present?
         end
```
orta: I agree with all of these ^

app/admin/orders.rb
```diff
@@ -254,59 +288,70 @@
...
           format_money_cents order.total_list_price_cents
         end
-        row "Accepted Offer" do |_order|
-          format_money_cents order.items_total_cents
-        end if order.mode == Order::OFFER
+        if order.mode == Order::OFFER
+          row 'Accepted Offer' do |_order|
+            format_money_cents order.items_total_cents
+          end
+        end
```
orta: also think I prefer this, single expression postfix-ifs work when it's a single LOC, but harder to grok IMO when multiline like this

app/controllers/api/health_controller.rb
```diff
@@ -6,10 +6,15 @@ class HealthController < Api::BaseApiController
...
-      }
+               api_healthy: true,
+               worker_health:
+                 if sidekiq_default_queue_size < SIDEKIQ_ALERT_THRESHOLD
+                   'OK'
+                 else
+                   'OY'
+                 end,
+               worker_default_queue_size: sidekiq_default_queue_size
+             }
```
orta: I prefer the old way personally

app/graphql/types/application_error_type.rb
```diff
@@ -11,24 +11,26 @@ def self.from_application(err)
...
+      data: { message: err.message },
+      message: err.message
+    )
   end
 
   # For root level errors expect a message key and our detail data should be under extension
   def self.format_root_level_error(type:, code:, data: nil, message: nil)
     message ||= "type: #{type}, code: #{code}, data: #{data}"
+
     {
```
orta: interesting that it adds new lines 👍

app/graphql/types/date_type.rb
```diff
@@ -3,12 +3,12 @@ class Types::DateType < Types::BaseScalar
   DATE_FORMAT = '%Y-%m-%d'.freeze
 
   def self.coerce_input(input_value, _context)
-    # Parse the incoming object into a DateTime
     Date.strptime(input_value, DATE_FORMAT)
   end
+  # Parse the incoming object into a DateTime
```
orta: This is a bug /cc @kddeisz

app/graphql/types/offer_order_type.rb
```diff
@@ -5,19 +5,26 @@ class Types::OfferOrderType < Types::BaseObject
...
   end
 
   def my_last_offer
     return unless context[:current_user][:id]
 
-    object.offers.where(creator_id: context[:current_user][:id]).order(created_at: :desc).first
+    object.offers.where(creator_id: context[:current_user][:id]).order(
+      created_at: :desc
+    )
+      .first
```
orta: This feels weird

app/graphql/types/query_type.rb
```diff
@@ -81,32 +93,45 @@ def sales_admin?
...
-    return if trusted? || sales_admin? ||
-              (order.buyer_type == Order::USER && order.buyer_id == context[:current_user][:id]) ||
-              (order.seller_type != Order::USER && context[:current_user][:partner_ids].include?(order.seller_id))
+    if trusted? || sales_admin? ||
+      (order.buyer_type == Order::USER &&
+        order.buyer_id == context[:current_user][:id]) ||
+      (order.seller_type != Order::USER &&
+        context[:current_user][:partner_ids].include?(order.seller_id))
+      return
+    end
```
orta: The before does read better, but I bet this is a line-length thing, so it'd probably need to be made into vars for readablilty

app/models/admin_note.rb
```diff
@@ -10,5 +10,6 @@ class AdminNote < ApplicationRecord
     execution_contacted_seller: 'execution_contacted_seller'.freeze,
     execution_initiated_refund: 'execution_initiated_refund'.freeze,
     case_closed: 'case_closed'.freeze
-  }.freeze
+  }
+    .freeze
```
orta: I wonder if `.freeze` could be special-cased to live on the same line as it's object if it's the only expression after?

app/services/order_approve_service.rb
```diff
@@ -9,8 +9,13 @@ def initialize(order, user_id)
...
     @order.approve! do
-      @transaction = PaymentService.capture_authorized_charge(@order.external_charge_id)
-      raise Errors::ProcessingError.new(:capture_failed, @transaction.failure_data) if @transaction.failed?
+      @transaction =
+        PaymentService.capture_authorized_charge(@order.external_charge_id)
+      if @transaction.failed?
+        raise Errors::ProcessingError.new(
+                :capture_failed, @transaction.failure_data
+              )
+      end
```
orta: probably line length, but the before reads really well, and after does less so

but the switch from a single postfix-if on the raise feels a bit off

app/services/order_service.rb
```diff
@@ -1,30 +1,51 @@
...
+      raise Errors::ValidationError.new(:invalid_state, state: order.state)
+    end
 
     credit_card = Gravity.get_credit_card(credit_card_id)
-    raise Errors::ValidationError.new(:invalid_credit_card, credit_card_id: credit_card_id) unless credit_card.dig(:user, :_id) == order.buyer_id
+    unless credit_card.dig(:user, :_id) == order.buyer_id
+      raise Errors::ValidationError.new(
+              :invalid_credit_card, credit_card_id: credit_card_id
+            )
+    end
```
orta: same with all these, maybe `raise` and `if`/`unless` could be fine for single expressions?

config/initializers/active_admin.rb
```diff
@@ -1,10 +1,7 @@
 ActiveAdmin.setup do |config|
-  # == Site Title
   #
-  # Set the title that is displayed on the main layout
-  # for each of the active admin pages.
-  #
-  config.site_title = 'Exchange'
```
orta: Likely a bug, deleting comments

config/initializers/datadog.rb
```diff
@@ -5,8 +5,16 @@
   hostname = ENV.fetch('DATADOG_TRACE_AGENT_HOSTNAME', 'localhost')
   debug = ENV['DATADOG_DEBUG'] == 'true'
 
-  c.tracer enabled: enabled, hostname: hostname, distributed_tracing: true, debug: debug
-  c.use :rails, service_name: 'exchange', distributed_tracing: true, controller_service: 'exchange.controller', cache_service: 'exchange.cache', database_service: 'exchange.postgres'
+  c.tracer enabled: enabled,
+  hostname: hostname,
+  distributed_tracing: true,
+  debug: debug
```
orta: in other cases I've seen these params being indented, but they're not here

config/initializers/datadog_statsd.rb
```diff
@@ -3,7 +3,7 @@ module StatsdDisabled
     attr_accessor :disabled
 
     def send_to_socket(message)
-      print message # rubocop:disable Rails/Output
+      print message
```
orta: Deleted comment

db/schema.rb
```diff
@@ -11,217 +11,252 @@
...
+  create_table 'offer_versions', force: :cascade do |t|
+    t.string 'item_type', null: false
+    t.uuid 'item_id', null: false
+    t.string 'event', null: false
+    t.string 'whodunnit'
+    t.jsonb 'object'
+    t.jsonb 'object_changes'
+    t.datetime 'created_at'
+    t.index ['item_type', 'item_id'],
+    name: 'index_offer_versions_on_item_type_and_item_id'
```
orta: this might be the same issue about indenting the params too

spec/controllers/api/requests/fulfill_at_once_request_spec.rb
```diff
@@ -77,49 +79,64 @@
...
           response = client.execute(mutation, fulfill_at_once_input)
-          expect(response.data.fulfill_at_once.order_or_error.order.id).to eq order.id.to_s
-          expect(response.data.fulfill_at_once.order_or_error.order.state).to eq 'FULFILLED'
-          response.data.fulfill_at_once.order_or_error.order.line_items.edges.each do |li|
+          expect(
+            response.data.fulfill_at_once.order_or_error.order.id
+          ).to eq order.id.to_s
+          expect(
+            response.data.fulfill_at_once.order_or_error.order.state
+          ).to eq 'FULFILLED'
```
orta: again, probably line length, but these don't work as well

ashfurrow: Yeah. There is a configurable line length but it defaults to 80.

config/initializers/datadog.rb
ashfurrow: Yeah it seems inconsistent.

app/admin/orders.rb
ashfurrow: We can, I'll re-run it with double quotes.

config/routes.rb
```diff
@@ -2,7 +2,10 @@
 Rails.application.routes.draw do
   ActiveAdmin.routes(self)
   mount ArtsyAuth::Engine => '/'
-  constraints ->(req) { req.session[:access_token].present? && ArtsyAuthToken.new(req.session[:access_token]).admin? } do
+  constraints lambda do |req|
+                req.session[:access_token].present? &&
+                  ArtsyAuthToken.new(req.session[:access_token]).admin?
+              end do
```
ashfurrow: This is no longer valid syntax and is causing a test failure on CI.

app/admin/orders.rb
ashfurrow: Yeah this is really nice.

app/models/admin_note.rb
ashfurrow: Oh yeah, that would be nice.

app/admin/orders.rb
```diff
@@ -1,316 +1 @@
...
-        column :created_at
-        column "Note Type" do |admin_note|
-          admin_note.note_type.to_s.humanize
-        end
-        column :description
-      end
-    end
-  end
-end
+:active
```
ashfurrow: This file was almost entirely deleted when I re-ran with printWidth at 120 and preferSingleQuotes set to false.

kddeisz:
Wow this is so great - thanks so much for using the plugin and being open to helping out! I'm checking out all of these failures and will post back when I make progress.

app/admin/orders.rb
kddeisz: Wow... I have no idea why it would do that, but it's no longer doing it with v0.3.3. Hilarious.

app/controllers/api/health_controller.rb
kddeisz: Yeah I do too. Going to see if I can find a way to make it do that.

app/events/order_event.rb
```diff
@@ -23,28 +23,30 @@ class OrderEvent < Events::BaseEvent
...
+    state
     state_reason
     state_expires_at
     tax_total_cents
     transaction_fee_cents
     updated_at
     total_list_price_cents
-  ].freeze
+  ]
+    .freeze
```
kddeisz: Eww, going to look into this.

app/graphql/types/date_type.rb
kddeisz: Nice catch (https://github.com/prettier/plugin-ruby/pull/59/files)

config/initializers/active_admin.rb
kddeisz: Yup that's a bug - https://github.com/prettier/plugin-ruby/pull/60

config/initializers/datadog_statsd.rb
kddeisz: https://github.com/prettier/plugin-ruby/pull/61

config/routes.rb
```diff
@@ -1,15 +1 @@
-require 'sidekiq/web'
```
kddeisz: .... wat.

This no longer does this on master.

db/migrate/20181009163308_add_commission_rate_to_order.rb
```diff
@@ -1,7 +1,5 @@
 class AddCommissionRateToOrder < ActiveRecord::Migration[5.2]
   def change
-    change_table :orders, bulk: true do |t|
-      t.column :commission_rate, :float
-    end
+    change_table :orders, bulk: true do |t| t.column :commission_rate, :floatend
```
kddeisz: https://github.com/prettier/plugin-ruby/pull/65

kddeisz:
Okay so I think I've addressed most of the issues in here. I'm going to cut a new release - if you could rerun once it's live that'd be really great!

orta:
💟

yuki24:
Wow, very impressed by the amount of energy involved here... Really appreciate all your hard work, @kddeisz @ashfurrow @orta!

ashfurrow:
@kddeisz thanks so much for looking into those! I'm going to re-apply the new version of prettier to the current master, and then force-push, to get an accurate picture of what it would look like for a fresh use.

ashfurrow:
Okay, I've pushed those changes up and confirmed the crashes are fixed – awesome work on a quick fix, @kddeisz! I also timed the process, which took 32.616s for 282 files.

I've resolved the outdated conversations as well, to tidy up the thread.

app/controllers/api/graphql_controller.rb
```diff
@@ -5,10 +5,11 @@ def execute
       query = params[:query]
       operation_name = params[:operationName]
       context = {
-        # Query context goes here, for example:
-        current_user: current_user,
-        user_agent: request.headers['User-Agent'],
-        user_ip: request.headers['x-forwarded-for']
+        current_user:
+          # Query context goes here, for example:
+          current_user,
```
ashfurrow: This is really interesting – I don't think this is intended.

app/models/order.rb
```diff
@@ -101,7 +94,13 @@ class Order < ApplicationRecord
   scope :pending, -> { where(state: PENDING) }
   scope :active, -> { where(state: [Order::APPROVED, Order::SUBMITTED]) }
   scope :approved, -> { where(state: APPROVED) }
-  scope :by_last_admin_note, ->(note_types) { where('(SELECT note_type FROM admin_notes WHERE order_id = orders.id ORDER BY created_at DESC limit 1) in (?)', note_types) }
+  scope :by_last_admin_note,
+        lambda do |note_types|
```
ashfurrow: Still seeing a test failure here from the lambda syntax:

```
ArgumentError: tried to create Proc object without a block
```

app/admin/orders.rb
ashkan18: i agree too, i think originally to make 🤖 👮  happy, we basically ignored this file :)

app/controllers/api/graphql_controller.rb
ashkan18: yes, i think we still need the comment on the top, maybe correct place would be even before `context =` 

on a non-related note, we can just remove this comment, this was auto-generated by graphql code generator thingy

app/controllers/api/health_controller.rb
ashkan18: yes, we had to allow the old style specifically in Rubocop too, I also prefer the old style :)

app/graphql/concerns/request_validator.rb
```diff
@@ -8,23 +8,29 @@ def authorize_buyer_request!(item)
   end
 
   def authorize_offer_owner_request!(offer)
-    raise Errors::ValidationError, :not_found unless context[:current_user]['id'] == offer.from_id && offer.from_type == Order::USER
+    unless context[:current_user]["id"] == offer.from_id && offer.from_type == Order::USER
```
ashkan18: this is more readable too 👍

app/graphql/concerns/request_validator.rb
```diff
@@ -8,23 +8,29 @@ def authorize_buyer_request!(item)
...
 
   def authorized_seller?(item)
     order =
       case item
-      when Order then item
-      when Offer then item.order
+      when Order
+        item
+      when Offer
+        item.order
```
ashkan18: i kind of prefer the one line with `then` what everyone else thinks?

app/graphql/mutations/add_initial_offer_to_order.rb
```diff
@@ -3,14 +3,21 @@ class Mutations::AddInitialOfferToOrder < Mutations::BaseMutation
...
+  field :order_or_error, Mutations::OrderOrFailureUnionType, "A union of success/failure", null: false
 
   def resolve(order_id:, amount_cents:, note: nil)
     order = Order.find(order_id)
     authorize_buyer_request!(order)
     raise Errors::ValidationError, :cannot_offer unless order.state == Order::PENDING
 
-    OfferService.create_pending_offer(order, amount_cents: amount_cents, note: note, from_id: current_user_id, from_type: Order::USER, creator_id: current_user_id)
+    OfferService.create_pending_offer(
+      order,
```
ashkan18: 💯

app/graphql/types/query_type.rb
ashkan18: yes i agree, this code either way seems not easy to read, maybe this is trying to tell us we need to have the last two conditions in specific methods too so something like 
```ruby
return if trusted? || sales_admin? || buyer? || partner_admin?
```

app/graphql/types/query_type.rb
```diff
@@ -106,7 +109,8 @@ def validate_orders_request!(params)
   end
 
   def validate_competing_orders_request!(order)
-    not_submitted_error = Errors::ValidationError.new(:order_not_submitted, message: 'order id belongs to order not submitted')
+    not_submitted_error =
+      Errors::ValidationError.new(:order_not_submitted, message: "order id belongs to order not submitted")
```
ashkan18: maybe we should increase the line length limit?

on a separate note, looking at this method looks like we can just do something like

```ruby
raise Error::ValidationError, :order_not_submitted unless order.state == Order::SUBMITTED
```
for the whole method, not sure we are create an instance of error beforehand and possibly not even using it.

app/jobs/offer_respond_reminder_job.rb
```diff
@@ -4,9 +4,9 @@ class OfferRespondReminderJob < ApplicationJob
   def perform(order_id, offer_id)
     order = Order.find(order_id)
     offer = Offer.find(offer_id)
-    return unless order.state == Order::SUBMITTED &&
-                  Time.now <= order.state_expires_at &&
-                  order.last_offer.id == offer_id
+    unless order.state == Order::SUBMITTED && Time.now <= order.state_expires_at && order.last_offer.id == offer_id
```
ashkan18: i think old style may be little more readable.

app/jobs/order_follow_up_job.rb
```diff
@@ -11,10 +11,10 @@ def perform(order_id, state)
     when Order::SUBMITTED
       cancel_submitted_order(order)
     when Order::APPROVED
-      # Order was approved but has not yet fulfilled,
-      # post an event so we can contact partner
-      OrderEvent.post(order, 'unfulfilled', nil)
+      OrderEvent.post(order, "unfulfilled", nil)
     end
+    # Order was approved but has not yet fulfilled,
+    # post an event so we can contact partner
```
ashkan18: i think updated location of this comment doesn't make sense anymore, right?

app/models/order.rb
ashkan18: i don't mind the old style here, is the change because of the line length limit?

zephraph:
@ashkan18 there's a `.prettierignore` file that works like `.gitignore`.

app/admin/orders.rb
```diff
@@ -181,42 +181,32 @@
       else
         h3 "Failed to fetch partner info"
       end
+      #TODO - handle multiple partner_locations properly, instead of just taking the first.
```
zephraph: It's really weird how it relocates comments

app/controllers/api/graphql_controller.rb
zephraph: Yeah, generally looks like there are all around issues w/ handling comments still.

app/controllers/api/health_controller.rb
zephraph: Wait... why? It seems weird to have one item not aligned. Is there some legibility thing that I'm missing?

app/controllers/concerns/error_handler.rb
```diff
@@ -28,7 +44,7 @@ def self.included(clazz)
 
   def log_exception(exception)
     Rails.logger.error(exception.message)
-    trace = exception.backtrace.join("\n")
+    trace = exception.backtrace.join("undefined")
```
zephraph: uuuummmmmm lol. I feel like this is a bug.

app/controllers/errors/error_types.rb
```diff
@@ -64,9 +64,7 @@ module Errors
...
       undeduct_inventory_failure
     ],
-    internal: %i[
-      generic
-      gravity
-    ]
-  }.freeze
+    internal: %i[generic gravity]
+  }
+    .freeze
```
zephraph: I like everything up until the freeze. I think something similar to this was commented somewhere else though.

app/graphql/concerns/request_validator.rb
zephraph: I agree!

app/graphql/concerns/request_validator.rb
```diff
@@ -8,23 +8,29 @@ def authorize_buyer_request!(item)
...
   end
 
   def authorized_buyer?(item)
     case item
-    when Order then context[:current_user]['id'] == item.buyer_id
-    when Offer then context[:current_user]['id'] == item.order.buyer_id
+    when Order
+      context[:current_user]["id"] == item.buyer_id
+    when Offer
+      context[:current_user]["id"] == item.order.buyer_id
```
zephraph: Same here as the comment above... the one line then feels a little more legible.

app/graphql/types/query_type.rb
zephraph: I don't think we should increase it any more. As is, the line length is just enough  to not wrap on inline github diffs. To me that makes doing reviews easier. Plus, longer than 120 starts making things wrap in weird ways generally. Can be hard to grok something that does too much on one line.

app/jobs/order_follow_up_job.rb
zephraph: agreed

app/models/admin_note.rb
zephraph: I know the JS version of prettier does a similar thing as this is doing. It can be useful if you have a chain of methods happening, but maybe the special case should be the _number_ of items. If it's just one, keep it on the same line. Seems like that'd be more general.

app/models/offer.rb
```diff
@@ -37,8 +37,10 @@ def awaiting_response_from
     return unless submitted?
 
     case from_participant
-    when Order::BUYER then Order::SELLER
-    when Order::SELLER then Order::BUYER
+    when Order::BUYER
```
zephraph: `when` should `then` before a new line begin

app/models/order.rb
```diff
@@ -1,91 +1,84 @@
...
 
   # For more docs about states go to:
   # https://www.notion.so/artsy/37c311363ef046c3aa546047e60cc58a?v=de68d5bbc30748f88b0d92a059bc0ba8
   STATES = [
-    PENDING = 'pending'.freeze,
+    PENDING = "pending".freeze,
     # Buyer starts checkout flow but never submits
-    ABANDONED = 'abandoned'.freeze,
+    ABANDONED =
+      "abandoned".freeze,
```
zephraph: This is interesting. Why did this wrap? `PENDING` didn't... it's definitely not a line length thing.

zephraph: Hmmm... maybe because pending is the first item in the array?

app/models/order.rb
```diff
@@ -1,91 +1,84 @@
...
-    USER = 'user'.freeze,
-    PARTNER = 'partner'.freeze
-  ].freeze
+  PARTY_TYPES = [USER = "user".freeze, PARTNER = "partner".freeze].freeze
 
   REMINDER_EVENT_VERB = {
-    pending_approval: 'pending_approval'.freeze,
-    pending_fulfillment: 'pending_fulfillment'.freeze
-  }.freeze
+    pending_approval: "pending_approval".freeze, pending_fulfillment: "pending_fulfillment".freeze
```
zephraph: I'm pretty sure the javascript version of prettier will respect new lines in objects if you have multiple objects. It might be worth having a similar functionality in ruby. I just think the original here was cleaner and easier to scan at a glance.

app/models/order.rb
zephraph: To me this new layout is a bit easier to read, granted the failure is no bueno.

app/models/order.rb
```diff
@@ -121,11 +120,9 @@ def competing_orders
...
     edition_set_ids = line_items.select(:edition_set_id)
 
-    Order
-      .joins(:line_items)
-      .where.not(id: id)
-      .where(state: SUBMITTED)
-      .where('(line_items.artwork_id IN (?) OR line_items.edition_set_id IN (?))', artwork_ids, edition_set_ids)
+    Order.joins(:line_items).where.not(id: id).where(state: SUBMITTED).where(
+      "(line_items.artwork_id IN (?) OR line_items.edition_set_id IN (?))", artwork_ids, edition_set_ids
+    )
```
zephraph: Hmmm... I think the original was cleaner. Again I feel like there's a thing in JS where if there are multiple things chained on new lines that it'll respect that.

app/models/order.rb
```diff
@@ -196,13 +192,15 @@ def state_expiration_reminder_time(time_to_expiration = DEFAULT_EXPIRATION_REMIN
   private
 
   def state_reason_inclusion
-    errors.add(:state_reason, "Current state not expecting reason: #{state}") if state_reason.present? && !REASONS.key?(state)
-    errors.add(:state_reason, 'Invalid state reason') if REASONS[state] && !REASONS[state].value?(state_reason)
+    if state_reason.present? && !REASONS.key?(state)
+      errors.add(:state_reason, "Current state not expecting reason: #{state}")
+    end
+    errors.add(:state_reason, "Invalid state reason") if REASONS[state] && !REASONS[state].value?(state_reason)
```
zephraph: This is so much better ❤️

app/models/transaction.rb
```diff
@@ -1,18 +1,11 @@
...
+  has_paper_trail class_name: "PaperTrail::TransactionVersion"
 
   belongs_to :order
 
-  TYPES = [
-    HOLD = 'hold'.freeze,
-    CAPTURE = 'capture'.freeze,
-    REFUND = 'refund'.freeze
-  ].freeze
+  TYPES = [HOLD = "hold".freeze, CAPTURE = "capture".freeze, REFUND = "refund".freeze].freeze
```
zephraph: Similar to my comment on objects, I think arrays that have multiple items and with a new-line separating them should be respected.

app/services/stripe_webhook_service.rb
```diff
@@ -18,12 +18,23 @@ def process!
...
   def process_refund_event
     order = Order.find_by(external_charge_id: @event.data.object.id)
-    raise Errors::ProcessingError.new(:unknown_event_charge, event_id: @event.id, charge_id: @event.data.object.id) unless order
+    unless order
+      raise Errors::ProcessingError.new(:unknown_event_charge, event_id: @event.id, charge_id: @event.data.object.id)
+    end
     return if order.state == Order::REFUNDED # ignore if it's already refunded
-    raise Errors::ProcessingError.new(:received_partial_refund, event_id: @event.id, charge_id: @event.data.object.id) unless @event.data.object.refunded
+    unless @event.data.object.refunded
+      raise Errors::ProcessingError.new(:received_partial_refund, event_id: @event.id, charge_id: @event.data.object.id)
```
zephraph: This is the only line that I've seen that goes to 120 and it shows up as off the screen in the diff by 3 characters.

config/initializers/datadog.rb
```diff
@@ -1,14 +1,19 @@
...
-  c.use :graphql, service_name: 'exchange.graphql', schemas: [ExchangeSchema]
-  c.use :redis, service_name: 'exchange.redis'
-  c.use :sidekiq, service_name: 'exchange.sidekiq'
-  c.use :http, service_name: 'exchange.http', distributed_tracing: true
+  c.use :rails,
+  service_name: "exchange",
+  distributed_tracing: true,
+  controller_service: "exchange.controller",
+  cache_service: "exchange.cache",
+  database_service: "exchange.postgres"
```
zephraph: It it possible to have some visual distinction for this group of args to show that it's associated with the `c.use`? I don't know if it'd be valid to use white spaces here, but I can see where this kind of thing would get hard to ready fairly quick. If this is the only way to make it work then having a newline after the last argument would help create a visual separation.

lib/tax/collection_service.rb
```diff
@@ -44,17 +38,13 @@ def construct_tax_params(args = {})
...
-        }
-      end,
+      nexus_addresses:
+        @seller_nexus_addresses.map do |ad|
+          { country: ad.country, zip: ad.postal_code, state: ad.region, city: ad.city, street: ad.street_line1 }
+        end,
       shipping: UnitConverter.convert_cents_to_dollars(@effective_shipping_total_cents)
-    }.merge(args)
+    }
+      .merge(args)
```
zephraph: So this is a place, similar to the `.freeze` examples above where if it were just one chained option it would make sense to leave it how the original is. If there are multiple it'd make sense to have them similar to this.

zephraph:
I love this so much! ❤️

app/admin/orders.rb
anandaroop: Hmm, I find this reads very strangely. 

To me eyes, the result conflates the `row` from the ActiveAdmin DSL and the logic of pulling something out of the `order` instance. I feel this actually reads much better in the original, esp if you are skimming the entire file.

app/controllers/concerns/error_handler.rb
anandaroop: Oh yeah, good spot! 😄

app/events/offer_event.rb
```diff
@@ -1,19 +1,14 @@
 class OfferEvent < Events::BaseEvent
-  TOPIC = 'commerce'.freeze
-  ACTIONS = [
-    SUBMITTED = 'submitted'.freeze,
-    PENDING_RESPONSE = 'pending_response'.freeze
-  ].freeze
+  TOPIC = "commerce".freeze
+  ACTIONS = [SUBMITTED = "submitted".freeze, PENDING_RESPONSE = "pending_response".freeze].freeze
```
anandaroop: Brrrr ❄️

app/graphql/concerns/request_validator.rb
anandaroop: Either probably fine in _this_ case — but note that Order and Offer have the same number of chars. If there were many cases and the conditions had widely different line lengths, I think the prettier version might be more readable/skimmable.

app/graphql/mutations/create_offer_order_with_artwork.rb
```diff
@@ -1,28 +1,34 @@
...
+  argument :find_active_or_create,
+           Boolean,
+           "When set to false, we will create a new order. Otherwise if current user has pending/submitted orders on same artwork/edition with same quantity, we will return that",
+           required: false, default_value: true
 
-  field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure. If find_active_or_create is not false, it will return existing pending/submitted order for current user if exists, otherwise it will return newly created order', null: false
+  field :order_or_error,
+        Mutations::OrderOrFailureUnionType,
+        "A union of success/failure. If find_active_or_create is not false, it will return existing pending/submitted order for current user if exists, otherwise it will return newly created order",
+        null: false
```
anandaroop: I guess corrections like these are unavoidable if we're enforcing a line length.

app/graphql/types/base_field.rb
```diff
@@ -7,8 +7,10 @@ def initialize(buyer_only: false, seller_only: false, **kwargs, &block)
 
   def authorized?(obj, ctx)
     case obj
-    when Order then order_field_authorized?(obj, ctx[:current_user])
-    when LineItem then order_field_authorized?(obj.order, ctx[:current_user])
+    when Order
+      order_field_authorized?(obj, ctx[:current_user])
+    when LineItem
+      order_field_authorized?(obj.order, ctx[:current_user])
```
anandaroop: This kind of illustrates my comment above.

app/jobs/offer_respond_reminder_job.rb
anandaroop: Might be another case where what the formatter is really telling us is to wrap the condition up in a var or method

app/models/offer.rb
anandaroop: I'm actually not accustomed to seeing `then` —  i think that's only for the one-liner version of `when` ?

app/models/order.rb
anandaroop: Yeah, agree

app/services/order_service.rb
anandaroop: Line length tho...

kddeisz:
There was a bug in `v0.3.6` that had a bug with escaped quotes. If you want to run it again, it'll get rid of the undefineds.

app/admin/orders.rb
```diff
@@ -112,17 +118,13 @@
...
             user_info[:name]
           end
-          row 'Location' do
-            if user_info[:location][:display].empty?
-              div "No location for user"
-            else
-              user_info[:location][:display]
-            end
+          row "Location" do
+            if user_info[:location][:display].empty?div "No location for user"elseuser_info[:location][:display]end
```
yuki24: This conditional was squeezed into a one-liner but seems a bit off here.

config/initializers/datadog.rb
alloy: I feel like idiomatic Ruby is to indent these args.

app/admin/orders.rb
ashfurrow: Agreed, it seems like they're tracking that bug.

ashfurrow:
Thanks @kddeisz! I'll upgrade and re-run today. Looks like 0.4.0 it out already, so it looks like I should use that. Let me know what else we can do to support your exciting plugin work!

app/admin/orders.rb
kddeisz: I am! This should be fixed in the 0.3.7.

app/controllers/concerns/error_handler.rb
kddeisz: lol yeah that got fixed in 0.3.7. Yikes.

app/models/offer.rb
kddeisz: Yeah I don't support `then` at all right now, we're just always getting rid of it. I wasn't sure of the correct logic for when we would want `then`.

app/models/order.rb
kddeisz: This one is being tracked on the repo.

config/initializers/datadog.rb
kddeisz: This is 100% a bug. It should be indenting but isn't.

ashfurrow:
Okay, pushed the changes from 0.4.0 of the plugin 👍

kddeisz:
@ashfurrow Could you reset your HEAD to master, run the latest version of prettier over this, and then force push this branch? Right now it's been run so many times that previous bugs are still present.

ashfurrow:
@kddeisz I thought I was – I will try again.

ashfurrow:
@kddeisz Okay, I _think_ that should do it. I `rebase -i`'d to get rid of the commit where I modified things, rebased in all latest changes from master, and re-ran the following script:

> ./node_modules/.bin/prettier --write --plugin=@prettier/plugin-ruby --print-width 120 --prefer-single-quotes false **/*.rb

Let me know what else I can do 👍

app/events/offer_event.rb
yuki24: This is one thing that Rubocop does well - probably a separate change but I think we should add `# frozen-string-literal: true` and stop worrying about freezing.

app/admin/orders.rb
kddeisz: @ashfurrow This shouldn't be happening in the latest version, that's so weird! I'm not sure why this is happening

ashfurrow: Hmm! I just `rm -rf`'d the `node_modules` directory and _am_ getting some different results. `yarn upgrade --latest` must not work the way I think it does. I'll push the latest.

ashfurrow: Still seeing this, the latest commit is here: https://github.com/artsy/exchange/commit/c6c5a9a9313ce384da3e2b1821cf4914155866c1

kddeisz: Found this, this is going to be in `0.5.1`

app/admin/orders.rb
```diff
@@ -181,42 +181,32 @@
...
-        end
+        row "Type"(&:mode)
         row :state
-        row 'Reason' do |order|
-          order.state_reason
-        end
-        row 'Last Updated At' do |order|
-          order.updated_at
-        end
+        row "Reason"(&:state_reason)
```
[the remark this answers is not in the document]
kddeisz: Found this, going to be in `v0.5.1`

app/admin/orders.rb
```diff
@@ -230,54 +220,58 @@
...
           no_credit_card_found = !credit_card_info.present?
-        rescue
+        rescue StandardError
           no_credit_card_found = true
         end
         if no_credit_card_found
-          h5 "Paid #{format_money_cents(order.buyer_total_cents)} on #{pretty_format(order[:created_at])} (Failed to get credit card info)"
+          h5 "Paid #{format_money_cents(order.buyer_total_cents)} on #{pretty_format(
+               order[:created_at]
+             )} (Failed to get credit card info)"
```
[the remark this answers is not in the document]
kddeisz: Yeah this is definitely one that even the JS prettier struggled with. I'll look into it but it's definitely going to be a future enhancement.

app/controllers/api/graphql_controller.rb
kddeisz: Found this, going to be in `v0.5.1`

app/models/order.rb
kddeisz: Funny enough this wraps because the comment gets attached to the `"abandoned".freeze` node, and then forces it to break.

config/initializers/datadog.rb
kddeisz: This will be fixed and released as part of `v0.5.1`

app/models/offer.rb
zephraph: I gotcha. I guess it's fine either way. The idea being that if it can fit on one line then just use then to bridge it. If not, then break it down how it currently is.

kddeisz:
Okay friends, status update. I've just released `v0.5.1`. This contains fixes for a lot of stuff in this PR. A couple of things that I'm still tracking:

* lambdas as arguments to commands without parens that hit the line limit break using `do...end`, which causes a syntax error because the block gets associated with the command and not the lambda. In my code I've typically made them break but used `{...}`, so I'm probably going to roll with that when I get the chance. To my knowledge this is the only bug bug that still exists.
* method chains that break are still a little weird, in that they break the final method calls first. I want to fix this to make it newline and indent after the receiver, but I haven't gotten around to it yet. This is not a bug in that the syntax is totally valid, but I agree it doesn't look as good.
* case blocks with short `then` clauses could potentially use `then` in order to shorten. I'm not sure how I feel about this one, so I'm probably going to support it as a configuration option. Y'all can then decide which one you'd like to see.
* method calls on array and hash literals that break and odd (these are the `.freeze` examples you see above). While it's not a bug per se, they're still pretty ugly and I'd like to clean that up.
* Also on my priority list is gem-ifying prettier-ruby but shipping the necessary JS assets as part of a gem and then making an executable that will shell out to node. That will hopefully make this process a lot simpler in the future and won't require knowledge of JS package managers for new ruby devs.

From what I can tell, the only actual blocker is the first one on that list, the other ones would be nice and I will get to them as a lower priority. Feel free to comment here or [open an issue on the repo](https://github.com/prettier/plugin-ruby/issues) if I missed anything.

@ashfurrow if you wouldn't mind killing the changes, rerunning with `v0.5.1`, and force-pushing, we can see if I missed anything besides the ones listed above.

ashfurrow:
Awesome work, @kddeisz! I've upgraded and pushed the latest changes.

config/boot.rb
```diff
@@ -1,3 +1,3 @@
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
+ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../Gemfile", __dir__)
 
-require 'bundler/setup' # Set up gems listed in the Gemfile.
+require
```
kddeisz: Very confused as to what happened here. Can't reproduce it either.

spec/controllers/api/requests/error_handling_request_spec.rb
```diff
@@ -1,16 +1,12 @@
...
-        artworkId: 'test'
-      }
-    end
+  describe "ErrorType" do
+    let(:auth_headers) { jwt_headers(user_id: "user-id", partner_ids: %w[p1], roles: nil) }
+    let(:mutation_input) { { artworkId: "test" } }
     let(:mutation) do
       <<-GRAPHQL
-        mutation($input: CreateOrderWithArtworkInput!) {
+              mutation($input: CreateOrderWithArtworkInput!) {
```
yuki24: This is syntactically correct, but I don't believe we want this.

lib/payment_service.rb
```diff
@@ -42,7 +60,9 @@ def self.refund_charge(charge_id)
     generate_transaction_from_exception(e, Transaction::REFUND, charge_id: charge_id)
   end
 
-  def self.generate_transaction_from_exception(exc, type, credit_card: nil, merchant_account: nil, buyer_amount: nil, charge_id: nil)
+  def self.generate_transaction_from_exception(
+    exc, type, credit_card: nil, merchant_account: nil, buyer_amount: nil, charge_id: nil
+  )
```
yuki24: I think the root cause here is the long method name which is a separate issue, but I also don't mind the original version.

db/schema.rb
```diff
@@ -25,9 +24,9 @@
     t.uuid "author_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["author_type", "author_id"], name: "index_active_admin_comments_on_author_type_and_author_id"
-    t.index ["namespace"], name: "index_active_admin_comments_on_namespace"
-    t.index ["resource_type", "resource_id"], name: "index_active_admin_comments_on_resource_type_and_resource_id"
+    t.index %w[author_type author_id], name: "index_active_admin_comments_on_author_type_and_author_id"
+    t.index %w[namespace], name: "index_active_admin_comments_on_namespace"
+    t.index %w[resource_type resource_id], name: "index_active_admin_comments_on_resource_type_and_resource_id"
```
yuki24: We may want a way to exclude auto-generated files like this.

ashfurrow: Good idea, I'll add that to `.prettierignore` on the next pass 👍

config/boot.rb
ashfurrow: Hmm, that's strange. I made sure to delete the `node_modules` directory first this time to avoid using an outdated version.

app/models/offer.rb
alloy: I wouldn’t want the two variants to be mixed (not sure if prettier has support for backtracking)

alloy:
I personally would vote for _no_ configuration option for `then`.

My favorite feature of prettier is no more discussions over each formatting style. So if no satisfying rubrics can be thought of then I prefer going with a single case that works for all situations, especially for things that don’t occur often. For the rare cases, where a file gets so dense that artisanal formatting makes sense, we can still do so by making prettier ignore it.

kddeisz:
Hi friends! It's been a busy two months, but with it being rubykaigi, I've managed to find some time and push out of a bunch of fixes. If you're still interested, would you be able to open a new PR with version `0.12.0`? I think a new PR would be good since this one is taking a super long time to load with 127 (soon to be 128) comments. Also if y'all are around at rubykaigi and want to discuss any of this I'll be here through Sunday.

ashfurrow:
This is obviated by #407 so I'm going to close, thanks @yuki24!

PR #394, opened 2019-02-12
Fix Missing Failed Transaction

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-872
When we call Stripe to process a charge (hold/capture) regardless of the success or failure, we want to store the response as  a `Transaction`. For failed transactions we also post event so we can get notified about them on Slack.
We were notified about a failed charge yesterday and noticed we haven't received slack notifications for them. Looking deeper at the issue, we noticed this failed transaction was not stored on the order.

# Cause
This bug was introduced as part of #379 . What happens here is, we had test in newly added `OrderProccessor` to check if we create a failed transaction in the order in case of Stripe charge failures BUT `order_proccessor.charge` and `order_proceessor.hold` where called in one of `order.<action>` blocks which wraps the whole thing in a database transaction. So charge failures were raising error which would eventually lead to rollback of the whole thing in the block including adding failed transaction to the order.
```ruby
order.submit! do
  # everything in this block is wrapped in database transaction
  order_processor.charge! 
 # ☝️ would add transaction to the order in case of failure but since it's wrapped in transaction
 # it will rollback
end
```

# Solution
We now let `OrderProcessor` raise newly added `Errors::FailedTransactionError` in case of failed transaction which includes the failed transaction. 
Its now up to caller to rescue from this exception and properly add transaction to the order and post events if needed.

# Manual Migration
We've been monitoring failed transactions from Stripe dashboard and as of now (Feb, 12) we had only one failed transaction. Im going to manually add that stripe transaction to the order.

### DIFF
--- a/app/services/offer_service.rb
+++ b/app/services/offer_service.rb
@@ -78,11 +78,18 @@ module OfferService
         seller_total_cents: totals.seller_total_cents
       )
       order_processor.charge
+      order.transactions << order_processor.transaction
     end
     OrderEvent.delay_post(order, Order::APPROVED, user_id)
     OrderFollowUpJob.set(wait_until: order.state_expires_at).perform_later(order.id, order.state)
     ReminderFollowUpJob.set(wait_until: order.state_expiration_reminder_time).perform_later(order.id, order.state)
     Exchange.dogstatsd.increment 'order.approved'
+  rescue Errors::FailedTransactionError => e
+    transaction = e.transaction
+    return if transaction.blank?
+
+    order.transactions << transaction
+    PostTransactionNotificationJob.perform_later(transaction.id, TransactionEvent::CREATED, user_id)
   end
 
   class << self
--- a/app/services/offer_service.rb
+++ b/app/services/offer_service.rb
@@ -77,7 +77,7 @@ module OfferService
         commission_fee_cents: totals.commission_fee_cents,
         seller_total_cents: totals.seller_total_cents
       )
-      order_processor.charge
+      order_processor.charge!
       order.transactions << order_processor.transaction
     end
     OrderEvent.delay_post(order, Order::APPROVED, user_id)
@@ -90,6 +90,7 @@ module OfferService
 
     order.transactions << transaction
     PostTransactionNotificationJob.perform_later(transaction.id, TransactionEvent::CREATED, user_id)
+    raise e
   end
 
   class << self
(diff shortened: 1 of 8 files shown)


spec/controllers/api/requests/offers/seller_accept_offer_mutation_request_spec.rb
```diff
@@ -154,14 +154,46 @@
...
+      let(:partner_account_request) { stub_request(:get, "#{Rails.application.config_for(:gravity)['api_v1_root']}/partner/#{order_seller_id}/all").to_return(status: 200, body: gravity_v1_partner.to_json) }
+      before do
+        deduct_inventory_request
+        merchant_account_request
+        credit_card_request
+        artwork_request
+        partner_account_request
+      end
+
+      context 'with failed stripe charge' do
```
sweir27: Mind updating the spacing in this file? 😁

ashkan18:
I can imagine it will be something like
```ruby
stripe_transaction_id = <find this from stripe>
stripe_failure_code = <find this from stripe>
stripe_failure_message = <:point_up:>
order.transaction.create!(external_id: stripe_transaction_id, failure_code: stripe_failure_code, failure_message: stripe_failure_message, status: Transaction::FAILURE, created_at: <date of transaction>, amount_cents: order.items_total_cents, source_id: <user_credit_card>, destination_id: <partner's merchant account id>)
```

app/services/offer_service.rb
```diff
@@ -77,12 +77,20 @@ def self.accept_offer(offer, user_id)
...
     end
     OrderEvent.delay_post(order, Order::APPROVED, user_id)
     OrderFollowUpJob.set(wait_until: order.state_expires_at).perform_later(order.id, order.state)
     ReminderFollowUpJob.set(wait_until: order.state_expiration_reminder_time).perform_later(order.id, order.state)
     Exchange.dogstatsd.increment 'order.approved'
+  rescue Errors::FailedTransactionError => e
+    transaction = e.transaction
+    return if transaction.blank?
+
+    order.transactions << transaction
```
joeyAghion: Not crazy about this duplication (similarly below), but struggling to describe a better solution.


One small way to eliminate the duplication is to save the transaction in an `ensure` block, then invoke the different follow-ups (`OrderEvent` etc. if successful, or `PostTrans...` if not) via some conditional logic applied to that transaction. However surfacing the transaction (either from the exception or processor) may be awkward.

PR #395, opened 2019-02-15
Set order last_transaction_failed when an offer order is accepted

sweir27:
This PR adds a new flag, `last_transaction_failed` to record whether or not an order's last transaction is a failure. If `true`, it means the buyer needs to update their payment method or call their bank before the order can successfully be approved.

We will use this in future PR's for validation logic around when it's OK to add a new payment method to a submitted order + accept, etc..

I added a flag instead of just querying for the `order.transactions.last` because we also want to show this on the admin interface and return it in graphql responses.

### Bonus
Added a one-liner so we now redirect requests to `/` to `/admin`.

### DIFF
--- a/app/graphql/types/order_interface.rb
+++ b/app/graphql/types/order_interface.rb
@@ -20,7 +20,7 @@ module Types::OrderInterface
   field :items_total_cents, Integer, null: true, description: 'Item total in cents, for Offer Orders this field reflects current offer'
   field :last_approved_at, Types::DateTimeType, null: true
   field :last_submitted_at, Types::DateTimeType, null: true
-  field :last_transaction_failed, Boolean, null: true
+  field :last_transaction_failed, Boolean, null: true, method: :last_transaction_failed?
   field :line_items, Types::LineItemType.connection_type, null: true
   field :requested_fulfillment, Types::RequestedFulfillmentUnionType, null: true
   field :seller_total_cents, Integer, null: true, seller_only: true
@@ -74,10 +74,6 @@ module Types::OrderInterface
       precision: 2,
       strip_insignificant_zeros: true
     )
-  end
-
-  def last_transaction_failed
-    object.last_transaction_failed?
   end
 
   # Optional, see below
--- a/app/graphql/types/order_interface.rb
+++ b/app/graphql/types/order_interface.rb
@@ -20,6 +20,7 @@ module Types::OrderInterface
   field :items_total_cents, Integer, null: true, description: 'Item total in cents, for Offer Orders this field reflects current offer'
   field :last_approved_at, Types::DateTimeType, null: true
   field :last_submitted_at, Types::DateTimeType, null: true
+  field :last_transaction_failed, Boolean, null: true
   field :line_items, Types::LineItemType.connection_type, null: true
   field :requested_fulfillment, Types::RequestedFulfillmentUnionType, null: true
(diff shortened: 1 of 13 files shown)


app/services/offer_service.rb
```diff
@@ -88,7 +89,10 @@ def self.accept_offer(offer, user_id)
     transaction = e.transaction
     return if transaction.blank?
 
+    order.reload
```
sweir27: cc @joeyAghion in case this looks familiar to you.

By the time we reach the exception, we've rolled back the db from our transaction (in `order.approve! do` above). However, `order` still points to the un-rolled-back version and so running `order.update!` was causing some weird side effects.

`order.reload` fixes it, but also seems a bit weird. Is there something obvious we're missing?

app/graphql/types/order_interface.rb
```diff
@@ -75,6 +76,10 @@ def display_commission_rate
     )
   end
 
+  def last_transaction_failed
```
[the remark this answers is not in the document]
sweir27: Yeah, could definitely go either way.

sweir27:
I was thinking it would be good to check this field from reaction's side when determining whether or not to show the new payment page. It's probably not a huge perf loss to query for it each time but this seemed a little cleaner.

What do you think? I actually implemented it in the method-form first 😝 but then migrated to this. It's easy to go back!

ashkan18:
I can go eitherway, on one hand this is a very special case and might be useful to make it more obvious  and a first class citizen but on the other hand not having it as a column makes things little easier and no need for migrating existing data. Also backfilling is not going to be that complicated anyways.
Maybe now that you have the column already we can go this route

app/services/offer_service.rb
joeyAghion: Weird... you're saying a rolled-back transaction leaves you with an activerecord object with wrong attributes? I've never run into this myself and would hope the `reload` isn't necessary. Did some quick googling and there seems to have been multiple rounds of reports and fixes of this kind of thing.

If a realistic spec dictates that this is broken otherwise, I believe it.

Just for my edification, what's an example of a field that ends up in the wrong state within the rescue block?

sweir27: Yeah it was for sure surprising.

An example of an outdated field is `order.state`, which in the rescue block is `'approved'` unless you reload, in which case it's `'submitted'`.

sweir27:
@ashkan18 updated this to not persist `last_transaction_failed`. This seems cleaner for now, and we can always update later.

app/graphql/types/order_interface.rb
```diff
@@ -20,6 +20,7 @@ module Types::OrderInterface
   field :items_total_cents, Integer, null: true, description: 'Item total in cents, for Offer Orders this field reflects current offer'
   field :last_approved_at, Types::DateTimeType, null: true
   field :last_submitted_at, Types::DateTimeType, null: true
+  field :last_transaction_failed, Boolean, null: true
```
ashkan18: we can set `method` in here and remove the method definition below, so something like
```ruby
field :last_transaction_failed, Boolean, null: true, method: :last_transaction_failed?
```
https://graphql-ruby.org/fields/introduction.html#field-resolution

ashkan18: not sure if it gets accepted, but ended up also opening https://github.com/rmosolgo/graphql-ruby/pull/2119

PR #396, opened 2019-02-20
Retry payment mutation

ds300:
Backend time once again :tada:

# new mutation for retrying payment

## The name

I went for `retryAcceptOfferWithNewPayment` but totally open to other suggestions. Personally I'm not sure I like the verb 'accept' here since it's not necessarily the buyer doing the accepting... But can't think of a better name.

## Sanity checks

It only allows the charge to go through if

- The credit card belongs to the buyer
- A previous attempt to charge the buyer failed (last_transaction_failed?)
- The order is in 'SUBMITTED' state.
- The offer is the last offer


spec/controllers/api/requests/retry_accept_offer_with_new_payment_request_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+                expect(order.last_transaction_failed?).to be true
+              end
+
+              it 'undeducts inventory' do
+                client.execute(mutation, mutation_input)
+                expect(undeduct_inventory_request).to have_been_requested
+              end
+            end
+
+            it 'approves the order' do
```
sweir27: Can we also add tests for a buyer accepting an offer that the _buyer_ has made? This mutation will need to handle both cases.

app/graphql/mutations/retry_accept_offer_with_new_payment.rb
```diff
@@ -0,0 +1,28 @@
+class Mutations::RetryAcceptOfferWithNewPayment < Mutations::BaseMutation
```
ashkan18: the name `accept` here maybe little misleading in case that offer was accepted by seller. I don't have a better name though maybe we could say something like `retry_process_offer`? if we go this route i guess we could add a new `action` 
https://github.com/artsy/exchange/blob/93f5d9f544c43d320ea22541522d03db7f07211f/app/models/order.rb#L64
and possibly a service method. Not sure if its worth changing compared to current simple solution though.

app/graphql/mutations/retry_accept_offer_with_new_payment.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    # Are there any more sanity checks we can do here?
+    raise Errors::ValidationError.new(:invalid_state, state: order.state) unless
+      order.state == Order::SUBMITTED &&
+      order.last_transaction_failed? &&
+      offer.id == order.last_offer.id
+
+    order = OrderService.set_payment!(order, credit_card_id)
+
+    OfferService.accept_offer(offer, current_user_id)
```
ashkan18: maybe worth adding a comment here describing the case where seller had already attempted to accept and are now trying it from buyer side

app/graphql/mutations/retry_accept_offer_with_new_payment.rb
sweir27: I actually think `accept` makes sense! You are retrying the "approve/accept" action, it doesn't specify who initiated it.

ds300:
Thanks for the feedback, friends! :heart:

I've addressed all the things except for the renaming business. I agree that the word 'accept' doesn't feel appropriate for the case where the seller has already 'accept'ed. But also I don't think it's valuable to go fiddling with state machines at this point so ideally we can find a name for the current behaviour which everyone likes.

how about `retry_charge_with_new_payment`?

sweir27:
The reason I like `accept_offer` is because it makes it clear that the order will change state here (vs. just charging the card). Although the buyer or seller has _attempted_ to `accept` the order before this point, the order itself has not changed state so we must do that action again.

Maybe even the `retry_` is misleading since the buyer is not necessarily retrying their own. We could just say `accept_offer_with_new_payment`.

ashkan18:
i can go eitherway, said your friend with worst naming experience.

ds300:
> Maybe even the `retry_` is misleading since the buyer is not necessarily retrying their own. We could just say `accept_offer_with_new_payment`.

Yeah they're not necessarily retrying their own *accept* action but then if we call it `accept_offer_with_new_payment` it still implies that it's the buyer doing the accepting without further implying that something went wrong and needs to be fixed.

... How about `fix_failed_payment`? It only implies that a payment previously failed without implying who accepted what. It also implies the state change from 'accepted but not quite completed' to 'completed' which is what's actually happening despite the fact that our underlying model doesn't make it obvious.

ashkan18:
`fix_failed_payment` sounds good to me! very descriptive of what we are actually doing.

ds300:
🍏

ashkan18:
🙌

PR #397, opened 2019-03-11
Add payment method (include support for wire transfer)

sweir27:
We need to be able to represent orders that are paid for by methods other than `credit_card`, even if we don't have _full_ support for those payment methods just yet.

This PR adds a `payment_method` field to the `Order` model and raises errors in places where we'd attempt to charge a cc for an order that had this set.

From an admin-perspective, if we have a UI for creating orders with `wire_transfer`, they'll have to be in the "approved" or "fulfilled" state, since we don't want to unknowingly mess with inventory or availability just yet. It gets complex 😅

### DIFF
--- a/app/controllers/errors/error_types.rb
+++ b/app/controllers/errors/error_types.rb
@@ -1,7 +1,6 @@
 module Errors
   ERROR_TYPES = {
     validation: %i[
-      can_only_process_credit_cards
       cannot_offer
       cant_submit
       cannot_accept_offer
@@ -49,6 +48,7 @@ module Errors
       unknown_participant_type
       unknown_partner
       unpublished_artwork
+      unsupported_payment_method
       unsupported_shipping_location
       wrong_fulfillment_type
     ],
--- a/app/controllers/errors/error_types.rb
+++ b/app/controllers/errors/error_types.rb
@@ -1,6 +1,7 @@
 module Errors
   ERROR_TYPES = {
     validation: %i[
+      can_only_process_credit_cards
       cannot_offer
       cant_submit
       cannot_accept_offer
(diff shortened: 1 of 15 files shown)


app/controllers/errors/error_types.rb
```diff
@@ -1,6 +1,7 @@
 module Errors
   ERROR_TYPES = {
     validation: %i[
+      can_only_process_credit_cards
```
joeyAghion: "Unsupported payment method" would be a more general error to raise. This label embeds information about what's supported _today_ (and might be misleading once, e.g., ACH is supported).

sweir27:
@joeyAghion from speaking with Wendy/Caroline, individual payment methods _may_ have implications beyond simply how one is charged, so I think the transaction-level is too granular.

ashkan18:
knowing that exchange as it is, is only setup for one successful transaction per Order i think its ok to have this payment method on the `Order` level. My only concern is if changing payment method started happening more often we'd lose the _history_ of payments method on an Order currently as a first class thing. We can always find the history with audit log we are doing on the order, if we are ok with that im 👍 for having it on the `Order` for now.
I remember the same discussion came up for lewitt-api and we kind of ended up with same decision :)

app/services/order_service.rb
```diff
@@ -1,6 +1,16 @@
 module OrderService
   def self.create_with_artwork!(buyer_id:, buyer_type:, mode:, quantity:, artwork_id:, edition_set_id: nil, user_agent:, user_ip:, find_active_or_create: false)
-    order_creator = OrderCreator.new(buyer_id: buyer_id, buyer_type: buyer_type, mode: mode, quantity: quantity, artwork_id: artwork_id, edition_set_id: edition_set_id, user_agent: user_agent, user_ip: user_ip)
+    order_creator = OrderCreator.new(
```
ashkan18: 😍

db/migrate/20190311132730_add_payment_method_to_orders.rb
```diff
@@ -0,0 +1,11 @@
+class AddPaymentMethodToOrders < ActiveRecord::Migration[5.2]
+  def up
+    add_column :orders, :payment_method, :string, null: true
+    Order.update_all(payment_method: Order::CREDIT_CARD)
+    change_column :orders, :payment_method, :string, null: false
```
ashkan18: nice migration 🙌

sweir27:
cc @ashkan18 I think this is good to go now!

ashkan18:
🙌

PR #398, opened 2019-03-25
Exchange stores and returns the decline_code when a charge/hold/etc. fails

lilyfromseattle:
When a charge fails, we’re not currently capturing the decline_code from Stripe. We need this data so we can distinguish between a charge failing due to insufficient_funds vs. other failures, like do_not_honor.

See https://stripe.com/docs/declines/codes for examples.

We’ll want to store this information on a Transaction in exchange, and expose it through the mutation payload and in TransactionEvent.

### DIFF
--- a/db/migrate/20190319164840_add_decline_code_to_transactions.rb
+++ b/db/migrate/20190319164840_add_decline_code_to_transactions.rb
@@ -1,4 +1,5 @@
 class AddDeclineCodeToTransactions < ActiveRecord::Migration[5.2]
+  # decline_code is returned from stripe when a transaction is declined. It contains the reason for the card decline (i.e. insufficient_funds). 
   def change
     add_column :transactions, :decline_code, :string
   end
--- a/db/migrate/20190319164840_add_decline_code_to_transactions.rb
+++ b/db/migrate/20190319164840_add_decline_code_to_transactions.rb
@@ -1,0 +1,5 @@
+class AddDeclineCodeToTransactions < ActiveRecord::Migration[5.2]
+  def change
+    add_column :transactions, :decline_code, :string
+  end
+end
--- a/app/events/transaction_event.rb
+++ b/app/events/transaction_event.rb
@@ -39,6 +39,7 @@ class TransactionEvent < Events::BaseEvent
       },
       failure_code: @object.failure_code,
       failure_message: @object.failure_message,
+      decline_code: @object.decline_code,
       transaction_type: @object.transaction_type,
       status: @object.status
     }
--- a/app/events/transaction_event.rb
+++ b/app/events/transaction_event.rb
@@ -16,32 +16,48 @@ class TransactionEvent < Events::BaseEvent
   end
 
   def properties
-    order = @object.order
     {
-      order: {
-        id: order.id,
-        mode: order.mode,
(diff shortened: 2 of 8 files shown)


db/migrate/20190319164840_add_decline_code_to_transactions.rb
```diff
@@ -0,0 +1,5 @@
+class AddDeclineCodeToTransactions < ActiveRecord::Migration[5.2]
+  def change
+    add_column :transactions, :decline_code, :string
```
ashkan18: another alternative could have been adding a `additional_data` or `metadata` kind of field which could be a `jsonb` type of column and have `decline_code` be just an item in that json, what do you think? knowing `decline_code` is probably the only thing we're going to have there for now makes me feel we could postpone a more generic field but then `decline_code` might be too specific to how Stripe works too, so i kind of vote for more generic approach.

ashkan18:
for the rubocop thing, maybe it makes sense to take the `order` section of the event and have it as a separate method, so something like

```ruby
  def properties
    {
      failure_code: @object.failure_code,
      failure_message: @object.failure_message,
      transaction_type: @object.transaction_type,
      status: @object.status
      order: order_details,
    }
  end

  private

  def order_details
    order = @object.order
    {
      id: order.id,
      mode: order.mode,
      buyer_id: order.buyer_id,
      buyer_total_cents: order.buyer_total_cents,
      buyer_type: order.buyer_type,
      code: order.code,
      created_at: order.created_at,
      currency_code: order.currency_code,
      fulfillment_type: order.fulfillment_type,
      items_total_cents: order.items_total_cents,
      line_items: order.line_items.map { |li| line_item_detail(li) },
      seller_id: order.seller_id,
      seller_type: order.seller_type,
      state: order.state,
      state_reason: order.state_reason,
      state_expires_at: order.state_expires_at,
      updated_at: order.updated_at
    }
  end

  def line_item_detail(line_item)
    {
      price_cents: line_item.list_price_cents,
      list_price_cents: line_item.list_price_cents,
      artwork_id: line_item.artwork_id,
      edition_set_id: line_item.edition_set_id,
      quantity: line_item.quantity,
      commission_fee_cents: line_item.commission_fee_cents
    }
  end
```

db/migrate/20190319164840_add_decline_code_to_transactions.rb
sweir27: Aghh this is hard. What other data do we get back from Stripe? If any of those other fields seem useful, we should probably just store everything.

ashkan18:
talked to @sweir27 , since this is not exposed in our GraphQL endpoint, i think we are good with current approach and in the future once we started to support more types of transactions we can re-visit (possible migrate) easily.

For now i'd only suggest adding some comment about the field definition (maybe in migration?) describing what it holds and how it's going to be used.

ashkan18:
migration we ran to backfill existing failed transactions:
```ruby
ts = Transaction.where(status: "failure", failure_code: "card_declined")
maping = ts.map { |t| [t.id, Stripe::Charge.retrieve(t.external_id).outcome.reason] }
# verify ☝️ 
maping.each do |m|
  transaction = Transaction.find(m[0])
  transaction.update!(decline_code: m[1])
end
```

PR #399, opened 2019-03-26
added tests and fix for last_offer issue on transaction_event

lilyfromseattle:


spec/events/transaction_event_spec.rb
```diff
@@ -92,6 +100,8 @@
       expect(event.properties[:order][:created_at]).not_to be_nil
       expect(event.properties[:order][:line_items].count).to eq 2
       expect(event.properties[:order][:line_items]).to match_array(line_item_properties)
+      # expect(event.properties[:order][:last_offer][:from_participant]).to eq 'buyer'
```
ashkan18: was this intentional?

spec/events/transaction_event_spec.rb
```diff
@@ -15,6 +15,12 @@
       shipping_region: 'IL'
     }
   end
+
+  let(:last_offer) do
```
ashkan18: we can possibly consolidate in one line and do
```ruby
let(:last_offer) { Fabricate(:offer, order: order, amount_cents: 380) }
```
also note that whats missing here is, the offer you are creating here is not assigned to the order we are expecting.

ashkan18:
one last rubocop issue and this should be good to go.
you can also locally run `rubocop` in excahnge which can catch these failures earlier. in `exchange` root folder you can do a `rubocop` and it should do it.

PR #400, opened 2019-03-27
Post events when requested on ApplicationError

ashkan18:
# Problem
We want to get notified on slack when some special errors happened in Exchange. Mainly things that are actionable like when partner address is invalid.

# Solution
Add new `ApplicationErrorEvent` and add optional `post_event` param to exception initialize which by default is `false`. When it's set to true we post a newly added `ApplicaitonErrorEvent`.

This new event follows are existing pattern of `object`, `subject`, `verb` and `properties` but `subject` is always `nil`.

# When we are posting this event?
Currently we only request to post this event on address verification errors, but talked to @sweir27 about opening it up to all validation errors.

### DIFF
--- a/spec/controllers/errors/application_error_spec.rb
+++ b/spec/controllers/errors/application_error_spec.rb
@@ -1,0 +1,18 @@
+require 'rails_helper'
+
+describe Errors::ApplicationError do
+  describe 'post_error_event' do
+    it 'posts application_error_event when asked for' do
+      Errors::ApplicationError.new(:validation, :missing_region, { what_was_invalid: 'everything!' }, true)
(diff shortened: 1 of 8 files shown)


ashkan18:
This is ready for actual review now :)

spec/controllers/errors/application_error_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+describe Errors::ApplicationError do
+  describe 'post_error_event' do
+    it 'posts application_error_event when asked for' do
```
sweir27: #minor and unrelated but can you add some spaces before these `if`s here and in the below test file? 😁

PR #401, opened 2019-03-29
Add has_offer_note filter to admin interface

ds300:
This ActiveAdmin thing is deep magic ✨

image

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -21,6 +21,7 @@ ActiveAdmin.register Order do
   filter :fulfillment_type, as: :check_boxes, collection: proc { Order::FULFILLMENT_TYPES }
   filter :state, as: :check_boxes, collection: proc { Order::STATES }
   filter :state_reason, as: :check_boxes, collection: proc { Order::REASONS.values.map(&:values).flatten.uniq.map!(&:humanize) }
+  filter :has_offer_note , as: :check_boxes, label: 'Has Offer Note'
 
   index do
     column :code do |order|
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -263,4 +263,8 @@ class Order < ApplicationRecord
   def complete_shipping_details?
     [shipping_name, shipping_address_line1, shipping_city, shipping_country, buyer_phone_number].all?(&:present?)
   end
+
+  ransacker :has_offer_note do
+    Arel.sql('(select exists (select 1 from offers where offers.order_id = orders.id and offers.note <> \'\' and offers.submitted_at is not null))')
+  end
 end


app/admin/orders.rb
```diff
@@ -21,6 +21,7 @@
   filter :fulfillment_type, as: :check_boxes, collection: proc { Order::FULFILLMENT_TYPES }
   filter :state, as: :check_boxes, collection: proc { Order::STATES }
   filter :state_reason, as: :check_boxes, collection: proc { Order::REASONS.values.map(&:values).flatten.uniq.map!(&:humanize) }
+  filter :has_offer_note , as: :check_boxes, label: 'Has Offer Note'
```
ashkan18: very interesting, i have to refresh my mind on this magical thingy but have you tried
```ruby
filter :offers_note_present, as: :check_boxes
```
haven't tried it but based on ransack docs it may work https://github.com/activerecord-hackery/ransack/wiki/basic-searching#present

ashkan18: if this doesn't work, we can also try with defining a scope maybe? so in our model add something like

```ruby
# not tested code
scope :has_offer_note, -> { where.not(offers: { note: nil } ) } 
```

ds300: offers_note_present didn't work 😢 

```
{"errors":[{"message":"undefined method `offers_note_present_in' for Ransack::Search\u003cclass: Order, base: Grouping \u003ccombinator: and\u003e\u003e:Ransack::Search","extensions":{"code":"generic","data":{"message":"undefined method `offers_note_present_in' for Ransack::Search\u003cclass: Order, base: Grouping \u003ccombinator: and\u003e\u003e:Ransack::Search"},"type":"internal"}}]}
```

A scope would work, that's what i did at first, but the jira ticket specifically asks for a filter in the sidebar.

app/models/order.rb
```diff
@@ -263,4 +263,8 @@ def build_machine
   def complete_shipping_details?
     [shipping_name, shipping_address_line1, shipping_city, shipping_country, buyer_phone_number].all?(&:present?)
   end
+
+  ransacker :has_offer_note do
+    Arel.sql('(select exists (select 1 from offers where offers.order_id = orders.id and offers.note <> \'\' and offers.submitted_at is not null))')
```
sweir27: It seems like a bit of a smell to have this active-admin specific code in the model. Is it possible to use the scope as suggested above, but expose it in the way we need for active admin?

ds300: I didn't see anything about that in the ActiveAdmin docs. They're awfully sparse docs tho. 

Seems like ransack syntax is the only way to create filters and the only way to customise ransack logic is by appending magical ransack things to models? 🤷‍♂️ 

I agree this is smelly fwiw. Just not sure if there's a viable alternative unless folks are cool with only having a scope. Otherwise, IMO it would feel wrong to compromise on usability just because of a tiny bit of smelly code.

ds300: could add a comment above this explaining it's only for the filter in the admin interface?

sweir27: Actually-- that's a good question. I'm not sure we care about the orders that _don't_ have offer notes (seems more likely we'll just want to look at the ones that _do_). If that requirement changed could the implementation be updated?

ds300: Alas I couldn't figure out how to do a single checkbox rather than the 'yes'/'no' options.

PR #404, opened 2019-04-09
#coordinate-deploy Only pass valid addresses to taxjar #DONOTMERGE

sepans:
e.g. 'Business' and 'Sales tax nexus'

**#do_not_merge:** we need to validate partner addresses before merging this

Depends on https://github.com/artsy/gravity/pull/12274

Fixes [GALL-1453](https://artsyproduct.atlassian.net/browse/GALL-1453)

### DIFF
--- a/lib/gravity.rb
+++ b/lib/gravity.rb
@@ -33,8 +33,10 @@ module Gravity
     nil
   end
 
-  def self.fetch_partner_locations(partner_id)
-    locations = Adapters::GravityV1.get("/partner/#{partner_id}/locations?private=true")
+  def self.fetch_partner_locations(partner_id, tax_valid = true)
+    query = "/partner/#{partner_id}/locations?private=true"
+    query += '&address_type[]=Business&address_type[]=Sales tax nexus' if tax_valid
+    locations = Adapters::GravityV1.get(query)
     raise Errors::ValidationError.new(:missing_partner_location, partner_id: partner_id) if locations.blank?
 
     locations.map { |loc| Address.new(loc) }
--- a/lib/gravity.rb
+++ b/lib/gravity.rb
@@ -33,10 +33,11 @@ module Gravity
     nil
   end
 
-  def self.fetch_partner_locations(partner_id, tax_valid = true)
-    query = "/partner/#{partner_id}/locations?private=true"
-    query += '&address_type[]=Business&address_type[]=Sales tax nexus' if tax_valid
-    locations = Adapters::GravityV1.get(query)
+  def self.fetch_partner_locations(partner_id, tax_only: false)
+    query = "/partner/#{partner_id}/locations"
+    params = { private: true }
+    params = params.merge(address_type: ['Business', 'Sales tax nexus']) if tax_only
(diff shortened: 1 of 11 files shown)


lib/gravity.rb
```diff
@@ -33,8 +33,10 @@ def self.get_artwork(artwork_id)
     nil
   end
 
-  def self.fetch_partner_locations(partner_id)
-    locations = Adapters::GravityV1.get("/partner/#{partner_id}/locations?private=true")
+  def self.fetch_partner_locations(partner_id, tax_valid = true)
+    query = "/partner/#{partner_id}/locations?private=true"
+    query += '&address_type[]=Business&address_type[]=Sales tax nexus' if tax_valid
+    locations = Adapters::GravityV1.get(query)
```
starsirius: `Adapters::GravityV1.get` actually takes [params](https://github.com/artsy/exchange/blob/b6f1b5563cb7f4b4aba9310f17f8e964cddee5b1/lib/adapters/gravity_v1.rb#L5). Let's pair and refactor this to use that instead of constructing the query string, which can be error prone.

lib/gravity.rb
```diff
@@ -33,8 +33,11 @@ def self.get_artwork(artwork_id)
     nil
   end
 
-  def self.fetch_partner_locations(partner_id)
-    locations = Adapters::GravityV1.get("/partner/#{partner_id}/locations?private=true")
+  def self.fetch_partner_locations(partner_id, tax_only: false)
+    query = "/partner/#{partner_id}/locations"
```
starsirius: #minor. This can be renamed to `url`.

lib/gravity.rb
```diff
@@ -33,8 +33,11 @@ def self.get_artwork(artwork_id)
     nil
   end
 
-  def self.fetch_partner_locations(partner_id)
-    locations = Adapters::GravityV1.get("/partner/#{partner_id}/locations?private=true")
+  def self.fetch_partner_locations(partner_id, tax_only: false)
+    url = "/partner/#{partner_id}/locations"
+    params = { private: true }
+    params = params.merge(address_type: ['Business', 'Sales tax nexus']) if tax_only
```
ashkan18: are these address type values case sensitive? if they are i'd suggest maybe in Gravity we need to make them all lower case or upercase. what do you think?

starsirius: Ohh good call! This will require a migration and change in Volt and Vibrations. Maybe we can follow up properly migrate them over? Looks like the [Partner](https://github.com/artsy/gravity/blob/c2599496bd098c489bb884f41e415bf4ff6739db/app/models/domain/partner.rb#L14-L16) model also suffered the same issue. 🙈

sepans: Good point. Gravity does a check to make sure the address type parameters **exactly** match what it expects so we are safe from passing the wrong parameters and getting the wrong data: [code](https://github.com/artsy/gravity/pull/12274/files#diff-acd9190bc20845116583302fd5afd390R72 )

If we want to change it to lower case do we need to do a migration to make the values in DB all lowercase?

PR #406, opened 2019-04-19
Set up Prettier Ruby

yuki24:
This PR sets up the Prettier Ruby plugin in Exchange. Since #390 the plugin has evolved and now we can simply use the rake tasks below to run Prettier:

```
bundle exec rake prettier:install prettier
```

(`rake prettier:install` needs to be run only once)

While this feels more natural in the Ruby world, it also comes with a catch - right now a whole Node.js runtime needs to be downloaded in order to run the `prettier` rake task. It might make sense to add an option to the upstream of the Ruby plugin so that it'll use a locally available runtime if the user doesn't mind having Node.js as a dependency.

## To-dos

 * [x] Remove Rubocop
 * [ ] Figure out a way to run Prettier without having to download the whole Node.js runtime

cc @kddeisz

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,7 @@ gem 'taxjar-ruby', require: 'taxjar'
 group :development, :test do
   gem 'byebug'
   gem 'graphlient'
+  gem 'prettier', git: 'https://github.com/prettier/plugin-ruby.git'
   gem 'rspec-rails'
   gem 'rubocop'
 end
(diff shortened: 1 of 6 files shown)


kddeisz:
Yeah this is definitely something I want to fix. The gem for some reason isn't working with `pkg` just yet, but I should be able to get that working soon so you won't need a node runtime.

yuki24:
I just sent another PR to add executables to the gem itself: https://github.com/prettier/plugin-ruby/pull/277

kddeisz:
@yuki24 I just released the gem, so this process should be even simpler.

yuki24:
Closing in favor of https://github.com/artsy/exchange/pull/407.

PR #408, opened 2019-04-23
Add Accept Offer Button

ashkan18:
# Changes

Remove approve order button when order is not BN.

Add accept offer button for submitted offer orders.

fixes: https://artsyproduct.atlassian.net/browse/PURCHASE-1039

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -57,7 +57,7 @@ ActiveAdmin.register Order do
   member_action :accept_offer, method: :post do
     return unless resource.mode == Order::OFFER && resource.state == Order::SUBMITTED
     OfferService.accept_offer(resource.last_offer, current_user[:id])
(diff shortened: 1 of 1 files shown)


app/admin/orders.rb
```diff
@@ -54,6 +54,12 @@
     redirect_to resource_path, notice: "Order approved!"
   end
 
+  member_action :accept_offer, method: :post do
+    return unless resource.mode == Order::OFFER && resource.state == Order::SUBMITTED
+    OfferService.accept_offer(resource.last_offer, current_user[:id])
+    redirect_to resource_path, notice: "Order approved!"
```
ansor4: Do we want to say "Offer Accepted" here instead?

ashkan18: good point! thanks! updated

ashkan18:
merge on green

PR #411, opened 2019-05-23
[WIP] Fix not recalculating shipping fees after changing fulfillment

ashkan18:
# BUG 🐛
https://artsy.slack.com/archives/CB110C6LE/p1558627040005300

Follow 👇 going to first add a failing spec, describe the issue and then commit fixes.

### DIFF
--- a/lib/buy_order_totals.rb
+++ b/lib/buy_order_totals.rb
@@ -6,11 +6,11 @@ class BuyOrderTotals
   end
 
   def tax_total_cents
-    @tax_total_cents ||= @order.tax_total_cents || @order.line_items.map(&:sales_tax_cents).sum
+    @tax_total_cents ||= @order.line_items.map(&:sales_tax_cents).sum
   end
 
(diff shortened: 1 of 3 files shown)


ashkan18:
Added failing test and verified without fix we have 2 issues:

# After updating shipping address we don't update order's totals
## Impact
This impacts orders that new address changes shipping from domestic to international or other way around.
It also impacts orders that change their fulfillment type (pickup/shipping)

## Cause
In `BuyOrderTotal` if order already has `buyer_total_cents` set on it we won't calculate again.

# After switching fulfillment type from Ship to Pickup we don't clean shipping data
## Impact
This is less problematic than first issue but still a problem where we don't clean up shipping information, we don't show them anywhere either on pickup order so we are safe. but data-wise this is an issue.

## Cause
We don't nilify shipping info when setting fulfillment to pickup.

starsirius:
Thanks for the quick fix! 💯

PR #412, opened 2019-05-24
#FF Add end-to-end buy now happy path testing

ashkan18:
# Problem
We have test on mutation/request level, but we don't currently have test for testing a life cycle of an order. This lowers the confidence 📉 on deploys and changes. 

# Start Of a Solution
Added example buy now happy path end to end test.


spec/scenario/reset_shipping_request_spec.rb
```diff
@@ -92,75 +49,81 @@
...
+        expect(order.reload).to have_attributes(
+          shipping_country: 'IR',
+          shipping_city: 'Tehran',
+          shipping_region: 'TH',
+          shipping_postal_code: '09821',
+          buyer_phone_number: '0989121324',
+          shipping_name: 'NameF NameL',
+          shipping_address_line1: 'Vanak',
+          shipping_address_line2: 'Suite 81'
+        )
```
ds300: Nice change!

spec/controllers/api/requests/fulfill_at_once_request_spec.rb
```diff
@@ -11,62 +11,12 @@
       Array.new(2) { Fabricate(:line_item, order: order, list_price_cents: 200) }
```
joeyAghion: What does `include_context 'GraphQL Client'` (a few lines above) refer to? It's confusingly similar to `GraphQL Client Helpers` used elsewhere.

spec/scenario/buy_now_happy_path_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+    let(:seller_addresses) { [Address.new(state: 'NY', country: 'US', postal_code: '10001'), Address.new(state: 'MA', country: 'US', postal_code: '02139')] }
+    let(:seller_merchant_account) { { external_id: 'ma-1' } }
+    let(:buyer_client) { graphql_client(user_id: buyer_id, partner_ids: [], roles: 'user') }
+    let(:seller_client) { graphql_client(user_id: 'partner_admin_id', partner_ids: [seller_id], roles: 'user') }
+
+    before do
+      stub_tax_for_order(tax_amount: 100)
+      allow(Adapters::GravityV1).to receive(:get).with('/artwork/a-1').and_return(gravity_artwork)
+      allow(Gravity).to receive_messages(
+        get_artwork: gravity_artwork,
```
joeyAghion: If you stub `get_artwork` here, do you also need to stub the `GravityV1` adapter's response above?

spec/support/graphlient_client.rb
```diff
@@ -19,3 +19,16 @@
     end
   end
 end
+
+RSpec.shared_context 'GraphQL Client Helpers', shared_context: :metadata do
+  def graphql_client(user_id:, partner_ids:, roles:, user_agent: 'user-agent', user_ip: 'user-ip')
```
joeyAghion: Does this need to be a "shared context," or could it just be a helper method?

spec/support/query_helper.rb
```diff
@@ -0,0 +1,154 @@
...
+              code
+              data
+              type
+            }
+          }
+        }
+      }
+    }
+  ).freeze
+end
```
joeyAghion: I'm curious if these are declared anywhere else already... they must be in the code (though maybe in a different form), right? Is that duplication acceptable for developers, or might there be ways to avoid it?

spec/scenario/buy_now_happy_path_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+        fetch_partner_locations: seller_addresses,
+        fetch_partner: gravity_partner,
+        get_credit_card: buyer_credit_card,
+        deduct_inventory: nil,
+        get_merchant_account: seller_merchant_account
+      )
+    end
+
+    it 'succeeds with all expected values' do
```
sweir27: "Happy path" isn't super descriptive. Could we update either the `describe` block above or this to say what is actually happening?

spec/controllers/api/requests/fulfill_at_once_request_spec.rb
ashkan18: yes, that `'GraphQL Client'` shared context is the older style and im thinking of changing those to new method style where you can create client for specific user. But decided not to do that in this PR since it will add lot of extra diffs.

spec/scenario/buy_now_happy_path_spec.rb
ashkan18: oh i added new stubbed artwork one and forgot to remove the adaptor level one, removing that one.

spec/support/graphlient_client.rb
ashkan18: ideally this would have been a simple helper method but because of few lines below where we need to tell the client to use `app` which is only accessible in the `context` we could change it to a method but then caller has to pass `app` which feels little weird. for now i went with simpler context approach.

spec/support/query_helper.rb
ashkan18: we have test for each mutation which would have similar queries in there. The way i was thinking about those is, those queries are more into details and they also verify the response coming back from mutations while these current tests mostly test the state of the order.
That being said, we can still share the actual mutation/query defs in one place and depending on type of test care or don't care about response coming back.

spec/scenario/buy_now_happy_path_spec.rb
ashkan18: good point! i added more description

spec/scenario/buy_now_happy_path_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+      # seller accepts order
+      expect do
+        seller_client.execute(QueryHelper::APPROVE_ORDER, input: { id: order.id.to_s })
+      end.to change(order.transactions, :count).by(1)
+      expect(order.reload).to have_attributes(
+        state: Order::APPROVED,
+        items_total_cents: 1000_00,
+        shipping_total_cents: 200_00,
+        tax_total_cents: 100_00,
+        buyer_total_cents: 1300_00,
```
starsirius: Maybe we can also check seller total fields at this point?

spec/support/taxjar_helper.rb
```diff
@@ -42,19 +42,19 @@ def order_fixture
...
       "taxable_amount": 16.5,
-      "amount_to_collect": 1.16,
+      "amount_to_collect": tax_amount || 1.16,
       "rate": 0.07,
       "has_nexus": true,
       "freight_taxable": true,
       "tax_source": 'destination',
       "breakdown": {
-        "taxable_amount": 16.5,
+        "taxable_amount": tax_amount || 16.5,
```
starsirius: I might be confused but maybe we want to use `tax_amount` for the `tax_collectable` below, instead of `taxable_amount`? This applies to breakdowns, shipping, line items, etc. On the other hand, it may be unnecessary to introduce the complexity here if we are not actually using breakdown, etc.

spec/scenario/buy_now_happy_path_spec.rb
```diff
@@ -0,0 +1,135 @@
+require 'rails_helper'
+require 'support/gravity_helper'
+require 'support/taxjar_helper'
+
+describe Api::GraphqlController, type: :request do
```
starsirius: Looks like the scenario folder is not the convention RSpec uses for [request specs](https://relishapp.com/rspec/rspec-rails/docs/directory-structure). We can consider renaming to requests/ or integration/.

spec/support/taxjar_helper.rb
ashkan18: oh yes, good catch, we only care about `amount_to_collect` on total. I added it to breakdown to wrong field. yes, at the moment we don't use breakdown amount. if you think we need to switch we can track it somewhere.

PR #414, opened 2019-05-30
#FF Add clc_assisted to Orders and to admin

ashkan18:
# Problem
When CRT assists a buyer during checkout, they often mention in slack that this order was created with CLC's help. 
This information seems to be important to them but currently lost in Slack.

https://artsyproduct.atlassian.net/browse/PURCHASE-1147

# Solution
Add a `clc_assisted` flag to `Order` and allow to toggle it from the admin UI.

# Selfie

Screen Shot 2019-05-31 at 8 33 33 AM

Screen Shot 2019-05-31 at 8 33 42 AM

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -77,7 +77,7 @@ ActiveAdmin.register Order do
 
   member_action :toggle_assisted, method: :post do
     resource.toggle!(:assisted)
-    redirect_to resource_path, notice: "toggled CLC assisted flag!"
+    redirect_to resource_path, notice: "toggled assisted flag!"
   end
 
   action_item :refund, only: :show do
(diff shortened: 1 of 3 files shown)


joeyAghion:
I've read this PR and all the code updates, but still have no idea what "CLC" is. Maybe it could be more self-explanatory, like `assisted`?

More out-there idea: what if orders could have a list of arbitrary tags and activeadmin let admins filter by them? Maybe this kind of process improvement won't need code changes in the future.

ashkan18:
oooh `jsonb`!!! the UI for that is something i have no clue on `ActiveAdmin` but we should not be the only one doing it!

app/admin/orders.rb
```diff
@@ -74,6 +75,11 @@
     redirect_to resource_path, notice: "Fulfillment confirmed!"
   end
 
+  member_action :toggle_clc_assisted, method: :post do
+    resource.toggle!(:clc_assited)
```
sweir27: Ditto to removing `clc` (or switching to @joeyAghion 's suggestion!)

In any case this is a typo. 😄

joeyAghion:
See this PR where we fixed editing JSONB-editing in activeadmin for Horizon's `Profile#environment` field: https://github.com/artsy/horizon/pull/21

ashkan18:
Im kind of leaning towards going with the column approach for now and if we ended up doing more tagging the migration shouldn't be that hard. What do you think?

PR #419, opened 2019-06-20
Updates config to use orb

kierangillen:
- Updates `config.yml` to use hokusai orb

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,53 +1,53 @@
-version: 2
-jobs:
-  test:
-    docker:
-      - image: artsy/hokusai:0.4.4
-    steps:
-      - add_ssh_keys
-      - checkout
-      - setup_remote_docker
-      - run:
-          name: Test
-          command: hokusai test
-  push:
-    docker:
-      - image: artsy/hokusai:0.4.4
-    steps:
-      - add_ssh_keys
-      - checkout
-      - setup_remote_docker
-      - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,53 +1,53 @@
...
+version: 2.1
+
+orbs:
+  hokusai: artsy/hokusai@0.3.0
+
+not_staging_or_release: &not_staging_or_release
+  filters:
+    branches:
+      ignore:
+        - staging
```
ashkan18: we don't have `staging` or `release`. Looking at this, we may want to create them now.

.circleci/config.yml
```diff
@@ -1,53 +1,53 @@
...
+          project-name: exchange
           requires:
-            - push
+            - push-staging-image
+
+      # release
+      - hokusai/push:
+          name: push-production-image
+          <<: *only_release
+      - hokusai/deploy-production:
```
ashkan18: nice! currently we deploy using `hokusai pipeilne promote` if the plan is to switch to PR-based deploy as part of this, we should update readme https://github.com/artsy/exchange/blob/master/README.md#production

kierangillen: Just did that, thanks!

.circleci/config.yml
kierangillen: We think that github will create them automatically. We will verify.

PR #423, opened 2019-06-25
PURCHASE-1138: Adds spec for the single counteroffer happy path scenario

sweir27:
This PR adds a spec for the "happy path" scenario in which:
1. Buyer creates initial offer (and adds shipping, payment, etc.)
2. Seller counters that offer
3. Buyer accepts that offer

Since the first step here (buyer creating the offer) is validated in the "make offer happy path" spec, I had it go through the actions but I didn't repeat the assertions. Happy to revisit if you think I'm losing value here!

### DIFF
--- a/spec/integration/single_counteroffer_happy_path_spec.rb
+++ b/spec/integration/single_counteroffer_happy_path_spec.rb
@@ -1,0 +1,136 @@
+require 'rails_helper'
+require 'support/gravity_helper'
+require 'support/taxjar_helper'
+
+describe Api::GraphqlController, type: :request do
+  include_context 'use stripe mock'
+  include_context 'GraphQL Client Helpers'
+  describe 'make offer happy path' do
+    let(:seller_id) { 'gravity-partner-id' }
+    let(:buyer_id) { 'user-id1' }
+    let(:buyer_shipping_address) do
+      {
+        name: 'Fname Lname',
+        country: 'US',
+        city: 'New York',
+        region: 'NY',
+        postalCode: '10012',
+        phoneNumber: '617-718-7818',
+        addressLine1: '401 Broadway',
+        addressLine2: 'Suite 80'
+      }
+    end
+    let(:buyer_credit_card) { { id: 'cc-1', user: { _id: buyer_id }, external_id: stripe_customer.default_source, customer_account: { external_id: stripe_customer.id } } }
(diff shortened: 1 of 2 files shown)


spec/integration/single_counteroffer_happy_path_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+      # Buyer sets credit card
+      buyer_client.execute(QueryHelper::SET_CREDIT_CARD, input: { id: order.id.to_s, creditCardId: buyer_credit_card[:id] })
+
+      # Buyer submits offer order
+      buyer_client.execute(OfferQueryHelper::SUBMIT_ORDER_WITH_OFFER, input: { offerId: offer.id.to_s })
+
+      # Seller submits counteroffer
+      expect do
+        seller_client.execute(OfferQueryHelper::SELLER_COUNTER_OFFER, input: { offerId: offer.id.to_s, amountCents: 700_00 })
+      end.to change(order.transactions, :count).by(0)
```
ashkan18: maybe we can also check reation of new offer
```
to change(order.transactions, :count).by(0).and change(order.offers, :count).by(1)
```

spec/integration/single_counteroffer_happy_path_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+        transaction_fee_cents: 29_30,
+        seller_total_cents: 900_70,
+        fulfillment_type: Order::SHIP,
+        shipping_country: 'US',
+        credit_card_id: 'cc-1'
+      )
+      expect(order.transactions.order(created_at: :desc).first).to have_attributes(
+        transaction_type: Transaction::CAPTURE,
+        amount_cents: 1000_00,
+        status: Transaction::SUCCESS
```
ashkan18: maybe add 
```
source_id: a_string_starting_with('test_')
```
this is how that other bug was found when we were putting json in this field.

spec/integration/single_counteroffer_happy_path_spec.rb
sweir27: Oooh didn't know you could do two `change` assertions! Will add.

sweir27:
Updated!

ashkan18:
merge on green

PR #424, opened 2019-06-25
PURCHASE-1118 - Finishes steps for exchange hokusai updates 

kierangillen:
- Finishes exchange hokusai update steps as described in https://docs.google.com/document/d/1i0P_wOQ7cY_2AmFvy5zV5hiQg6BDgL4m9RjhKCaL33E/edit

- https://artsyproduct.atlassian.net/browse/PURCHASE-1118


hokusai/development.yml
```diff
@@ -2,13 +2,38 @@
 version: '2'
 services:
   exchange:
-    command: bundle exec rails server -b 0.0.0.0
+    command: sh -c 'bundle exec rails db:setup && bundle exec puma -C config/puma.config'
```
izakp: Prefer the "exec" form of the `command` / `CMD` in the Dockerfile with arguments as elements in an array, so `["bundle", "exec" ... "config/puma.config"]` and removing the call to `sh`

See https://stackoverflow.com/questions/42805750/dockerfile-cmd-shell-versus-exec-form

config/puma.config
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env puma
+
+directory "/app"
+environment ENV.fetch("RAILS_ENV") { "development" }
+daemonize false
+pidfile  "/shared/pids/puma.pid"
+state_path "/shared/pids/puma.state"
+threads ENV.fetch("PUMA_THREAD_MIN") { 0 }.to_i, ENV.fetch("PUMA_THREAD_MAX") { 16 }.to_i
+bind ENV.fetch("PUMA_BIND") { "tcp://0.0.0.0:#{ENV.fetch('PORT') { 3000 }}" }
```
izakp: Would use either `PUMA_BIND` or `PORT` here to avoid confusion

kierangillen: Could you elaborate on this? Not sure exactly what you mean.

hokusai/development.yml
kierangillen: Cool, updated this.

config/puma.config
izakp: Actually I think this is okay... it respects `PUMA_BIND` if set but then falls back to `PORT` defaulting to `3000`.  Let's just be consistent with how either `PUMA_BIND` or `PORT` is set in environments

hokusai/production.yml
```diff
@@ -27,6 +26,8 @@ spec:
           env:
             - name: PORT
```
izakp: @kierangillen relating to my other comment - so here we're setting both `PORT` and `PUMA_BIND` which essentially do the same thing.  So let's only set `PORT`?

PR #429, opened 2019-07-08
Gem updates, mainly active-admin and stripe-mock

ashkan18:
# Why? Why?!?!?
As part of SCA and switching to Stripe's `payment_intent` I wanted to update `stripe_ruby_mock` but ended up doing a `bundle update` which showed we are behind on few libs, the bigger changes were:
- `activeadmin (1.3.1)` -> `2.1.0`: this meant we need to remove `.coffee` file and change to `js` since active admin has moved away from ☕️ script
- `paper_trail (10.0.1)`  -> `10.3.0`: this required updating all `has_papertrail` methods to wrap `class` in `versions: {...}`.
- `graphql (1.8.11)` -> `1.9.7`, this introduced better error messages for invalid params passed to query/mutation. Updated our tests to match new changes.


sweir27:
Woooooo!

adventure_party

sweir27:
Reran the workflow, will see if it passes.

ashkan18:
looks like its a rubocop failure where rubocop has removed `--rails` option https://github.com/rubocop-hq/rubocop/issues/5976 going to fix it.

ashkan18:
had to revert `puma` update because of https://github.com/puma/puma/issues/1828

ashkan18:
@sweir27 back to 🍏 so those ice creams can be happy again

PR #430, opened 2019-07-08
[PURCHASE-1180] Extend approval expiration to 3 days

ashkan18:
# Change
We need to extend approval expiration to 72 hours.
fixes https://artsyproduct.atlassian.net/browse/PURCHASE-1180

# Does this reflect existing pending orders?
No, once deployed to prod, it will impact *only* orders submitted after the deployment, for those order, approval expiration will be 3 days.

# Should this be an env variable?
I thought about it, but I think this is somehow significant enough change that might still be better to go  through PR and code review. For example extending this beyond 7 days will have charge expiration issue. 

# Related PRs
Pulse and Force have been already updated to show expiration based on `stateExpiresAt` and not hardcode in:
https://github.com/artsy/reaction/pull/2571
https://github.com/artsy/pulse/pull/372

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -57,7 +57,7 @@ class Order < ApplicationRecord
 
   STATE_EXPIRATIONS = {
     'pending' => 2.days,
-    'submitted' => 2.days,
+    'submitted' => 3.days,
     'approved' => 7.days
   }.freeze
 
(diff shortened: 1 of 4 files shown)


ashkan18:
Dont merge yet -> https://artsy.slack.com/archives/C9YNS4X32/p1562771682442200

sweir27:
Merging since we got the go-ahead in slack!

PR #432, opened 2019-07-12
Update offer expiration to match submitted order expiration

ashkan18:
# Problem
When we tested #430 on staging, looks like buy now orders get proper expiration but for MO we still get 2 days.

# Cause
We had a separate const for offer expiration, this probably originally was designed this way so we can decouple offer and order expirations since offers are less timely. For example, its ok to set expiration on an offer to be 8 days but we can't do that on order since stripe hold will expire.

# Solution
Update offer expiration to 3 days.


sweir27:
merge on green

PR #436, opened 2019-07-17
Make GBP a Supported Currency for Orders (GALL-1966)

iskounen:
This PR adds GBP to the list of supported currencies in Exchange. The list of supported currencies is used to validate the currency of a new order, and without GBP, collectors will not be able to click on the "Buy now" or "Make offer" buttons on Artsy.net to begin the order or offer flow.

#### Problem

unable to create order

#### Notes

- In order for artworks enrolled in NSO to be listed in GBP, partners need to have access to the `CMS UK Commerce` lab feature, so this change does not pose a risk to existing NSO artworks in production.
- This PR adds "happy path" integration tests for GBP artworks to ensure that orders can be created from GBP artworks and retain the same currency through the order and offer flows. These tests do not, however, use UK addresses for the seller and buyer since more work will be needed to [correctly calculate tax](https://artsyproduct.atlassian.net/browse/GALL-1979) and shipping for UK addresses.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -2,7 +2,7 @@ class Order < ApplicationRecord
   include OrderHelper
   has_paper_trail versions: { class_name: 'PaperTrail::OrderVersion' }
 
-  SUPPORTED_CURRENCIES = %w[USD].freeze
+  SUPPORTED_CURRENCIES = %w[USD GBP].freeze
 
   DEFAULT_EXPIRATION_REMINDER = 5.hours
 
--- a/spec/integration/gbp/buy_now_happy_path_spec.rb
+++ b/spec/integration/gbp/buy_now_happy_path_spec.rb
@@ -1,0 +1,141 @@
+require 'rails_helper'
+require 'support/gravity_helper'
+require 'support/taxjar_helper'
+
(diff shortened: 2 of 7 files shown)


mdole:
@iskounen this looks reasonable to me! Is the screenshot meant to show the current state of affairs? If so, would you want to update with an "After" screenshot?

iskounen:
@mdole yes, this is the current state of affairs on staging. I'm guessing that this change is sufficient to successfully create an order in Exchange and will address the problem. However, I haven't tested that locally because of the effort required to set up Exchange and Force to run together locally.

mdole:
ah I see I see. it doesn't seem like this would be likely to break something, but I'm also a little wary of pushing code that we haven't seen in action up to staging - @ashkan18 what're your thoughts? Do you have recs for testing this type of change based on your previous experience?

iskounen:
I've started to write integration tests for the UK NSO happy path. I decided to use UK addresses in addition to setting the artwork price to GBP and started noticing issues related to tax and shipping calculations. I can revert this to use US addresses with GBP prices and update the integration tests to include UK addresses once we've figured out how to determine whether taxes should be calculated or not for UK partners.

PR #437, opened 2019-07-17
[PURCHASE-1315] Cancel PaymentIntent on order expiration

lilyfromseattle:
I added `cancel_payment_intent` to `PaymentService`.

I updated `OrderFollowUpJob` so that `cancel_payment_intent` is called for expired orders if the order state is `submitted` or `pending`. I added tests for this.

For some reason the test `PaymentIntent`s that I'm creating have a status of `succeeded` even though I am setting `capture_method` to manual. They _should_ have a status of `requires_capture`. For this reason the actual cancelling of the payment intents isn't working (payment intents with a `succeeded` status can't be cancelled).

### DIFF
--- a/app/jobs/order_follow_up_job.rb
+++ b/app/jobs/order_follow_up_job.rb
@@ -7,10 +7,8 @@ class OrderFollowUpJob < ApplicationJob
 
     case order.state
     when Order::PENDING
-      PaymentService.cancel_payment_intent(order.transactions.last.external_id)
       OrderService.abandon!(order)
     when Order::SUBMITTED
-      PaymentService.cancel_payment_intent(order.transactions.last.external_id)
       cancel_submitted_order(order)
     when Order::APPROVED
       # Order was approved but has not yet fulfilled,
--- a/app/jobs/order_follow_up_job.rb
+++ b/app/jobs/order_follow_up_job.rb
@@ -7,8 +7,10 @@ class OrderFollowUpJob < ApplicationJob
 
     case order.state
     when Order::PENDING
+      PaymentService.cancel_payment_intent(order.transactions.last.external_id)
       OrderService.abandon!(order)
     when Order::SUBMITTED
+      PaymentService.cancel_payment_intent(order.transactions.last.external_id)
       cancel_submitted_order(order)
     when Order::APPROVED
       # Order was approved but has not yet fulfilled,
--- a/lib/payment_service.rb
+++ b/lib/payment_service.rb
@@ -97,9 +97,9 @@ module PaymentService
     end
   end
 
-  def self.cancel_payment_intent(external_id, cancelation_reason)
+  def self.cancel_payment_intent(external_id, cancellation_reason)
     payment_intent = Stripe::PaymentIntent.retrieve(external_id)
-    payment_intent.cancel(cancelation_reason: cancelation_reason)
+    payment_intent.cancel(cancellation_reason: cancellation_reason)
     Transaction.new(
       external_id: payment_intent.id,
       source_id: payment_intent.payment_method,
--- a/lib/payment_service.rb
+++ b/lib/payment_service.rb
@@ -99,7 +99,7 @@ module PaymentService
 
   def self.cancel_payment_intent(external_id, cancelation_reason)
     payment_intent = Stripe::PaymentIntent.retrieve(external_id)
-    payment_intent.cancel
+    payment_intent.cancel(cancelation_reason: cancelation_reason)
     Transaction.new(
       external_id: payment_intent.id,
       source_id: payment_intent.payment_method,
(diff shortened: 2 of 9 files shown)


spec/jobs/order_follow_up_job_spec.rb
```diff
@@ -34,7 +34,7 @@
         let(:order) { Fabricate(:order, mode: mode, state: state, buyer_id: buyer_id, seller_id: seller_id, seller_type: seller_type, buyer_type: buyer_type, external_charge_id: payment_intent.id) }
         let(:credit_card) { { external_id: stripe_customer.default_source, customer_account: { external_id: stripe_customer.id } } }
         let(:charge) { Stripe::Charge.create(amount: 20_00, currency: 'usd', source: credit_card) }
-        let(:payment_intent) { Stripe::PaymentIntent.create(amount: 20_00, currency: 'usd', charges:[charge], payment_method: stripe_customer.default_source, capture_method: "manual", confirmation_method: "manual", confirm: false) }
+        let(:payment_intent) { Stripe::PaymentIntent.create(amount: 20_00, currency: 'usd', charges:[charge], payment_method: stripe_customer.default_source, capture_method: "manual", confirmation_method: "manual") }
```
ashkan18: this looks like the same `payment_intent` above, right? we can probably just use the one defined in higher level scope.

spec/jobs/order_follow_up_job_spec.rb
```diff
@@ -54,7 +54,7 @@
         context 'Offer order' do
           let(:mode) { Order::OFFER }
           let(:offer) { Fabricate(:offer, from_id: seller_id, order: order, from_type: seller_type, submitted_at: Time.now.utc) }
-          let(:payment_intent) { Stripe::PaymentIntent.create(amount: 20_00, currency: 'usd', payment_method: stripe_customer.default_source, capture_method: "manual", confirmation_method: "manual", confirm: false) }
+          let(:payment_intent) { Stripe::PaymentIntent.create(amount: 20_00, currency: 'usd', payment_method: stripe_customer.default_source, capture_method: "manual", confirmation_method: "manual") }
```
ashkan18: same thing with this, i know it was not added in your PR but may make sense to consolidate these.

app/jobs/order_follow_up_job.rb
```diff
@@ -7,8 +7,10 @@ def perform(order_id, state)
 
     case order.state
     when Order::PENDING
+      PaymentService.cancel_payment_intent(order.transactions.last.external_id)
```
ashkan18: this call belongs to those services, so i would move this to `OrderService.abandon!`

we need to make sure the transaction is also a `payment_intent` transaction to be safe, so something more like `order.transactions.payment_intent.last.external_id` (assumed we will add a scope for `payment_intent` transactions)

app/jobs/order_follow_up_job.rb
```diff
@@ -7,8 +7,10 @@ def perform(order_id, state)
...
       OrderService.abandon!(order)
     when Order::SUBMITTED
+      PaymentService.cancel_payment_intent(order.transactions.last.external_id)
```
ashkan18: same comment as above, this call should go to `OrderCancelationService`

lib/payment_service.rb
```diff
@@ -31,7 +31,7 @@ def self.refund_charge(external_id)
     else
       # we need to refund using payment_intent
       payment_intent = Stripe::PaymentIntent.retrieve(external_id)
-      refund = Stripe::Refund.create(charge: payment_intent.charge.id, reverse_transfer: true)
+      refund = Stripe::Refund.create(charge: payment_intent.charges.last.id, reverse_transfer: true)
```
ashkan18: we need to document (comment) the fact that we made assumption that we will only have one charge. alternative is to iterate through charges and be more future-proof, but that seems more confusing.

lib/payment_service.rb
```diff
@@ -74,6 +74,10 @@ def self.create_payment_intent(credit_card:, buyer_amount:, seller_amount:, merc
     end
   end
 
+  def self.cancel_payment_intent(external_id)
+    payment_intent = Stripe::PaymentIntent.retrieve(external_id)
+    payment_intent.cancel
```
ashkan18: looks like we can also pass `cancelation_reason` https://stripe.com/docs/api/payment_intents/cancel#cancel_payment_intent-cancellation_reason would be useful to set them for abandoned and by customer cases.

also it seems it can potentially fail, we should probably check for that

ashkan18: this should also create a `transaction` for canceling the payment. All of our external calls to Stripe should lead to a new row in `transaction` table for that order.

spec/jobs/order_follow_up_job_spec.rb
lilyfromseattle: Actually the one above doesn't have a charge but the one below can use this

lib/payment_service.rb
lilyfromseattle: In what case would we be refunding a charge that isn't the most recent?

lib/payment_service.rb
lilyfromseattle: We want to make a cancel payment transaction even in the case where we are already creating a refund transaction?

lilyfromseattle: So two transactions?

lilyfromseattle: And in the case that canceling the payment fails do we still create a transaction? Like use `generate_transaction_from_exception`?

ashkan18:
@lilyfromseattle going to close this for now and we can come back to it once done with make offer SCA

PR #438, opened 2019-07-18
Transaction external type

ashkan18:
# Change
This adds `external_type` to `transactions` table, this way we can distinguish between `charge`, `refund` and newly added `payment_intent` transactions.
This way when trying to `capture` or `refund` we first find the `transaction` for this `external_id` and based on it's type decide if we need to do a `charge` or `payment_intent` refund/capture.

### DIFF
--- a/app/graphql/mutations/order_or_failure_union_type.rb
+++ b/app/graphql/mutations/order_or_failure_union_type.rb
@@ -8,7 +8,7 @@ class Mutations::OrderWithMutationFailure < Types::BaseObject
   field :error, Types::ApplicationErrorType, null: false
 end
 
-class Mutations::OrderNeedsAction < Types::BaseObject
+class Mutations::OrderRequiresAction < Types::BaseObject
   description 'A response indicating order needs action to be able to be submitted'
   field :action_data, JSON, null: false, description: 'Data related to action needed'
 end
--- a/app/graphql/mutations/order_or_failure_union_type.rb
+++ b/app/graphql/mutations/order_or_failure_union_type.rb
@@ -10,16 +10,16 @@ end
 
 class Mutations::OrderRequiresAction < Types::BaseObject
   description 'A response indicating order needs action to be able to be submitted'
-  field :action_data, JSON, null: false, description: 'Data related to action needed'
+  field :action_data, GraphQL::Types::JSON, null: false, description: 'Data related to action needed'
 end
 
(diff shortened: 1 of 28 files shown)


app/graphql/mutations/order_or_failure_union_type.rb
```diff
@@ -8,7 +8,7 @@ class Mutations::OrderWithMutationFailure < Types::BaseObject
   field :error, Types::ApplicationErrorType, null: false
 end
 
-class Mutations::OrderNeedsAction < Types::BaseObject
+class Mutations::OrderRequiresAction < Types::BaseObject
```
sweir27: Why this change?

sweir27: Also does the `OrderOrFailureUnionType` need to have this as part of the `possible_types`?

ashkan18: this new name matches the actual `transaction_status`. Yes, we added it to union type.

db/migrate/20190717171856_populate_transaction_external_type.rb
```diff
@@ -0,0 +1,10 @@
+class PopulateTransactionExternalType < ActiveRecord::Migration[5.2]
+  def up
+    Transaction.where('external_id LIKE :prefix', prefix: 'ch_').update_all(external_type: Transaction::CHARGE)
+    Transaction.where('external_id LIKE :prefix', prefix: 're_').update_all(external_type: Transaction::REFUND)
```
ashkan18: we also have `re_` transactions which are for `reufnds` so we need to update them, the tricky thing about refunds is their `transaction_type` and `external_type` is both `Transaction::REFUNDS`

lib/payment_service.rb
```diff
@@ -1,44 +1,59 @@
...
+  def self.hold_charge(charge_params)
     create_payment_intent(charge_params.merge(capture: false))
   end
 
-  def self.create_and_capture_charge(charge_params)
+  def self.immediate_capture(charge_params)
     create_payment_intent(charge_params.merge(capture: true))
   end
 
   def self.capture_authorized_charge(external_id)
```
ashkan18: im not in ❤️ with this method getting `external_id` and not our internal `transaction` `id`. but this may be a bigger change.

PR #442, opened 2019-07-23
Refactor order_processor to not raise error

ashkan18:
# Problem
When we started updating our logic around adding support for `PaymentIntents` which would add more conditions to a _failed payment_. Things got pretty complicated between `OrderProcessor` and how it's being used by `OrderService` and `OfferService`.

One thing that complicated things was who and where we raise exceptions and how and where we rescue from them. We were relying on exception raised from `OrderProcessor` to _rollback_ the status changes during order/offer `submit`.

# Change
instead of raising exceptions from `OrderProcessor`, have it set the status internally so the callers can decide what to do with the failed reason. 

This will help us later when we try to handle `requires_action` case of `PaymentIntent` but we should look at this PR regardless of the coming changes.

### DIFF
--- a/app/services/offer_service.rb
+++ b/app/services/offer_service.rb
@@ -27,7 +27,7 @@ module OfferService
   def self.submit_pending_offer(offer)
     order = offer.order
     raise Errors::ValidationError, :invalid_offer if offer.submitted?
-    raise Errors::ProcessingError, :insufficient_inventory unless order.inventory?
+    raise Errors::InsufficientInventoryError unless order.inventory?
 
     order = offer.order
     offer_order_totals = OfferOrderTotals.new(offer)
@@ -67,7 +67,7 @@ module OfferService
 
     order = offer.order
     order_processor = OrderProcessor.new(order, user_id)
-    raise Errors::ValidationError, order_processor.error unless order_processor.valid?
+    raise Errors::ValidationError, order_processor.validation_error unless order_processor.valid?
 
     order.approve! do
       totals = OfferOrderTotals.new(offer)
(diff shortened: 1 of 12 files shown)


app/services/order_service.rb
```diff
@@ -57,17 +57,24 @@ def self.submit!(order, user_id)
...
+      # in case of failed transaction, we need to rollback this block,
+      # but still need to add transaction, so we raise an ActiveRecord::Rollback
+      raise ActiveRecord::Rollback if order_processor.failed_payment?
+
+      order.update!(external_charge_id: order_processor.transaction.external_id)
     end
 
+    order.transactions << order_processor.transaction
+    if order_processor.transaction.failed?
+      PostTransactionNotificationJob.perform_later(order_processor.transaction.id, TransactionEvent::CREATED, user_id)
```
ashkan18: maybe we should send this all the time.

app/services/offer_service.rb
```diff
@@ -78,19 +78,18 @@ def self.accept_offer(offer, user_id)
         seller_total_cents: totals.seller_total_cents
       )
       order_processor.charge!
-      order.transactions << order_processor.transaction
+      raise ActiveRecord::Rollback if order_processor.failed_payment?
+      raise Errors::InsufficientInventoryError if order_processor.failed_inventory?
```
joeyAghion: Potentially confusing for future developers how this raises exceptions up to callers while the line above just jumps out of the current block. Might be worth an inline comment (😲) if intentional.

app/services/offer_service.rb
```diff
@@ -78,19 +78,18 @@ def self.accept_offer(offer, user_id)
...
+    end
+    order.transactions << order_processor.transaction
```
joeyAghion: 👍 about consolidating this into a single location

app/services/offer_service.rb
```diff
@@ -78,19 +78,18 @@ def self.accept_offer(offer, user_id)
...
+    if order_processor.transaction.failed?
+      PostTransactionNotificationJob.perform_later(order_processor.transaction.id, TransactionEvent::CREATED, user_id)
```
joeyAghion: If we could _always_ queue this job upon creating a transaction (as the name implies), the other failure handling could be 1-lined and might be easier to follow.

app/services/order_service.rb
```diff
@@ -57,17 +57,24 @@ def self.submit!(order, user_id)
...
+      raise Errors::FailedTransactionError.new(:charge_authorization_failed, order_processor.transaction)
+    end
```
joeyAghion: Similar thoughts here.

ashkan18:
@joeyAghion addressed our concerns

app/services/order_service.rb
```diff
@@ -57,17 +57,23 @@ def self.submit!(order, user_id)
...
+      # in case of failed transaction, we need to rollback this block,
+      # but still need to add transaction, so we raise an ActiveRecord::Rollback
+      raise ActiveRecord::Rollback if order_processor.failed_payment?
+
+      order.update!(external_charge_id: order_processor.transaction.external_id)
     end
 
+    order.transactions << order_processor.transaction
+    PostTransactionNotificationJob.perform_later(order_processor.transaction.id, user_id)
+    raise Errors::FailedTransactionError.new(:charge_authorization_failed, order_processor.transaction) if order_processor.transaction.failed?
```
joeyAghion: The `raise ActiveRecord::Rollback...` line above checks `order_processor.failed_payment?`, but this checks `order_processor.transaction.failed?`. Is that difference intentional?

ashkan18: current logic works too since `transaction.failed?` is similar to `failed_payment?` but makes sense to use the same thing here, going to update.

PR #440, opened 2019-07-18
[WIP] For SCA work

ashkan18:
WIP

PR #443, opened 2019-07-24
Switch to payment intent stuff

ashkan18:
# Change
https://artsyproduct.atlassian.net/browse/PURCHASE-1331
As part of our path to support for SCA, we need to switch to Stripe's [`PaymentIntent` API](https://stripe.com/docs/payments/payment-intents). 
This means we need to move way from directly using `Charge` API. We need to create a `PaymentIntent` at the time we are trying hold or capture a charge and confirm it. 

* A successfully confirmed `PaymentIntent` will lead to a charge.
* PaymentIntents can fail for different reason when we confirm:
  - They can fail confirmation: Payment intent goes to `require_action` state (not supported in this PR) meaning we need to ask buyer to go through SCA.
  - They can fail processing charge: (think of insufficient funds and etc.) Payment intent goes to `require_payment_method` state (maps to our current charge failures cases)
 
# Stripe API version update?
We are currently on a relatively old version `2017-08-15` which is also our default version on Stripe dashboard. We need to update to latest version `2019-05-16` so we can use these statuses mentioned ☝️. In the older version they are set to different statuses.
We don't want to update our default version in Stripe dashboard since that impacts all places we use Stripe in our services. but we can manually set the version for Exchange by configuring `Stripe` in `stripe.rb`.

# Migration?
We now store `PaymentIntent` in our `Transaction` table, this means now transaction table can have different external types, for that we added `external_type` to `Transaction` and we have a data migration to set it for existing records. 
During capture already held charges and also for refunds, we check to see whats the type of the existing transaction and based on that we decide if we need to use old Charge API or fancy shiny new Payment Intent.
We now also store `payload` of the response coming back from Stripe in `transaction` as `jsonb` for possible later use.

# What this PR is NOT
We don't yet support payment intent in `requires_action` state. This is the state needed for SCA support and will be supported in follow up PRs, this PR is purely switching to PaymentIntent for existing functionality.

# stripe-ruby-mock?
On our first attempt #440 for SCA, we ran into lot of issues with `stripe-ruby-mock` which didn't support `PaymentIntent`s yet. We tried to add support to it by submitting 2 PRs but it was slowing us down, so in this PR we tried to move away from it and use plain Ruby mocks to mock Stripe's calls during tests.
This normally would be worrisome since it feels less realistic when we mock our calls, but now after 2 PRs knowing how `stripe-ruby-mock` works, i don't think our tests are much less realistic than using `stripe-ruby-mock`. This way of testing is def lot easier and less magical.

### DIFF
--- a/app/controllers/errors/error_types.rb
+++ b/app/controllers/errors/error_types.rb
@@ -54,6 +54,7 @@ module Errors
     ],
     processing: %i[
       artwork_version_mismatch
+      cannot_capture
       capture_failed
       charge_authorization_failed
       insufficient_inventory
@@ -62,8 +63,8 @@ module Errors
       tax_calculator_failure
       tax_recording_failure
       tax_refund_failure
-      unknown_event_charge
       undeduct_inventory_failure
+      unknown_event_charge
     ],
     internal: %i[
       generic
--- a/app/services/order_cancellation_service.rb
+++ b/app/services/order_cancellation_service.rb
@@ -51,7 +51,7 @@ class OrderCancellationService
   def process_stripe_refund
     raise Errors::ValidationError.new(:unsupported_payment_method, @order.payment_method) unless @order.payment_method == Order::CREDIT_CARD
 
-    @transaction = PaymentService.refund_charge(@order.external_charge_id)
+    @transaction = PaymentService.refund_payment(@order.external_charge_id)
     raise Errors::ProcessingError.new(:refund_failed, @transaction.failure_data) if @transaction.failed?
   end
 
--- a/app/services/order_cancellation_service.rb
+++ b/app/services/order_cancellation_service.rb
@@ -51,7 +51,8 @@ class OrderCancellationService
   def process_stripe_refund
     raise Errors::ValidationError.new(:unsupported_payment_method, @order.payment_method) unless @order.payment_method == Order::CREDIT_CARD
 
-    @transaction = PaymentService.refund_payment(@order.external_charge_id)
+    payment_transaction = @order.transactions.where(external_id: @order.external_charge_id).first
+    @transaction = PaymentService.refund(@order.external_charge_id, payment_transaction.external_type)
     raise Errors::ProcessingError.new(:refund_failed, @transaction.failure_data) if @transaction.failed?
   end
 
--- a/db/migrate/20190723204947_populate_transaction_external_type.rb
+++ b/db/migrate/20190723204947_populate_transaction_external_type.rb
@@ -1,0 +1,10 @@
+class PopulateTransactionExternalType < ActiveRecord::Migration[5.2]
+  def up
+    Transaction.where('external_id LIKE :prefix', prefix: 'ch_').update_all(external_type: Transaction::CHARGE)
+    Transaction.where('external_id LIKE :prefix', prefix: 're_').update_all(external_type: Transaction::REFUND)
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration, "This was a data migration, can't be easily rollback"
+  end
+end
--- a/db/migrate/20190723204947_populate_transaction_external_type.rb
+++ b/db/migrate/20190723204947_populate_transaction_external_type.rb
@@ -5,6 +5,6 @@ class PopulateTransactionExternalType < ActiveRecord::Migration[5.2]
   end
 
   def down
-    raise ActiveRecord::IrreversibleMigration, "This was a data migration, can't be easily rollback"
+    raise ActiveRecord::IrreversibleMigration, "This was a data migration, can't be easily rolled back"
   end
 end
--- a/lib/order_processor.rb
+++ b/lib/order_processor.rb
@@ -35,7 +35,7 @@ class OrderProcessor
     raise Errors::ValidationError, @validation_error unless valid?
 
     deduct_inventory
-    @transaction = PaymentService.immediate_capture_payment(construct_charge_params)
+    @transaction = PaymentService.capture_without_hold(construct_charge_params)
     undeduct_inventory if @transaction.failed?
   rescue Errors::InsufficientInventoryError
     undeduct_inventory
--- a/lib/order_processor.rb
+++ b/lib/order_processor.rb
@@ -24,7 +24,7 @@ class OrderProcessor
     raise Errors::ValidationError, @validation_error unless valid?
 
     deduct_inventory
-    @transaction = PaymentService.create_and_authorize_charge(construct_charge_params)
+    @transaction = PaymentService.hold_payment(construct_charge_params)
     undeduct_inventory if @transaction.failed?
   rescue Errors::InsufficientInventoryError
     undeduct_inventory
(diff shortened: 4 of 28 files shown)


spec/services/payment_service_spec.rb
```diff
@@ -22,66 +21,89 @@
...
+        failure_message: 'The card was declined',
+        decline_code: 'do_not_honor',
+        transaction_type: Transaction::HOLD,
+        status: Transaction::FAILURE,
+        payload: { 'id' => 'pi_1' }
+      )
     end
   end
 
-  describe '#capture_authorized_charge' do
```
lilyfromseattle: Should we wait till we remove `capture_authorized_charge` to remove this test?

db/migrate/20190723204947_populate_transaction_external_type.rb
```diff
@@ -0,0 +1,10 @@
+class PopulateTransactionExternalType < ActiveRecord::Migration[5.2]
+  def up
+    Transaction.where('external_id LIKE :prefix', prefix: 'ch_').update_all(external_type: Transaction::CHARGE)
+    Transaction.where('external_id LIKE :prefix', prefix: 're_').update_all(external_type: Transaction::REFUND)
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration, "This was a data migration, can't be easily rollback"
```
lilyfromseattle: Small typo - rollback should be rolled back

lib/payment_service.rb
```diff
@@ -1,59 +1,115 @@
...
+      amount_cents: payment_intent.amount,
+      transaction_type: Transaction::CAPTURE,
+      payload: payment_intent.to_h
+    )
+    update_transaction_with_payment_intent(new_transaction, payment_intent)
+    new_transaction
   end
 
-  def self.refund_charge(charge_id)
+  def refund_charge(charge_id)
```
lilyfromseattle: Why are we removing self here only for this method?

spec/support/stripe_helper.rb
```diff
@@ -0,0 +1,74 @@
+RSpec.shared_context 'include stripe helper' do
```
lilyfromseattle: Should we remove `stripe-ruby-mock` from the Gemfile now that we're no longer using it?

lib/payment_service.rb
```diff
@@ -1,59 +1,115 @@
...
     refund = Stripe::Refund.create(charge: charge_id, reverse_transfer: true)
-    Transaction.new(external_id: refund.id, transaction_type: Transaction::REFUND, status: Transaction::SUCCESS)
+    Transaction.new(external_id: refund.id, external_type: Transaction::REFUND, transaction_type: Transaction::REFUND, status: Transaction::SUCCESS, payload: refund.charge)
```
lilyfromseattle: Shouldn't `external_type`  be `charge` here?

app/services/order_cancellation_service.rb
```diff
@@ -51,7 +51,8 @@ def process_inventory_undeduction
   def process_stripe_refund
     raise Errors::ValidationError.new(:unsupported_payment_method, @order.payment_method) unless @order.payment_method == Order::CREDIT_CARD
 
-    @transaction = PaymentService.refund_charge(@order.external_charge_id)
+    payment_transaction = @order.transactions.where(external_id: @order.external_charge_id).first
```
ashkan18: im not in ❤️ with how we store the external id on our `order` as `external_charge_id`, i think we should store our internal `transaction` of that charge on the order instead. but this would be a bigger change for this, maybe we can do this in separate PR.

lib/payment_service.rb
ashkan18: what we get back from Stripe is a `Refund` object and historically we've been storing `refund.id` so i think it should be `REFUND` although i agree it's weird to have `external_type` be `refund` and `transaction_type` also be `refund` but it seems more realistic to me.

spec/support/stripe_helper.rb
ashkan18: we are still using it for testing `webhooks`, once thats removed we can ✂️ it.

spec/services/order_approve_service_spec.rb
```diff
@@ -1,76 +1,89 @@
...
+      context 'with an order that was paid for by wire transfer' do
+        it 'raises an error' do
+          order.update!(payment_method: Order::WIRE_TRANSFER)
+          expect { service.process! }.to raise_error do |e|
+            expect(e.code).to eq :unsupported_payment_method
+          end
         end
       end
     end
+    context 'capture a payment_intent' do
```
ashkan18: working on adding step here.

lib/payment_service.rb
lilyfromseattle: Ok should we add `REFUND` to `EXTERNAL_TYPES` in the model?

app/services/order_cancellation_service.rb
sweir27: Oh interesting. Yeah this does seem weird, but makes sense it would be a bigger change.

lib/order_processor.rb
```diff
@@ -35,7 +35,7 @@ def charge!
     raise Errors::ValidationError, @validation_error unless valid?
 
     deduct_inventory
-    @transaction = PaymentService.create_and_capture_charge(construct_charge_params)
+    @transaction = PaymentService.immediate_capture_payment(construct_charge_params)
```
sweir27: This method name is a little odd-sounding. Maybe `immediately_capture_payment` or `capture_payment_immediately`?

sweir27: Actually, why do we need the `immediate` at all?

ashkan18: yeah, do you think original name was better? the thing is with `payment_intent` now `create` is actually different than `charge`. Charge creation was actually process payment, where creation here may not, and the one that actually processes payment is _confirm_ so tried to take `create` out of it. basically want to reflect that this method doesn't do a hold

ashkan18:
@joeyAghion going to merge this so we can start testing on staging sooner, but please review and i'll address comments on follow up PRs

app/controllers/errors/error_types.rb
```diff
@@ -54,6 +54,7 @@ module Errors
     ],
     processing: %i[
       artwork_version_mismatch
+      cannot_capture
```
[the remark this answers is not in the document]
ashkan18: this would only happen if we try to capture an already captured payment intent. going to make it more obvious.

ashkan18:
thanks @joeyAghion for comments, I agree with all those steps, i think from our perspective in order to unblock ourselves for SCA work we just went with different order:
- **Refactor order processor so it doesn't raise exception**: this made it easier for us to switch to payment intents
- **Update API version and switch to payment Intent**(this PR):  we did this in one shot because it made update API version almost a one line change. The API changes were mostly impacting creating charges and it would have changed our we find source id's from a charge. Also in old version the status's of a `PaymentIntent` were also named differently. Doing it as part of this PR made API version upgrade lot easier.
- **Add support for PaymentIntent's require_action case **: #455 this exposes the requires action case to client's by adding new graphql type in the current mutation responses
- **Reuse existing payment intent on resubmit**
once all of those are supported and our front end work is blocked, we were thinking of doing some refactoring we talked about which technically would have no impact to clients:
- Remove `external_charge_id` and support list of associated payments to an order
- Move more logic from `OrderService` to `OrderProcessor` and try to avoid complications caused by using db transactions.

PR #445, opened 2019-07-25
Use nginx sidecar container

eessex:
- Adds nginx container to staging and production
- Adds missing `affinity` blob to staging


izakp:
LGTM!

PR #449, opened 2019-07-29
Fix data migration to actually use prefix

ashkan18:
Update data migration that was missing `%` when doing a prefix matching for `external_id`.


sweir27:
merge on green

joeyAghion:
Will this have any effect if the migration has already been applied?

ashkan18:
we manually ran that data migration on staging, it hasn't been run on prod yet.

PR #451, opened 2019-07-29
Fix missing setting external_charge_id on approved offers

ashkan18:
# Problem
When we tried to refund our test offer orders on production we noticed the refund fails.

# Cause
During [our refactor](https://github.com/artsy/exchange/pull/442) we moved the responsibility of setting `external_charge_id` on an `Order` from `order_processor` to the caller. We fixed `order_service` to set it in a successful transaction but we missed it on the `offer_service`.

# Impact
Any offers approved after deploy https://github.com/artsy/exchange/pull/444 won't have `external_charge_id` so can't be refunded. 

NOTE BN orders are fine and also partner's don't have the option for refund, this all happens by our admins from exchange admin.

# Solution
Update `offer_service` to set `external_charge_id` in case of success and add tests for it.  

## Migration
for offers approved in last 4 days without `external_charge_id` we have to set them to their successful transaction external id.

### DIFF
--- a/app/services/offer_service.rb
+++ b/app/services/offer_service.rb
@@ -82,6 +82,8 @@ module OfferService
       # in case of failed transaction, we need to rollback this block,
       # but still need to add transaction, so we raise an ActiveRecord::Rollback
       raise ActiveRecord::Rollback if order_processor.failed_payment?
+
+      order.update!(external_charge_id: order_processor.transaction.external_id)
     end
     order.transactions << order_processor.transaction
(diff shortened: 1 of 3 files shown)


ashkan18:
Migration we end up running
```
Order.where(state: 'approved', mode: 'offer', external_charge_id: nil).map { |o| o.update!(external_charge_id: o.transactions.first.external_id) }
```
it changed 5 orders, 2 of them being our own tests.

PR #453, opened 2019-07-29
Add internalID to all our types

ashkan18:
Follows https://artsyproduct.atlassian.net/browse/PLATFORM-1448 to add `internalID` to all our types pointing to ID.

# Possible follow ups
have our mutations/queries accept `internalID`


alloy:
Looks great, thanks! 🙌

There shouldn’t be a need to rename the arg, we’re not doing that in MP either.

alloy:
#MergeOnGreen

PR #455, opened 2019-07-30
Add support for `requires_action` payment_intent state

ashkan18:
# Change
during #443 we switched over to `PaymentIntent` instead of charge. But that change yet doesn't support new status we could get back from that API where PaymentIntent was not `succeeded` and needed `requires_action`.

## What does it mean transaction-wise?
In `requires_action` case, we store a transaction in `Transaction` with status being newly added `Transaction::REQUIRES_ACTION`.
We also update the order's `external_charge_id` to be this payment intent's id so we can reuse it later when user was done with 3d auth.

## Exposing this state via GraphQL
When a `PaymentIntent` goes to `requires_action` state, we need to return `client_secret` of the `payementIntent` back to the client, so client can use it to call Stripe.js to open popup and go through 3d auth process. The way we do this is by updating our existing `OrderOrFailureUnion` type to also now possibly return `OrderPaymentRequiresAction` type which would have a  `actionData` with new `orderActionData` type that exposes `client_secret`.

# What this PR is NOT
this PR doesn't support re-submitting an order after user been through 3d auth. Meaning if this goes live, we can get data about this state so client can show the pop up but once they passed that and re-submit we'd create a new payment intent which would go through the same state. This will be covered in a follow up PR.

### DIFF
--- a/lib/order_processor.rb
+++ b/lib/order_processor.rb
@@ -25,7 +25,7 @@ class OrderProcessor
 
     deduct_inventory
     @transaction = PaymentService.hold_payment(construct_charge_params)
-    undeduct_inventory if @transaction.failed?
+    undeduct_inventory if @transaction.failed? || @transaction.requires_action?
   rescue Errors::InsufficientInventoryError
     undeduct_inventory
     @insufficient_inventory = true
@@ -36,7 +36,7 @@ class OrderProcessor
 
     deduct_inventory
     @transaction = PaymentService.capture_without_hold(construct_charge_params)
-    undeduct_inventory if @transaction.failed?
+    undeduct_inventory if @transaction.failed? || @transaction.requires_action?
   rescue Errors::InsufficientInventoryError
     undeduct_inventory
     @insufficient_inventory = true
@@ -44,6 +44,14 @@ class OrderProcessor
 
   def failed_payment?
     @transaction.present? && @transaction.failed?
+  end
+
+  def requires_action?
+    @transaction.present? && @transaction.requires_action?
+  end
+
+  def action_data
+    requires_action? && { client_secret: @transaction.payload['client_secret'] }
   end
 
   def failed_inventory?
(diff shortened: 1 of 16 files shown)


lib/order_processor.rb
```diff
@@ -36,7 +36,7 @@ def charge!
 
     deduct_inventory
     @transaction = PaymentService.capture_without_hold(construct_charge_params)
-    undeduct_inventory if @transaction.failed?
+    undeduct_inventory if @transaction.failed? || @transaction.requires_action?
```
joeyAghion: Minor: since this `failed? || requires_action?` logic is repeated, it may indicate the need for a helper like `successfully_processed?` or maybe the criteria for `failed?` should be expanded.

lib/order_processor.rb
```diff
@@ -46,6 +46,14 @@ def failed_payment?
     @transaction.present? && @transaction.failed?
   end
 
+  def requires_action?
+    @transaction.present? && @transaction.requires_action?
```
joeyAghion: `@transaction&.requires_action?`

lib/order_processor.rb
ashkan18: i kind of like the visibility of when we need to undeduct. But maybe once we refactored `OrderProcessor` to not be responsible for undeduct (as we planned it in our brainstorm) we can merge the check.

ashkan18:
yes @sweir27 , i was thinking of adding integration test as part of next story where we close the loop on a SCA required payment, but one could argue we could have integration test till this point and update it in next PR to close the loop :)

since i've started working on next PR on a branch, i may vote for adding integration test as part of that. what do you think?

ashkan18:
thanks all, going to #mergeOnGreen it

ashkan18:
going to point our MP review app to staging and delete Exchange review app created pointing to this branch.

PR #456, opened 2019-07-30
Expose currency_code in offer graphQL response

sweir27:
This adds `currency_code` as a field returned in an `Offer`'s graphQL response.

This will allow `amount` fields on `Offer` to be formatted correctly in metaphysics.


ashkan18:
we talked about this potentially causing N+1 if we get list of `LineItems` but since we don't do that this looks 👌 to me.

PR #459, opened 2019-07-31
Prevent Tax Calculation for Artworks Outside the United States (GALL-1942)

iskounen:
### Description

The purpose of this PR is to update the tax calculation logic in Exchange to prevent sales tax being calculated for an order (or offer-order) if the buyer is purchasing an artwork located outside of the United States. This is being done so that orders for artwork located in the UK aren't taxed once we enable NSO for UK partners.

### Diagram

This diagram explains what happens after a buyer enters their shipping information during the Buy Now flow, with a focus on how sales tax is calculated. This PR adds the section highlighted in red.

Buy now shipping and tax (1)

### DIFF
--- a/lib/tax/calculator_service.rb
+++ b/lib/tax/calculator_service.rb
@@ -18,7 +18,7 @@ class Tax::CalculatorService
     @seller_nexus_addresses = process_nexus_addresses!(nexus_addresses)
     @fulfillment_type = fulfillment_type
     @tax_client = tax_client
-    @artwork_location = artwork_location
+    @artwork_location = process_artwork_location!(artwork_location)
     @shipping_address = shipping_address
     @shipping_total_cents = shipping_total_cents
     @transaction = nil
@@ -107,5 +107,11 @@ class Tax::CalculatorService
 
   def validate_nexus_address!(nexus_address)
     raise Errors::ValidationError, :invalid_seller_address if nexus_address.region.nil?
+  end
+
+  def process_artwork_location!(artwork_location)
+    raise Errors::ValidationError, :no_taxable_addresses unless address_taxable?(artwork_location)
+
+    artwork_location
   end
 end
--- a/lib/tax/calculator_service.rb
+++ b/lib/tax/calculator_service.rb
@@ -18,7 +18,7 @@ class Tax::CalculatorService
     @seller_nexus_addresses = process_nexus_addresses!(nexus_addresses)
     @fulfillment_type = fulfillment_type
     @tax_client = tax_client
-    @artwork_location = process_artwork_location!(artwork_location)
+    @artwork_location = artwork_location
     @shipping_address = shipping_address
     @shipping_total_cents = shipping_total_cents
     @transaction = nil
@@ -29,7 +29,7 @@ class Tax::CalculatorService
   end
 
   def sales_tax
-    @sales_tax ||= UnitConverter.convert_dollars_to_cents(fetch_sales_tax.amount_to_collect)
+    address_taxable?(@artwork_location) ? fetch_sales_tax : 0
   rescue Taxjar::Error => e
     raise Errors::ProcessingError.new(:tax_calculator_failure, message: e.message)
   end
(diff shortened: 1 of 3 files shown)


lib/tax/calculator_service.rb
```diff
@@ -108,4 +108,10 @@ def process_nexus_addresses!(seller_nexus_addresses)
   def validate_nexus_address!(nexus_address)
     raise Errors::ValidationError, :invalid_seller_address if nexus_address.region.nil?
   end
+
+  def process_artwork_location!(artwork_location)
+    raise Errors::ValidationError, :no_taxable_addresses unless address_taxable?(artwork_location)
```
ashkan18: what do you think about instead of adding this method and raising an error we would update `sales_tax` method to return 0 if artwork location is not us? I think `no_taxable_address` is considered an error while this case is not really and error and its more our business logic, thinking of something like
```ruby
def sales_tax
  @sales_tax ||= if requires_tax_calculation?(@artwork_location)
     UnitConverter.convert_dollars_to_cents(fetch_sales_tax.amount_to_collect)
  else
    0
  end
```

iskounen: I am in favor of adding this check to the `sales_tax` method. I would also like to move the check for US addresses in `process_nexus_addresses` to `sales_tax` for consistency. What do you think about that?

iskounen:
@ashkan18 we have not set a flag on artworks to indicate that a price includes VAT as a part of this PR because we do not want to calculate sales tax for an artwork located in the UK, regardless of whether its price includes VAT or not. Do you think that this flag will become useful somewhere else?

Regarding the diagram, I'm reluctant to add it to docs because it will need to be maintained as we continue to build out Exchange. What I'd be interested in doing is seeing how we can reorganize our tests so that they describe the same behaviors that are outlined in this diagram.

ashkan18:
oh didn't know they are separate concepts, i think if price includes VAT we need to somehow flag it in Exchange it will be useful later for reporting purposes. That can be a separate PR though then, it seems not directly related to this.
As for this PR, if `artwork_location` is tracked on artwork version we should be good, if it's not, we may want to add a flag in Exchange reflecting why tax was calculated as 0 at that time. Otherwise it would be hard to figure out why tax was calculated as 0 later since artwork data can change.

Good point on the diagram. Its really helping visually to find out what happens when but i agree these tend to get out of date really fast

lib/tax/calculator_service.rb
jonallured: I agree, I think checking to see if the artwork's location requires a tax calculation at class init time is probably too soon.

In terms of adding this to `sales_tax`, I think it might be time for a refactoring here - that method is responsible for kicking off the calculation, caching the result, rescuing and raising errors and converting to cents. Being able to separate this out would probably make for a class that's easier to change!

jonallured:
I checked in Gravity and artwork location is an association:

https://github.com/artsy/gravity/blob/master/app/models/domain/artwork.rb#L130

I didn't see anything about locations on an artwork version, but  I do see artist ids:

https://github.com/artsy/gravity/blob/master/app/models/domain/artwork_version.rb#L9

And that's an association too, so I guess we could add location ids if we wanted? It's not a fool-proof audit trail because ids that point to deleted locations would land us back in a state where we can't be sure why the tax calculation resulted in a particular value.

But, is that bad? Do we need to be able to reconstruct the state of things at the time of tax calculation? I can see that today we can't prove why tax came out the way it did, but I'm not clear on the use case for why we'd need to show our work? Is this a requirement coming from finance?

iskounen:
@jonallured I made a first attempt at refactoring `sales_tax`, but I'd like to know more about what you had in mind. For now, I removed the caching (`@sales_tax ||=`) since `LineItemsTotals` and `OfferTotals` both cache the result of the code that calls this (more info [in this ticket](https://artsyproduct.atlassian.net/browse/GALL-2011)). I also moved the call to convert cents to dollars into the `fetch_sales_tax` method to make `sales_tax` easier to read.

How were you thinking about separating out the work that this method does?

iskounen:
@ashkan18 correct me if I'm wrong, but there's no requirement from finance to flag orders with a reason why sales tax was not calculated. I think we're doing it to avoid a potential situation that was identified while discussing how VAT will be handled in BNMO (specifically that we need a paper trail in case a partner's or an artwork's tax situation changes).

We can confirm whether this is necessary with Finance and Product, but I would also be interested to see how we can add notes or memos to orders with this kind of information.

jonallured:
For completeness-sake, @iskounen and I chatted IRL and came to the conclusion that the refactors he's got are perfect! 🎉

PR #461, opened 2019-08-01
Confirm payment intent

ashkan18:
# Change
In #455 we started supporting `requires_action` state of `PaymentIntent`. In this case, we notify clients about this state and provide them with `client_secret` for this specific PaymentIntent which will be used to call `stripe.js`'s method to open 3D auth process.

Once 3D auth was successful, we call Exchange to submit the order again. In this case, we don't want to create a new payment intent, we have to confirm the existing one assuming its in `processing` [state](https://stripe.com/docs/payments/intents#intent-statuses).

This PR checks to see if during submission, order has `external_charge_id` set, we will confirm that payment intent instead of creating a new one.

# Follow up
When we get to requires action case, it's possible that user instead of going through 3D auth, they decide to change their payment method, in this case we cant just reuse existing payment intent and we need to either update existing one to use new payment method or create a whole new payment intent.

# Ultimate follow up ;)
`OrderService` code is getting super complicated and hard to read, planning to do some cleanup and explorations to make it more readable, for now trying to unblock client side work with existing structure.

### DIFF
--- a/lib/order_processor.rb
+++ b/lib/order_processor.rb
@@ -25,7 +25,7 @@ class OrderProcessor
 
     deduct_inventory
     @transaction = if @order.external_charge_id
-      # here are are assuming this external_charge_id is a PaymentIntent id.
+      # here we are assuming this external_charge_id is a PaymentIntent id.
       # we already have a payment intent on this order
       PaymentService.confirm_payment_intent(@order.external_charge_id)
     else
--- a/lib/order_processor.rb
+++ b/lib/order_processor.rb
@@ -24,7 +24,12 @@ class OrderProcessor
     raise Errors::ValidationError, @validation_error unless valid?
 
     deduct_inventory
-    @transaction = PaymentService.hold_payment(construct_charge_params)
+    @transaction = if @order.external_charge_id
+      # we already have a payment intent on this order
+      PaymentService.confirm(@order.external_charge_id)
+    else
+      PaymentService.hold_payment(construct_charge_params)
+    end
     undeduct_inventory if @transaction.failed? || @transaction.requires_action?
   rescue Errors::InsufficientInventoryError
     undeduct_inventory
--- a/lib/order_processor.rb
+++ b/lib/order_processor.rb
@@ -26,7 +26,7 @@ class OrderProcessor
     deduct_inventory
     @transaction = if @order.external_charge_id
       # we already have a payment intent on this order
-      PaymentService.confirm(@order.external_charge_id)
+      PaymentService.confirm_payment_intent(@order.external_charge_id)
     else
       PaymentService.hold_payment(construct_charge_params)
     end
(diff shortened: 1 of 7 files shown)


lib/order_processor.rb
```diff
@@ -24,7 +24,12 @@ def hold!
     raise Errors::ValidationError, @validation_error unless valid?
 
     deduct_inventory
-    @transaction = PaymentService.hold_payment(construct_charge_params)
+    @transaction = if @order.external_charge_id
```
joeyAghion: Is the assumption here that `external_charge_id` refers to a payment intent?

lib/payment_service.rb
```diff
@@ -118,9 +136,12 @@ def self.generate_transaction_from_exception(exc, type, external_id: nil, extern
     )
   end
 
-  def self.transaction_from_payment_intent_failure(exc, capture:)
+  def self.transaction_from_payment_intent_failure(exc, capture: nil)
     body = exc.json_body
     pi = body[:error][:payment_intent]
+    transaction_type = unless capture.nil?
+      capture ? Transaction::CAPTURE : Transaction::HOLD
```
joeyAghion: This `capture.nil?` / `capture ? ...` logic is confusing. What are the possible values? Maybe a `transaction_type` can be passed into this method directly instead?

spec/integration/buy_now_sca_path.rb
```diff
@@ -0,0 +1,169 @@
...
+              trackingId: 'fedex-123',
+              estimatedDelivery: '2018-12-15'
+            }
+          })
+      end.to change(order.line_items.first.fulfillments, :count).by(1)
+      expect(order.reload).to have_attributes(state: Order::FULFILLED)
+      expect(order.line_items.first.fulfillments.first).to have_attributes(courier: 'fedex', tracking_id: 'fedex-123', estimated_delivery: Date.strptime('2018-12-15', '%Y-%m-%d'))
+    end
+  end
+end
```
joeyAghion: 👌

lib/order_processor.rb
ashkan18: yes, but we could add more defensive code around it.

lib/payment_service.rb
ashkan18: yes, maybe passing transaction type would make it cleaner. giving it a try 👍

lib/order_processor.rb
ashkan18: actually for now it may be easier to make that assumption till we actually move away from having `external_charge_id` on `order`.

ashkan18: going to add comment to clarify it.

lib/order_processor.rb
```diff
@@ -24,7 +24,13 @@ def hold!
...
+      # here are are assuming this external_charge_id is a PaymentIntent id.
```
sweir27: #superminor `here are are` --> `here we are`

ashkan18:
#mergeOnGreen

PR #463, opened 2019-08-05
Cancel payment intent instead of refund when un-captured

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-1379
We learned the hard way that when a payment intent is not captured yet, we need to use `payment_intent.cancel` and if payment intent was captured, then we need to first find the charge associated with that payment intent and use refund API to refund it.
We were under impression that we can always refund the charge associated with payment intent knowing that a uncaptured payment intent will still have charge associated with it.

# Solution
In case of `seller_lapse`, `seller_reject` and buyer reject (only for BN), we now call newly added `cancel_payment_intent` method instead of calling refund.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,7 +30,9 @@ Metrics/MethodLength:
     - "db/migrate/*"
 
 Metrics/ModuleLength:
-  Max: 151
+  Max: 160
+  Exclude:
+    - "spec/**/*"
 
 Style/Documentation:
   Enabled: false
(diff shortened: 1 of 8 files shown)


ashkan18:
this happened again today https://artsy.slack.com/archives/CB110C6LE/p1565122754032200 @lilyfromseattle @sweir27 when had a chance please review and merge

ashkan18:
going to merge this so we can have it in next deploy but please do review when had a chance later and I'll address comments in follow up PR

PR #464, opened 2019-08-05
clear external_charge_id when updating payment_method

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-1369

# Solution
Went with easier solution in which we basically clean `external_charge_id` on the order which would lead to create new payment intent when user tries to submit again with new payment method.


sweir27:
Oh! This seems like a nice solution.

PR #465, opened 2019-08-07
Fix passing empty hash

ashkan18:
# Problem
During #462 we attempted to pass shipping information to Stripe when available. With current code if order type is pickup, we send empty hash which seems to cause stripe raising 
```
Invalid Hash.
```

# Cause
Its hard to test this since this happens from Stripe end and we dont have a proper Stripe mock server.

# Solution
pass all keys but use safe call


ashkan18:
going to #mergeongreen since staging is broken

PR #466, opened 2019-08-08
#FF Add external_id and type to transaction events

ashkan18:
# Problem
We want to be able to fetch stripe details for a transaction when receiving a transaction event from consumer side. For that to work we need `external_id` and `external_type` of a transaction.

# Solution
Add those to `TransactionEvent.properties`.

cc: @iskounen @starsirius @bhoggard


ashkan18:
#mergeongreen since this is trivial

PR #467, opened 2019-08-08
#FF Include Shipping Info in Transaction Event Payload

iskounen:
This PR updates the `TransactionEvent` event to include shipping information in the data that it publishes to a messaging queue when a transaction fails. This change allows CRT to have access to the shipping address' country and name fields when they are notified of a failed transaction, which will help them determine if there is a potential fraud case faster.


ashkan18:
Nice!! 🙌

PR #469, opened 2019-08-12
Fixed checking for wrong payment intent state after SCA

ashkan18:
Discussion (hmm... or more like Monolog) here https://artsy.slack.com/archives/C9YNS4X32/p1565635031042300

going to self-merge


ashkan18:
#mergeongreen

PR #473, opened 2019-08-13
Add statsd for require action

ashkan18:
# Problem
Not much honestly, just like to be able to know how often require action happens.

# Solution
add statsd with `submit.requires_action`


ashkan18:
actually going to self-merge this since it _looks_ pretty innocence 🤷‍♂ #mergeOnGreen

PR #475, opened 2019-08-15
Refactor order_processor to be less dependent to db transactions

ashkan18:
# Problem
As we started building on top of existing code around submitting an `Order`, things got more complicated. Looking deeper at our process and after few brainstorming, it seems the part that complicates things the most is us having most of business logic during submit in a `order.submit!` or `order.approve!` block. Anything in that block is using `order.with_lock` and is part of one database transaction.
While having everything during wrapped in one db transaction helps our data integrity, because of external dependencies like inventory deduction in Gravity and charge processing in Stripe, we can easily get into a rollback scenario. Rollback case also gets pretty complicated because in some cases we want to store transactions in some cases we don't.

# Solution
Instead of wrapping everything in a transaction block, do things in more sequential way and provide a proper rollback method in case of rollback in different stages.

In this new approach we rely more on `OrderProcessor` which used to be this weird little service only dealing with inventory deduction and charge processing. Now `OrderProcessor`:
- provides separate methods for charge related methods and deduct and undeducting inventory
- provides new methods for setting totals on the order
- provides a `rollback!` method that can decide what needs to be rolled back at current stage.

The big difference in the flow is now we optimistically set the order in new state at the beginning of the process and if anything goes wrong we rollback the state to previous one. We used to not be able to go from `submitted` state to `pending` but with new approach if order processing fails after our optimistic state update, we'll be going back from `submitted` to `pending` and in case of approve, from `approved` to `submitted`. This should be ok since we don't post events yet for any of these.

# How confident are we? %87.3
As long as this PR doesn't change any request level spec, i feel %87.3 confident. The rest of confidence happens when we see things in real 🌏 with our own 👁.

https://twitter.com/i/status/1139548144390352897

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -66,7 +66,7 @@ class Order < ApplicationRecord
     SHIP = 'ship'.freeze
   ].freeze
 
-  ACTIONS = %i[abandon submit approve reject fulfill seller_lapse buyer_lapse refund].freeze
+  ACTIONS = %i[abandon rollback submit approve reject fulfill seller_lapse buyer_lapse refund].freeze
   ACTION_REASONS = {
     seller_lapse: REASONS[CANCELED][:seller_lapsed],
     buyer_lapse: REASONS[CANCELED][:buyer_lapsed],
@@ -247,6 +247,7 @@ class Order < ApplicationRecord
     machine = MicroMachine.new(state)
     machine.when(:abandon, PENDING => ABANDONED)
     machine.when(:submit, PENDING => SUBMITTED)
+    machine.when(:rollback, APPROVED => SUBMITTED, SUBMITTED => PENDING)
     machine.when(:approve, SUBMITTED => APPROVED)
     machine.when(:reject, SUBMITTED => CANCELED)
     machine.when(:seller_lapse, SUBMITTED => CANCELED)
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -66,7 +66,7 @@ class Order < ApplicationRecord
     SHIP = 'ship'.freeze
   ].freeze
 
-  ACTIONS = %i[abandon rollback submit approve reject fulfill seller_lapse buyer_lapse refund].freeze
+  ACTIONS = %i[abandon revert submit approve reject fulfill seller_lapse buyer_lapse refund].freeze
   ACTION_REASONS = {
     seller_lapse: REASONS[CANCELED][:seller_lapsed],
     buyer_lapse: REASONS[CANCELED][:buyer_lapsed],
@@ -247,7 +247,7 @@ class Order < ApplicationRecord
     machine = MicroMachine.new(state)
     machine.when(:abandon, PENDING => ABANDONED)
     machine.when(:submit, PENDING => SUBMITTED)
-    machine.when(:rollback, APPROVED => SUBMITTED, SUBMITTED => PENDING)
+    machine.when(:revert, APPROVED => SUBMITTED, SUBMITTED => PENDING)
     machine.when(:approve, SUBMITTED => APPROVED)
     machine.when(:reject, SUBMITTED => CANCELED)
     machine.when(:seller_lapse, SUBMITTED => CANCELED)
--- a/app/services/offer_service.rb
+++ b/app/services/offer_service.rb
@@ -83,10 +83,8 @@ module OfferService
 
     elsif order_processor.requires_action?
       order_processor.revert!
-      order_processor.set_external_payment!
       raise Errors::PaymentRequiresActionError, order_processor.action_data
     end
-    order_processor.set_external_payment!
     order_processor.on_success
     order
   end
--- a/app/services/offer_service.rb
+++ b/app/services/offer_service.rb
@@ -68,26 +68,26 @@ module OfferService
     order = offer.order
     order_processor = OrderProcessor.new(order, user_id, offer)
     raise Errors::ValidationError, order_processor.validation_error unless order_processor.valid?
-    order_processor.transition(:approve!)
-    order_processor.deduct_inventory!
-    if order_processor.failed_inventory?
-      order_processor.rollback!
+
+    order_processor.advance_state(:approve!)
+    unless order_processor.deduct_inventory
+      order_processor.revert!
       raise Errors::InsufficientInventoryError
     end
     order_processor.set_totals!
-    order_processor.charge!
+    order_processor.charge
     order_processor.store_transaction
     if order_processor.failed_payment?
-      order_processor.rollback!
+      order_processor.revert!
       raise Errors::FailedTransactionError.new(:capture_failed, order_processor.transaction)
+
     elsif order_processor.requires_action?
-      order_processor.rollback!
-      order_processor.set_payment!
-      raise Errors::PaymentRequiresActionError.new(order_processor.action_data)
-    else
-      order_processor.set_payment!
-      order_processor.set_follow_ups
+      order_processor.revert!
+      order_processor.set_external_payment!
+      raise Errors::PaymentRequiresActionError, order_processor.action_data
     end
+    order_processor.set_external_payment!
+    order_processor.on_success
     order
   end
 
--- a/app/services/offer_service.rb
+++ b/app/services/offer_service.rb
@@ -66,33 +66,29 @@ module OfferService
     raise Errors::ValidationError, :not_last_offer unless offer.last_offer?
 
     order = offer.order
-    order_processor = OrderProcessor.new(order, user_id)
+    order_processor = OrderProcessor.new(order, user_id, offer)
     raise Errors::ValidationError, order_processor.validation_error unless order_processor.valid?
-
-    order.approve! do
-      totals = OfferOrderTotals.new(offer)
-      order.update!(
-        transaction_fee_cents: totals.transaction_fee_cents,
-        commission_rate: totals.commission_rate,
-        commission_fee_cents: totals.commission_fee_cents,
-        seller_total_cents: totals.seller_total_cents
-      )
-      order_processor.charge!
-      raise Errors::InsufficientInventoryError if order_processor.failed_inventory?
-      # in case of failed transaction, we need to rollback this block,
-      # but still need to add transaction, so we raise an ActiveRecord::Rollback
-      raise ActiveRecord::Rollback if order_processor.failed_payment?
-
-      order.update!(external_charge_id: order_processor.transaction.external_id)
+    order_processor.transition(:approve!)
+    order_processor.deduct_inventory!
+    if order_processor.failed_inventory?
+      order_processor.rollback!
+      raise Errors::InsufficientInventoryError
     end
-    order.transactions << order_processor.transaction
-    PostTransactionNotificationJob.perform_later(order_processor.transaction.id, user_id)
-    raise Errors::FailedTransactionError.new(:capture_failed, order_processor.transaction) if order_processor.failed_payment?
-
-    OrderEvent.delay_post(order, Order::APPROVED, user_id)
-    OrderFollowUpJob.set(wait_until: order.state_expires_at).perform_later(order.id, order.state)
-    ReminderFollowUpJob.set(wait_until: order.state_expiration_reminder_time).perform_later(order.id, order.state)
-    Exchange.dogstatsd.increment 'order.approved'
+    order_processor.set_totals!
+    order_processor.charge!
+    order_processor.store_transaction
+    if order_processor.failed_payment?
+      order_processor.rollback!
+      raise Errors::FailedTransactionError.new(:capture_failed, order_processor.transaction)
+    elsif order_processor.requires_action?
+      order_processor.rollback!
+      order_processor.set_payment!
+      raise Errors::PaymentRequiresActionError.new(order_processor.action_data)
+    else
+      order_processor.set_payment!
+      order_processor.set_follow_ups
+    end
+    order
   end
 
   class << self
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -63,11 +63,9 @@ module OrderService
       raise Errors::FailedTransactionError.new(:charge_authorization_failed, order_processor.transaction)
     elsif order_processor.requires_action?
       order_processor.revert!
-      order_processor.set_external_payment!
       Exchange.dogstatsd.increment 'submit.requires_action'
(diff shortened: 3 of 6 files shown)


ashkan18:
Still need to update some service layer tests but this should be reviewable at this point.

app/services/offer_service.rb
```diff
@@ -66,33 +66,29 @@ def self.accept_offer(offer, user_id)
...
+    order_processor.charge!
+    order_processor.store_transaction
+    if order_processor.failed_payment?
+      order_processor.rollback!
+      raise Errors::FailedTransactionError.new(:capture_failed, order_processor.transaction)
+    elsif order_processor.requires_action?
+      order_processor.rollback!
+      order_processor.set_payment!
+      raise Errors::PaymentRequiresActionError.new(order_processor.action_data)
+    else
```
joeyAghion: I'm not sure you need this `else`. Since both the above conditions break the control flow by raising exceptions, you could just move the following 2 lines to outside of the `if`.

app/services/order_service.rb
```diff
@@ -48,40 +48,29 @@ def self.submit!(order, user_id)
...
-      )
-      order_processor.hold!
-      raise Errors::InsufficientInventoryError if order_processor.failed_inventory?
-      # in case of failed transaction, we need to rollback this block,
-      # but still need to add transaction, so we raise an ActiveRecord::Rollback
-      raise ActiveRecord::Rollback if order_processor.failed_payment? || order_processor.requires_action?
-
-      order.update!(external_charge_id: order_processor.transaction.external_id)
+    order_processor.transition(:submit!)
+    order_processor.deduct_inventory!
```
joeyAghion: This too looks like it would raise an error, but doesn't.

Minor refactoring idea: there are several places where you call a method and then call another to determine whether the first was successful. Instead, the first method could return `true` or `false` and the logic could be combined into lines like:

    unless order_processor.deduct_inventory do
      # rollback, etc.
    end

Or, if the logic is super-simple, even:

    order_processor.deduct_inventory || rollback...

app/services/order_service.rb
```diff
@@ -48,40 +48,29 @@ def self.submit!(order, user_id)
...
+    if order_processor.failed_payment?
+      order_processor.rollback!
+      raise Errors::FailedTransactionError.new(:charge_authorization_failed, order_processor.transaction)
+    elsif order_processor.requires_action?
+      order_processor.rollback!
+      order_processor.set_payment!
       Exchange.dogstatsd.increment 'submit.requires_action'
-      raise Errors::PaymentRequiresActionError, order_processor.action_data
+      raise Errors::PaymentRequiresActionError.new(order_processor.action_data)
+    else
```
joeyAghion: Similar to above, could you get rid of this `else`?

lib/order_processor.rb
```diff
@@ -10,42 +10,63 @@ class OrderProcessor
...
     @transaction = nil
     @deducted_inventory = []
     @validated = false
     @insufficient_inventory = false
+    @totals_set = false
+    @state_changed = false
+  end
+
+  def rollback!
+    undeduct_inventory! unless @deducted_inventory.empty?
```
joeyAghion: Slight preference for the positive version: `if @deducted_inventory.any?`

lib/order_processor.rb
```diff
@@ -76,19 +97,36 @@ def valid?
...
+    undeduct_inventory
+    @insufficient_inventory = true
+  end
+
+  def store_transaction
+    order.transactions << transaction
+    PostTransactionNotificationJob.perform_later(transaction.id, @user_id)
+  end
+
+  def set_payment!
```
joeyAghion: This is an unexpected name for a method that stores an `external_charge_id`. If it remains I'd consider renaming.

lib/order_processor.rb
```diff
@@ -76,19 +97,36 @@ def valid?
...
+    @order.update!(external_charge_id: transaction.external_id)
+  end
+
+  def set_follow_ups
```
joeyAghion: The work below is only partially about follow-ups, so I might rename this to `on_submit` or `on_success`.

lib/payment_service.rb
```diff
@@ -158,6 +158,9 @@ def self.generate_transaction_from_exception(exc, type, external_id: nil, extern
   def self.transaction_from_payment_intent_failure(exc, transaction_type: nil)
     body = exc.json_body
     pi = body[:error][:payment_intent]
+    transaction_type = unless capture.nil?
+      capture ? Transaction::CAPTURE : Transaction::HOLD
+    end
```
joeyAghion: A bit confusing! I don't have a great suggestion though.

ashkan18: oy.... this is merge conflict actually when i got latest master, the method already gets `transaction_type`, no need for this mess 🤦‍♂

lib/order_processor.rb
```diff
@@ -10,42 +10,63 @@ class OrderProcessor
...
+    order.rollback! if @state_changed
+  end
+
+  def transition(action)
+    order.send(action)
+    @state_changed = true
+  end
+
+  def set_totals!
+    order.line_items.each { |li| li.update!(commission_fee_cents: li.current_commission_fee_cents) } if order.mode === Order::BUY
```
starsirius: Is the use of `===` here (and below) intentional? I rarely use it in Ruby and, in terms of strictness, it seems to be the opposite of Javascript. _I think_ it's the the same when comparing two [String instances](https://ruby-doc.org/core-2.4.0/String.html#method-i-3D-3D-3D), but maybe using `==` is more explicit.

app/services/offer_service.rb
```diff
@@ -66,33 +66,29 @@ def self.accept_offer(offer, user_id)
...
-      raise Errors::InsufficientInventoryError if order_processor.failed_inventory?
-      # in case of failed transaction, we need to rollback this block,
-      # but still need to add transaction, so we raise an ActiveRecord::Rollback
-      raise ActiveRecord::Rollback if order_processor.failed_payment?
-
-      order.update!(external_charge_id: order_processor.transaction.external_id)
+    order_processor.transition(:approve!)
+    order_processor.deduct_inventory!
+    if order_processor.failed_inventory?
+      order_processor.rollback!
```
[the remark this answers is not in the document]
starsirius: Maybe `transition` and `rollback` can pair better by renaming them to something like `advance_state(action)` and `revert_state`.

lib/order_processor.rb
```diff
@@ -10,42 +10,63 @@ class OrderProcessor
...
     @validated = false
     @insufficient_inventory = false
+    @totals_set = false
+    @state_changed = false
+  end
+
+  def rollback!
+    undeduct_inventory! unless @deducted_inventory.empty?
+    reset_totals! if @totals_set
+    order.rollback! if @state_changed
```
starsirius: Do we want to unset `@state_changed`?

lib/order_processor.rb
```diff
@@ -10,42 +10,63 @@ class OrderProcessor
...
+      commission_fee_cents: totals.commission_fee_cents,
+      seller_total_cents: totals.seller_total_cents
+    )
+
+    @totals_set = true
+  end
+
+  def reset_totals!
+    order.line_items.each { |li|  li.update!(commission_fee_cents: nil) } if order.mode === Order::BUY
+    order.update!(transaction_fee_cents: nil, commission_rate: nil, commission_fee_cents: nil, seller_total_cents: nil)
```
starsirius: Similarly, maybe we want to unset `@totals_set`.

app/services/offer_service.rb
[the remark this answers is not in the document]
ashkan18: yes! love these new names!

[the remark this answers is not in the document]
ashkan18: @joeyAghion  i think i'd just call this `order_processor` method `revert` since it may not always revert the state

app/services/offer_service.rb
```diff
@@ -66,33 +66,29 @@ def self.accept_offer(offer, user_id)
...
-    order.transactions << order_processor.transaction
-    PostTransactionNotificationJob.perform_later(order_processor.transaction.id, user_id)
-    raise Errors::FailedTransactionError.new(:capture_failed, order_processor.transaction) if order_processor.failed_payment?
-
-    OrderEvent.delay_post(order, Order::APPROVED, user_id)
-    OrderFollowUpJob.set(wait_until: order.state_expires_at).perform_later(order.id, order.state)
-    ReminderFollowUpJob.set(wait_until: order.state_expiration_reminder_time).perform_later(order.id, order.state)
-    Exchange.dogstatsd.increment 'order.approved'
+    order_processor.set_totals!
+    order_processor.charge!
```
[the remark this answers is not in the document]
ashkan18: yes, we used `!` in Exchange also for methods that update the model, kind of like `merge` vs. `merge!`, that said, charge doesn't do that anymore though, so im going to remove `!`

app/services/offer_service.rb
```diff
@@ -66,33 +66,29 @@ def self.accept_offer(offer, user_id)
...
+    order_processor.store_transaction
+    if order_processor.failed_payment?
+      order_processor.rollback!
+      raise Errors::FailedTransactionError.new(:capture_failed, order_processor.transaction)
+    elsif order_processor.requires_action?
+      order_processor.rollback!
+      order_processor.set_payment!
+      raise Errors::PaymentRequiresActionError.new(order_processor.action_data)
+    else
+      order_processor.set_payment!
```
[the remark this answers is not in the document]
ashkan18: interesting, in that case, store transaction needs to know that if charge was failed don't update `external_charge_id` on the `order`, maybe name `set_payment` is wrong, maybe it should be `set_related_payment` or something?

lib/order_processor.rb
ashkan18: renamed it to `set_external_payment`

lib/order_processor.rb
ashkan18: no! this must be the cause of using those weird fonts on my vscode! https://github.com/tonsky/FiraCode

lib/order_processor.rb
ashkan18: oh good point!

ashkan18:
@joeyAghion @starsirius addressed most of the comments, still adding more tests to `order_processor` but please take a look when had  a chance, would love to have this on staging sooner to unblock offer SCA works.

app/models/order.rb
```diff
@@ -66,7 +66,7 @@ class Order < ApplicationRecord
     SHIP = 'ship'.freeze
   ].freeze
 
-  ACTIONS = %i[abandon submit approve reject fulfill seller_lapse buyer_lapse refund].freeze
+  ACTIONS = %i[abandon rollback submit approve reject fulfill seller_lapse buyer_lapse refund].freeze
```
joeyAghion: I'd still prefer a non-transaction-sounding name for this action.

app/services/offer_service.rb
```diff
@@ -66,33 +66,29 @@ def self.accept_offer(offer, user_id)
...
-        seller_total_cents: totals.seller_total_cents
-      )
-      order_processor.charge!
-      raise Errors::InsufficientInventoryError if order_processor.failed_inventory?
-      # in case of failed transaction, we need to rollback this block,
-      # but still need to add transaction, so we raise an ActiveRecord::Rollback
-      raise ActiveRecord::Rollback if order_processor.failed_payment?
-
-      order.update!(external_charge_id: order_processor.transaction.external_id)
+    order_processor.advance_state(:approve!)
```
joeyAghion: The `:approve!` argument is a little strange. I'd expect a state or action name instead.

app/services/offer_service.rb
```diff
@@ -66,33 +66,29 @@ def self.accept_offer(offer, user_id)
...
-    PostTransactionNotificationJob.perform_later(order_processor.transaction.id, user_id)
-    raise Errors::FailedTransactionError.new(:capture_failed, order_processor.transaction) if order_processor.failed_payment?
-
-    OrderEvent.delay_post(order, Order::APPROVED, user_id)
-    OrderFollowUpJob.set(wait_until: order.state_expires_at).perform_later(order.id, order.state)
-    ReminderFollowUpJob.set(wait_until: order.state_expiration_reminder_time).perform_later(order.id, order.state)
-    Exchange.dogstatsd.increment 'order.approved'
+    order_processor.set_external_payment!
+    order_processor.on_success
+    order
```
joeyAghion: Actually this is pretty clear now!

app/services/order_service.rb
```diff
@@ -48,40 +48,27 @@ def self.submit!(order, user_id)
...
+      order_processor.set_external_payment!
       Exchange.dogstatsd.increment 'submit.requires_action'
       raise Errors::PaymentRequiresActionError, order_processor.action_data
     end
-
-    OrderEvent.delay_post(order, Order::SUBMITTED, user_id)
-    OrderFollowUpJob.set(wait_until: order.state_expires_at).perform_later(order.id, order.state)
-    ReminderFollowUpJob.set(wait_until: order.state_expiration_reminder_time).perform_later(order.id, order.state)
-    Exchange.dogstatsd.increment 'order.submitted'
+    order_processor.set_external_payment!
```
joeyAghion: I still wonder if the `set_external_payment` work could just be combined into `store_transaction`. `OrderProcessor` isn't a bad place for some logic about other columns to update based on the results of a transaction.

lib/order_processor.rb
```diff
@@ -1,54 +1,62 @@
...
+    @totals_set = false
+    @state_changed = false
+  end
+
+  def revert!
+    undeduct_inventory! if @deducted_inventory.any?
+    reset_totals! if @totals_set
+    return unless @state_changed
+
+    order.rollback!
```
joeyAghion: I liked the suggestion from earlier of calling this `revert`.

ashkan18: yes, going to do this now.

app/services/order_service.rb
ashkan18: agreed, this may make next refactoring easier where i think we should switch to point to our own transaction instead of external id

ashkan18:
alright! i actually addressed all comments now

app/services/offer_service.rb
ashkan18: `approve` is the action name, do you mean removing `!` or make it string or both?

joeyAghion: It's minor so feel free to do whatever's convenient here.

ashkan18:
alright... merging....

PR #486, opened 2019-08-28
PURCHASE-1428: Get failed payment flow working with case where payment intent of a already confirmed setupintent fails when seller accepts offer

lilyfromseattle:
# [PURCHASE-1428] Get failed payment flow working with case where payment intent of a already confirmed setupintent fails when seller accepts offer

current failed payment flow works when a seller accepts an offer and payment doesn't go through, we send email to buyer about the failed payment and they fix their issue.
we need to make sure this works also for the case were we confirmed setupintent and later when seller accepts offer we tried to process off-session payment intent but that fails

# Problem
Currently, whenever an offer is directed to a seller, the PaymentIntent is always off-session. However, for our existing failed payment flow, the buyer is technically accepting their own offer, which will be to a seller. Because of this, the PaymentIntent created has `off-session: true` when its really on-session. 

# Solution
I changed the logic in `accept_offer` so that a payment is off session when the offer is directed to the seller AND the `user_id` equals the `seller_id`.

[PURCHASE-1428]: https://artsyproduct.atlassian.net/browse/PURCHASE-1428

### DIFF
--- a/app/services/offer_service.rb
+++ b/app/services/offer_service.rb
@@ -61,7 +61,8 @@ module OfferService
     end
     order_processor.set_totals!
     # if seller is accepting the offer, this is an off-session
-    off_session = offer.to_participant == Order::SELLER
+    # for failed payment flow buyer is accepting their own offer, so its on-session
+    off_session = offer.to_participant == Order::SELLER && user_id !== order.seller_id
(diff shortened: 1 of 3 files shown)


ashkan18:
🙌 merging

PR #487, opened 2019-08-29
support display of pounds

oxaudo:
Reopening this for now with the version that I already had.

@joeyAghion - we agreed to place the ticket for replacing this with gravity's money_helper at the time we start adding more things here.

### DIFF
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -1,8 +1,12 @@
 module AdminHelper
-  def format_money_cents(amount_cents, negate: false)
+  def convert_currency_code_to_symbol(currency_code)
+    currency_code == 'GBP' ? '£' : '$'
+  end
+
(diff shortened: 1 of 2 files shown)


oxaudo:
@starsirius - not sure why spec is failing here. And have hard time figuring out the local run as well. Maybe you can help me to take a look at that if you have a min at some point? Thanks!

oxaudo:
@starsirius  - this should be ok to go. Got the tests green:)

app/helpers/admin_helper.rb
```diff
@@ -1,8 +1,12 @@
 module AdminHelper
-  def format_money_cents(amount_cents, negate: false)
+  def convert_currency_code_to_symbol(currency_code)
+    currency_code == 'GBP' ? '£' : '$'
```
oxaudo: Keeping this one for now. Will create a card for exchange to potentially get rid of this helper and use https://github.com/artsy/money_helper instead

PR #489, opened 2019-08-29
PURCHASE-1428: Get failed payment flow working with case where payment intent of a already confirmed setupintent fails when seller accepts offer

lilyfromseattle:
# [PURCHASE-1428] 
current failed payment flow works when a seller accepts an offer and payment doesn't go through, we send email to buyer about the failed payment and they fix their issue.
we need to make sure this works also for the case were we confirmed setupintent and later when seller accepts offer we tried to process off-session payment intent but that fails

# Problem
fix failed payment mutation isn't set up to return a requires action response

# Solution
Make the mutation return requires action response if submitting during failed payment flow fails

[PURCHASE-1428]: https://artsyproduct.atlassian.net/browse/PURCHASE-1428


ashkan18:
nice! 👮 seems to be unhappy! other than that, looks great! thanks for quick update! #mergeongreen

PR #490, opened 2019-08-29
Fix for failing tests on master

lilyfromseattle:
Order by created_at before calling .last


ashkan18:
thanks for quick fix! 🙌

PR #492, opened 2019-08-30
Reuse payment intent on resubmitting during failed payment flow

ashkan18:
# Problem
Current fix failed payment flow is currently broken when using a card that needs SCA.

# Cause
When buyer uses a SCA card during failed payment flow, it currently creates a new payment intent instead of using the one that was just confirmed as part of SCA popup. 

# Solution
Update `order_processor.charge` to do same thing as `hold` where we check if there is already `external_charge_id` on the `order`, instead of creating new payment intent, confirm the existing one.
This change above though brought up a new bug where we realized currently we set `external_charge_id` on the `order` if transaction didn't fail, so in case transaction `requires_action` we would set it on the `order` so we can reuse it later. BUT in case of `off-session` (seller accepting offer) we don't want to reuse that payment intent anymore so we don't want to set it on the `Order`. We basically ended up updating `OrderProcessor.store_transaction` to get `off_session` as an input and if it was set it would not update `external_charge_id` when transaction is in the `require_action` state.

### DIFF
--- a/app/graphql/mutations/fix_failed_payment.rb
+++ b/app/graphql/mutations/fix_failed_payment.rb
@@ -16,7 +16,7 @@ class Mutations::FixFailedPayment < Mutations::BaseMutation
       order.last_transaction_failed? &&
       offer.id == order.last_offer.id
 
-    order = OrderService.set_payment!(order, credit_card_id)
+    order = OrderService.set_payment!(order, credit_card_id) if credit_card_id != order.credit_card_id
 
     # Note that the buyer might be 'accepting' their own offer here.
     # If they are, we know the seller attepted to accept it before
--- a/lib/order_processor.rb
+++ b/lib/order_processor.rb
@@ -56,7 +56,11 @@ class OrderProcessor
   end
 
   def charge(off_session = false)
-    @transaction = PaymentService.capture_without_hold(construct_charge_params.merge(off_session: off_session))
+    @transaction = if @order.external_charge_id
+      PaymentService.confirm_payment_intent(@order.external_charge_id)
+    else
+      PaymentService.capture_without_hold(construct_charge_params.merge(off_session: off_session))
+    end
   end
 
   def failed_payment?
--- a/lib/order_processor.rb
+++ b/lib/order_processor.rb
@@ -100,9 +100,9 @@ class OrderProcessor
     false
   end
 
-  def store_transaction
+  def store_transaction(off_session = false)
     order.transactions << transaction
-    order.update!(external_charge_id: transaction.external_id) unless transaction.failed?
+    order.update!(external_charge_id: transaction.external_id) unless transaction.failed? || (transaction.requires_action? && off_session)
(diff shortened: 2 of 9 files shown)


app/graphql/mutations/fix_failed_payment.rb
```diff
@@ -16,7 +16,7 @@ def resolve(offer_id:, credit_card_id:)
       order.last_transaction_failed? &&
       offer.id == order.last_offer.id
 
-    order = OrderService.set_payment!(order, credit_card_id)
```
ashkan18: we need to reset payment method on the order only if the input one is different than current one, otherwise it would reset external_charge_id on the order.

lib/order_processor.rb
```diff
@@ -96,9 +100,9 @@ def deduct_inventory
     false
   end
 
-  def store_transaction
+  def store_transaction(off_session = false)
     order.transactions << transaction
-    order.update!(external_charge_id: transaction.external_id) unless transaction.failed?
+    order.update!(external_charge_id: transaction.external_id) unless transaction.failed? || (transaction.requires_action? && off_session)
```
ashkan18: hoping we could make this logic more readable. but this makes sure that in case transaction was off-session in case it got to `require_action` state, we don't update the order's `external_charge_id` since we can't reuse this transaction later.

spec/integration/fix_failed_payment_flow_with_sca_card.rb
```diff
@@ -0,0 +1,180 @@
+require 'rails_helper'
+require 'support/gravity_helper'
+require 'support/taxjar_helper'
+
+# create order -> set initial offer -> set payment to a card that always requires 3d auth -> set shipping -> submit offer -> seller accept but it fails processing
```
ashkan18: this path doesn't fit in one line anymore 🤦‍♂

spec/integration/fix_failed_payment_flow_with_sca_card.rb
```diff
@@ -0,0 +1,180 @@
...
+        commission_fee_cents: 50_00,
+        external_charge_id: nil
+      )
+
+      # Buyer resubmits with same card now its on-session needs sca
+      prepare_payment_intent_create_failure(status: 'requires_action', id: 'pi_2')
+      expect do
+        buyer_client.execute(OfferQueryHelper::FAILED_PAYMENT_QUERY, input: { offerId: offer.id.to_s, creditCardId: 'cc-1' })
+      end.to change(order.transactions, :count).by(1)
+      # @TODO: it sets external charge id on the order
```
sweir27: Why is this a TODO?

ashkan18: ooh good catch, adding the test now

ashkan18:
going to #mergeongreen this so we have things ready for QA session today but please review and i will address things in follow up PRs

PR #494, opened 2019-09-09
PURCHASE-1474: Wrong expiration on offers with failed payment flow

ashkan18:
# Problem
Fixes https://artsyproduct.atlassian.net/browse/PURCHASE-1474
As part of [the famous "refactor"](https://github.com/artsy/exchange/pull/475), its possible for an order to go to some state and then go back to previous one of the attempt to create charge or deduct failed on the order.

This seems to have cause a new bug where we update the `state_expires_at` on the orders and extend them in failed payment flow. Meaning if the order's expiration was tomorrow, and seller tries to accept the offer, order goes to `approved` for a sec, charge fails, and order goes back to `submitted` state, but because of new transition to `submitted` expiration will now be 3 days from now instead of original 1 day.

# Solution
In `OrderProcessor` before transitioning order state, keep track of current `state_expires_at` and if order failed and we wanted to `revert!` update `OrderProcessor.revert!` to set expiration back to original one.

### DIFF
--- a/spec/integration/fix_failed_payment_flow_with_sca_card.rb
+++ b/spec/integration/fix_failed_payment_flow_with_sca_card.rb
@@ -112,7 +112,7 @@ describe Api::GraphqlController, type: :request do
         commission_fee_cents: 50_00
       )
 
-      submitted_state_expiration = order.state_expires_at # setting things up to test for https://artsyproduct.atlassian.net/browse/PURCHASE-1474
+      submitted_state_expiration = order.state_expires_at # setting things up to make sure we don't reset expiration when payment fails.
 
       # seller accepts offer but off-session charge fails
       prepare_payment_intent_create_failure(status: 'requires_action')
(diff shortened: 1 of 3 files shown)


spec/integration/fix_failed_payment_flow_with_sca_card.rb
```diff
@@ -112,6 +112,8 @@
         commission_fee_cents: 50_00
       )
 
+      submitted_state_expiration = order.state_expires_at # setting things up to test for https://artsyproduct.atlassian.net/browse/PURCHASE-1474
```
sweir27: I think we could describe the issue briefly instead of linking to a Jira ticket (which may be dead someday).

joeyAghion: +1

ashkan18:
thanks all! addressed @sweir27 's comments, going to #mergeongreen and later follow up with possible more refactor.

PR #496, opened 2019-09-10
Fix issue with offer page in admin

ashkan18:
# Problem
currently order admin page is broken for offers

# Cause
We are trying to access `currency_code` on `offer` but thats set on `order`.

# Solution
Switch to `offer.order.currency_code`.

cc: @oxaudo

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -287,7 +287,7 @@ ActiveAdmin.register Order do
 
         order.offers.submitted.order(created_at: :desc).each do |offer|
           date = offer.created_at
-          amount = format_money_cents(offer.amount_cents, currency_code: offer.currency_code)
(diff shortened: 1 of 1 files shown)


ashkan18:
going to #mergeongreen it

app/admin/orders.rb
```diff
@@ -287,7 +287,7 @@
 
         order.offers.submitted.order(created_at: :desc).each do |offer|
           date = offer.created_at
-          amount = format_money_cents(offer.amount_cents, currency_code: offer.currency_code)
+          amount = format_money_cents(offer.amount_cents, currency_code: offer.order.currency_code)
```
oxaudo: Oh :( Thanks @ashkan18 ! Did not even realize that offers are children of orders. I thought those are like brothers and sisters!

PR #498, opened 2019-09-10
GALL-2120: Relax admin check

jonallured:
In order to allow admins into the order pages in CMS, they have to be able to see the order data at the GQL layer. Without a change like this nothing would be returned to Volt for them to look at!

Note: I believe that what this PR does is allow admins access to query for orders, but does not allow them access to mutate them, but would love confirmation of this!

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -76,12 +76,12 @@ class Types::QueryType < Types::BaseObject
     context[:current_user][:roles].include?('trusted')
   end
 
-  def sales_admin?
-    context[:current_user][:roles].include?('sales_admin')
+  def admin?
+    context[:current_user][:roles].include?('admin')
   end
 
   def validate_order_request!(order)
-    return if trusted? || sales_admin? ||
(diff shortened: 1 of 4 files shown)


ashkan18:
thanks! 🙌 yes, this would only open up the query portion of it, mutations use these helper methods 👇 
https://github.com/artsy/exchange/blob/500bba6835e86466aa33663d7d0abc3a41543350/app/graphql/mutations/submit_order.rb#L10

#mergeongreen

a nice followup to this would be adding a new role for `liaisons` so we can target them specifically on these cases, `admin` role basically opens this up to anyone at Artsy.

starsirius:
Also note that Exchange may authorize individual fields. For example, [commission fee](https://github.com/artsy/exchange/blob/6952cffa30f0587cfb37b068a6b85e2585f570d4/app/graphql/types/order_interface.rb#L14) is seller only so when admin views it in CMS, it is displayed "null" or blank.

It's a confusing admin experience and we have a [ticket](https://artsyproduct.atlassian.net/browse/GALL-1655) to track that.

PR #499, opened 2019-09-11
Introduce new other payment method

ashkan18:
We need this in case that transaction went offline and we are not sure if it was paid by wire or not.


sweir27:
#mergeongreen

PR #501, opened 2019-09-12
Restrict order access to just trusted admin roles

jonallured:
Hot on the heels of my other PR #498, I'm following up here with one that restricts order access to admins with either `partner_support` or `sales_admin` roles. While I was at it, I added tests for both of these.

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -76,12 +76,13 @@ class Types::QueryType < Types::BaseObject
     context[:current_user][:roles].include?('trusted')
   end
 
-  def admin?
-    context[:current_user][:roles].include?('admin')
+  def trusted_admin?
+    user_roles = context[:current_user][:roles]
+    ArtsyAuthToken.trusted_admin?(user_roles)
   end
 
   def validate_order_request!(order)
-    return if trusted? || admin? ||
+    return if trusted? || trusted_admin? ||
               (order.buyer_type == Order::USER && order.buyer_id == context[:current_user][:id]) ||
               (order.seller_type != Order::USER && context[:current_user][:partner_ids].include?(order.seller_id))
 
@@ -94,7 +95,7 @@ class Types::QueryType < Types::BaseObject
   end
 
   def validate_orders_request!(params)
-    return if trusted? || admin?
+    return if trusted? || trusted_admin?
 
     if params[:buyer_id].present?
       raise ActiveRecord::RecordNotFound unless params[:buyer_id] == context[:current_user][:id]
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -77,7 +77,7 @@ class Types::QueryType < Types::BaseObject
   end
 
   def trusted_admin?
-    user_roles = context[:current_user][:roles]
+    user_roles = Array(context[:current_user][:roles])
     ArtsyAuthToken.trusted_admin?(user_roles)
   end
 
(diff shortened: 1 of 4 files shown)


spec/controllers/api/requests/order_query_request_spec.rb
```diff
@@ -397,7 +397,26 @@
       end
 
       context "sales admin accessing another account's order" do
-        let(:jwt_roles) { 'admin' }
```
ashkan18: should we keep a test for `admin` but make sure it actually fails?

jonallured:
Yeah, good call, I'll get a commit in that adds a test for "just" admins. 😝 

In terms of deployment, I don't have anything in Volt deployed, so this is mostly just setup work. I think we can merge this and roll it out to production on exchange without affecting anything else, right?

My thought was to get this merged and then I'd add "partner_support" to one of my test accounts and then that would allow me to work on a PR for Volt that actually uses this new access. Does that make sense to you?

ashkan18:
would volt need to change because of this? Currently liaisons are able to see the orders page in CMS, im guessing with this getting deployed they won't be able to do that anymore till they are added to `partner_support`, right? maybe thats ok?

jonallured:
No, we only allow `sales_admin` users into the orders pages in CMS.

https://github.com/artsy/volt/blob/master/app/controllers/application_controller.rb#L196-L198
https://github.com/artsy/volt/blob/master/app/models/feature.rb#L141

So unless I'm missing something, there would need to be a few updates in CMS to use the new access provided here.

ashkan18:
oh cool! then this makes sense! merging! thanks for explaining!

app/graphql/types/query_type.rb
```diff
@@ -76,12 +76,13 @@ def trusted?
     context[:current_user][:roles].include?('trusted')
   end
 
-  def admin?
-    context[:current_user][:roles].include?('admin')
+  def trusted_admin?
+    user_roles = Array(context[:current_user][:roles])
```
starsirius: @jonallured @ashkan18 I believe `context[:current_user][:roles]` is a [comma delimited string](https://github.com/artsy/gravity/blob/6d021e21f91e69898802e7ecf9eb3fa06ac36bfa/app/models/util/user_trust.rb#L37), so for example, a typical sales admin would be `'admin,sales_admin'`. I wonder if the array conversion and comparison would actually work. For example, if we change [this jwt roles](https://github.com/artsy/exchange/pull/501/files#diff-7263081cce0e1c7060fdca29ad4d9ad8R416) to `'admin,sales_admin'`, maybe the spec will fail?

jonallured: Oh yikes, I didn't realize this! I'll check this after lunch and see what's happening on staging. Good catch!

lib/artsy_auth_token.rb
```diff
@@ -8,7 +13,7 @@ def initialize(jwt)
   def admin?
     return false unless @decoded_token
 
-    @decoded_token['roles'].include? ADMIN_ROLE
+    self.class.trusted_admin?(@decoded_token['roles'])
```
starsirius: @jonallured If we also want to modify this authorization check to the new trusted admin roles (looks like it's used for the [active admin access](https://github.com/artsy/exchange/blob/695936f788fe818e196f969699376cb339f575a8/app/controllers/concerns/auth_handler.rb#L16)), let's maybe rename this method to `trusted_admin?` to avoid confusion?

app/graphql/types/query_type.rb
ashkan18: yes, this seems to have caused issues with exchange admin, doing a quick fix for now.

PR #503, opened 2019-09-13
Fix role checking

jonallured:
The roles property is a string that is a comma separated list of roles. Need to split it up into an array in order to compare against trusted roles.


spec/controllers/api/requests/order_query_request_spec.rb
```diff
@@ -413,7 +413,7 @@
       end
 
       context "sales admin accessing another account's order" do
-        let(:jwt_roles) { 'sales_admin' }
+        let(:jwt_roles) { 'admin,sales_admin' }
```
starsirius: 💪

lib/artsy_auth_token.rb
```diff
@@ -13,7 +13,8 @@ def initialize(jwt)
   def admin?
     return false unless @decoded_token
 
-    self.class.trusted_admin?(@decoded_token['roles'])
+    user_roles = @decoded_token['roles'].split(',')
+    self.class.trusted_admin?(user_roles)
```
starsirius: Left a comment in the [previous PR](https://github.com/artsy/exchange/pull/501/files#r324286920) about change of definition for this method.

jonallured:
Fixed with #504

PR #506, opened 2019-09-13
Fix role checking

jonallured:
The roles property is a string that is a comma separated list of roles. Need to split it up into an array in order to compare against trusted roles. I also updated a spec to more closely match how roles are passed in (as a string that's actually a comma separated list of roles).

Note: this is a re-open of #503 after rebasing.

### DIFF
--- a/spec/controllers/api/requests/order_query_request_spec.rb
+++ b/spec/controllers/api/requests/order_query_request_spec.rb
@@ -413,7 +413,7 @@ describe Api::GraphqlController, type: :request do
       end
 
       context "sales admin accessing another account's order" do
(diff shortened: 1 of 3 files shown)


spec/controllers/api/requests/order_query_request_spec.rb
```diff
@@ -413,7 +413,7 @@
       end
 
       context "sales admin accessing another account's order" do
-        let(:jwt_roles) { 'sales_admin' }
+        let(:jwt_roles) { 'admin,sales_admin' }
```
starsirius: Thank you! 💯

PR #510, opened 2019-10-08
Fix fail confirm payment intent issue

ashkan18:
# Problem
When a user is presented with SCA and does not go through with it (cancel or fail it), when they resubmit the order, the order goes to `submitted` state which means we send emails to both side and all that jazz without actually processing the charge.

# Cause
In this case, Order already has a payment intent assigned to it on its `external_charge_id` attribute, so when user resubmits the order, we try to `confirm` that existing payment intent, this leads to a `Stripe::InvalidRequestError` which we were not rescuing from. Since we are doing a optimistic `submit` in our latest code we were not reverting the order.

# Solution
Rescue from `Stripe::InvalidRequest` in `PaymentService` and create a failed transaction. This way we will revert in `order_service.submit!` and order should be back to `pending` state.

We also added a _catch all_ in `order_service.submit` which will always revert in case anything went wrong.

# Followup
Check other places we feel optimistic about things.

### DIFF
--- a/lib/payment_service.rb
+++ b/lib/payment_service.rb
@@ -79,7 +79,7 @@ module PaymentService
     )
   rescue Stripe::CardError => e
     transaction_from_payment_intent_failure(e)
-  rescue Stripe::InvalidRequestError => e
+  rescue Stripe::StripeError => e
     transaction_from_payment_intent_failure(e)
   end
 
--- a/lib/payment_service.rb
+++ b/lib/payment_service.rb
@@ -79,6 +79,8 @@ module PaymentService
     )
   rescue Stripe::CardError => e
     transaction_from_payment_intent_failure(e)
+  rescue Stripe::InvalidRequestError => e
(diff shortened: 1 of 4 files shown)


lib/payment_service.rb
```diff
@@ -79,6 +79,8 @@ def self.confirm_payment_intent(payment_intent_id)
     )
   rescue Stripe::CardError => e
     transaction_from_payment_intent_failure(e)
+  rescue Stripe::InvalidRequestError => e
+    transaction_from_payment_intent_failure(e)
```
joeyAghion: There are potentially other kinds of Stripe errors. Should we handle `Stripe::StripeError` instead?

https://github.com/stripe/stripe-ruby/blob/c206a3cc0995954f112768995705b8db649b8c85/lib/stripe/errors.rb#L4

ashkan18: yes, makes sense. changing it.

PR #512, opened 2019-10-08
Make sure payment service returns failed transaction on failed confirm

ashkan18:
# Problem
We fixed a scary bug in #510 where we now check for other Stripe errors. But in case payment intent is not even in the state that we attempt confirming (which seems to be the bug we actually ran into) while our previous fix actually works in this case since we added the catch all, at the moment we don't track the fact that we attempted to confirm the payment intent.

# Solution
Instead of raising `ProcessingError`, actually create a proper `Transaction` in failed state so we can track the fact that user attempted to confirm a payment intent not ready in to be confirmed.

# Extra Changes
- In `OrderService` as part of #510 we do a `order_processor.revert` in our new catch all block, this potentially means we call revert twice since we get into this block in case of failed payment and requires action, remove the revert from those 2 checks. While current approach is still safe since `revert` is smart enough to not double revert, new approach seems safer.
- Fix https://sentry.io/organizations/artsynet/issues/851534327

### DIFF
--- a/app/jobs/undeduct_line_item_inventory_job.rb
+++ b/app/jobs/undeduct_line_item_inventory_job.rb
@@ -1,7 +1,7 @@
 class UndeductLineItemInventoryJob < ApplicationJob
   queue_as :default
   discard_on(Errors::ValidationError) do |job, _error|
-    Rails.warn.info "Line item #{job.line_item.id} has deleted artwork, skipping inventory undeduction"
+    Rails.logger.warn("Line item #{job.line_item.id} has deleted artwork, skipping inventory undeduction")
   end
 
   attr_accessor :line_item
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -58,11 +58,9 @@ module OrderService
     order_processor.hold
     order_processor.store_transaction
 
-    if order_processor.failed_payment?
-      order_processor.revert!
-      raise Errors::FailedTransactionError.new(:charge_authorization_failed, order_processor.transaction)
-    elsif order_processor.requires_action?
-      order_processor.revert!
+    raise Errors::FailedTransactionError.new(:charge_authorization_failed, order_processor.transaction) if order_processor.failed_payment?
+
+    if order_processor.requires_action?
       Exchange.dogstatsd.increment 'submit.requires_action'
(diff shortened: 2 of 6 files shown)


app/jobs/undeduct_line_item_inventory_job.rb
```diff
@@ -1,7 +1,7 @@
 class UndeductLineItemInventoryJob < ApplicationJob
   queue_as :default
   discard_on(Errors::ValidationError) do |job, _error|
-    Rails.warn.info "Line item #{job.line_item.id} has deleted artwork, skipping inventory undeduction"
+    Rails.logger.warn("Line item #{job.line_item.id} has deleted artwork, skipping inventory undeduction")
```
joeyAghion: 🙏

lib/payment_service.rb
```diff
@@ -65,7 +65,16 @@ def self.refund_payment(external_id)
...
+    if payment_intent.status != 'requires_confirmation'
+      return Transaction.new(
+        external_id: payment_intent_id,
+        external_type: Transaction::PAYMENT_INTENT,
+        transaction_type: Transaction::CONFIRM,
+        failure_code: 'cannot_confirm',
+        failure_message: 'Payment intent is not in a confirmable state',
+        status: Transaction::FAILURE
+      )
+    end
```
joeyAghion: This feels very "special." Maybe it should use the `transaction_from_payment_intent_failure` helper or a new one like it to keep this method focused on the happy path?

app/services/order_service.rb
```diff
@@ -58,11 +58,9 @@ def self.submit!(order, user_id)
     order_processor.hold
     order_processor.store_transaction
 
-    if order_processor.failed_payment?
-      order_processor.revert!
-      raise Errors::FailedTransactionError.new(:charge_authorization_failed, order_processor.transaction)
-    elsif order_processor.requires_action?
-      order_processor.revert!
```
ashkan18: @joeyAghion it doesn't show up in original diff here but wanted to get your specific opinion on removing `order_processor.revert!` from here and rely on catch all on line 71

joeyAghion: I don't see any problems with it [that I could tell from reading the code]. It seems equivalent or better.

PR #514, opened 2019-10-09
Add external_charge_id to order event

ashkan18:
# Change
We need this value on the APR side to be able to fetch payment intent from Stripe.

cc: @sepans @zephraph going to self-merge since we talked about this and trying to test this on staging since APR side was already merged with this assumption.


ashkan18:
#mergeongreen

PR #516, opened 2019-10-17
Add artist ids and names to charge metadata

ashkan18:
# Problème
https://artsyproduct.atlassian.net/browse/PURCHASE-1567
We want to be able to add Radar rules possibly based on artists. We can do this by adding `artist_ids` and `artist_names` to metadata passed to Stripe.

Nous voulons pouvoir ajouter des règles de radar basées éventuellement sur des artistes. Nous pouvons le faire en ajoutant `artist_ids` et` artist_names` aux métadonnées transmises à Stripe.

# Soluzione
Add `artist_ids` and `artist_names` for this order to metadata passed to stripe.

Aggiungi `artist_ids` e` artist_names` per questo ordine ai metadati passati a stripe.

### DIFF
--- a/lib/line_item_helper.rb
+++ b/lib/line_item_helper.rb
@@ -1,9 +1,7 @@
 module LineItemHelper
   def artwork
     @artwork ||= begin
-      artwork = Gravity.get_artwork(artwork_id)
-      validate_artwork!(artwork)
-      artwork
+      Gravity.get_artwork(artwork_id) || raise(Errors::ValidationError, :unknown_artwork)
     end
   end
 
@@ -31,12 +29,5 @@ module LineItemHelper
 
   def current_commission_fee_cents
     total_list_price_cents * order.current_commission_rate
-  end
-
-  private
-
-  def validate_artwork!(artwork)
-    raise Errors::ValidationError, :unknown_artwork unless artwork
-    raise Errors::ValidationError, :missing_artwork_location if artwork[:location].blank?
   end
 end
--- a/lib/order_helper.rb
+++ b/lib/order_helper.rb
@@ -12,7 +12,11 @@ module OrderHelper
   end
 
   def artworks
-    @artworks ||= line_items.uniq.map { |li| [li.artwork_id, li.artwork] }.to_h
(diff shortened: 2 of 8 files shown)


lib/line_item_helper.rb
```diff
@@ -32,11 +30,4 @@ def artwork_location
   def current_commission_fee_cents
     total_list_price_cents * order.current_commission_rate
   end
-
-  private
-
-  def validate_artwork!(artwork)
-    raise Errors::ValidationError, :unknown_artwork unless artwork
-    raise Errors::ValidationError, :missing_artwork_location if artwork[:location].blank?
```
sweir27: Do we feel comfortable not validating the location here? Not sure why this was removed.

lib/order_helper.rb
```diff
@@ -12,7 +12,11 @@ def assert_credit_card
   end
 
   def artworks
-    @artworks ||= line_items.uniq.map { |li| [li.artwork_id, li.artwork] }.to_h
+    @artworks ||= line_items.map(&:artwork)
+  end
+
+  def artists
+    @artists ||= artworks.map { |a| a[:artist] }
```
sweir27: Note that this doesn't handle _multiple_ artists.

lib/line_item_helper.rb
ashkan18: so the validation for actual existence of artwork is moved up to where we used to call this method. Artwork location validation also happens when we actually ask for artwork's location. I think that should be ok but can double check tomorrow

sweir27: Anything to verify here?

Otherwise this is 👍 to merge!

ashkan18: trusting this spec https://github.com/artsy/exchange/blob/aff0e3d87d36e972da91282b9c3a8c92ea959cb3/spec/controllers/api/requests/set_shipping_mutation_request_spec.rb#L238 going to merge this PR

lib/order_helper.rb
ashkan18: yes, thought about that, for now went with easier route, we can add that later if needed.

PR #522, opened 2019-11-05
Fix admin offer cancelation issue

ashkan18:
# Problem
When an offer order is canceled/rejected, we still queue jobs to undeduct the inventory. This is not correct because for an offer order at that stage we haven't deducted the inventory yet.
Because of this on the Gravity's end we may end up with artworks with `{"sold":-1,"count":2}` if an offer order on a unique artwork is canceled.

Jira: https://artsyproduct.atlassian.net/browse/PURCHASE-1641
Original slack conversation: https://artsy.slack.com/archives/CB110C6LE/p1572892624036100

# Solution
Only process undeduct if order is a `BUY` order.

Also added a new cancelation reason for `admin_canceled` to reflect the reality of the case that order was canceled by an admin and not by buyer.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -214,6 +214,11 @@ type BuyerRejectOfferPayload {
 
 enum CancelReasonTypeEnum {
   """
+  cancelation reason is: admin_canceled
+  """
+  ADMIN_CANCELED
+
+  """
   cancelation reason is: buyer_lapsed
   """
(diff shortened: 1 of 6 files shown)


ashkan18:
@sweir27 this should be good, added another button for buyer reject case

ashkan18:
@sweir27 going to merge since i addressed your comment 🙌

PR #525, opened 2019-11-12
Fix missing shipping_region in TransactionEvent.properties.order

ashkan18:
# Problem
Looking at APR events, we don't see shipping region for any of our failed transaction events.

# Cause
`shipping_region` was not in list of order fields exposed in `TransactionEvent`.

# Solution
Add it.


sweir27:
Cool! #mergeongreen

PR #528, opened 2019-11-16
remove usages of charge api

ashkan18:
Removing unused code refering to Stripe's charge API


ashkan18:
self-merging this one since this is just removing unused legacy code

PR #531, opened 2019-11-18
Ensure user can retry SCA after canceling

zephraph:
Fixes [PURCHASE-1663](https://artsyproduct.atlassian.net/browse/PURCHASE-1663)

There was an edge-case in the SCA flow that when we tried to confirm the payment intent we'd always error if the intent wasn't in the `requires_confirmation` state. If a user canceled the SCA dialog or came back later to complete their purchase then the confirmation process would happen but with a `requires_action` intent state which would trigger a failure. Essentially it made it where once a user canceled SCA they could never complete their purchase unless they updated their payment information. 

Now during the  confirmation process, if the intent is  in the `requires_action` state  it retriggers the payment intent flow and will reinvoke the SCA process.

### DIFF
--- a/lib/payment_service.rb
+++ b/lib/payment_service.rb
@@ -56,7 +56,7 @@ module PaymentService
 
   def self.confirm_payment_intent(payment_intent_id)
     payment_intent = Stripe::PaymentIntent.retrieve(payment_intent_id)
-    return payment_intent_transaction_failure(payment_intent_id: payment_intent_id, transaction_type: Transaction::CONFIRM, failure_code: 'cannot_confirm', failure_message: 'Payment intent is not in a confirmable state') if payment_intent.status != 'requires_confirmation'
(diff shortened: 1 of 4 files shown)


lib/payment_service.rb
```diff
@@ -174,14 +174,15 @@ def self.transaction_from_payment_intent_failure(exc, transaction_type: nil)
     )
   end
 
-  def self.payment_intent_transaction_failure(payment_intent_id:, transaction_type:, failure_code:, failure_message:)
+  def self.payment_intent_transaction_failure(payment_intent)
```
sweir27: I'm assuming this method was only invoked from one place based on the changes to make it less generic.

If that's the case, we could consider renaming it to be more explicit: `payment_intent_confirmation_failure`

zephraph:
@sweir27 I renamed the failure method as you suggested. This should be good to go once we're green.

PR #535, opened 2019-12-18
Refactor: new service style for approve

ashkan18:
# Problem
This is tech-debt left from our attempts to move old style services to new style. We still had to do it for approve and cancelation services.

This PR moves `OrderApproveService` to new style by moving the logic to a class method in `OrderService` and in `approve!` method.

https://artsyproduct.atlassian.net/browse/PURCHASE-901


sweir27:
This does seem simpler!

Just one more to go 🎉 !!

PR #537, opened 2019-12-18
[PURCHASE-1656] Adds phone number to set_shipping mutation pickup mode

sepans:
`setShippng` mutation when `fulfillment_type` is `PICKUP` now processes and saves phone number under order object as `buyer_phone_number`

Fixes [PURCHASE-1656](https://artsyproduct.atlassian.net/browse/PURCHASE-1656)

### DIFF
--- a/app/services/order_service.rb
+++ b/app/services/order_service.rb
@@ -24,7 +24,7 @@ module OrderService
 
     order_shipping = OrderShipping.new(order)
     case fulfillment_type
-    when Order::PICKUP then order_shipping.pickup!
+    when Order::PICKUP then order_shipping.pickup!(shipping)
     when Order::SHIP then order_shipping.ship!(shipping)
     end
     order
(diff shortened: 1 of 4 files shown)


app/services/order_service.rb
```diff
@@ -24,7 +24,7 @@ def self.set_shipping!(order, fulfillment_type:, shipping:)
 
     order_shipping = OrderShipping.new(order)
     case fulfillment_type
-    when Order::PICKUP then order_shipping.pickup!
+    when Order::PICKUP then order_shipping.pickup!(shipping)
```
ashkan18: this reads little weird, maybe we should do:
```ruby
order_shipping.pickup!(shipping_info[:phone_number])
```
and have 
```ruby
def pickup!(buyer_phone_number)
```

ashkan18:
🙌

PR #538, opened 2019-12-19
Refactor PaymentService by making it a class and initiate with Order

ashkan18:
# Problem
`PaymentService` used to get list of parameters as input to its methods. It provides method to our payment provider (Stripe) so the caller had to generate all those needed parameters based on the action.

This gets confusing since we have to pass those parameters between few methods and list of them was growing.

# Change
Make `PaymentService` a class instead of module and have it initiated using `Order`, this way payment service would be able to generate it's own parameters based on the order. Things like setting metadata and shipping info and etc.

# Extra
Added `transfer_group` to the passed params and set it to the `Order.id`. cc: @mdole

### DIFF
--- a/lib/payment_service.rb
+++ b/lib/payment_service.rb
@@ -30,7 +30,7 @@ class PaymentService
     )
     update_transaction_with_payment_intent(new_transaction, payment_intent)
     new_transaction
-  rescue Stripe::CardError => e
+  rescue Stripe::StripeError => e
     transaction_from_payment_intent_failure(e, transaction_type: Transaction::CAPTURE)
   end
 
@@ -61,7 +61,7 @@ class PaymentService
       source_id: payment_intent.payment_method,
       payload: payment_intent.to_h
     )
-  rescue Stripe::CardError, Stripe::StripeError => e
+  rescue Stripe::StripeError => e
     transaction_from_payment_intent_failure(e)
   end
 
@@ -99,7 +99,7 @@ class PaymentService
     )
     update_transaction_with_payment_intent(new_transaction, payment_intent)
     new_transaction
-  rescue Stripe::CardError, Stripe::StripeError => e
+  rescue Stripe::StripeError => e
     transaction_from_payment_intent_failure(e, transaction_type: capture ? Transaction::CAPTURE : Transaction::HOLD)
   end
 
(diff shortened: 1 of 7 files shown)


lib/payment_service.rb
```diff
@@ -84,31 +92,30 @@ def self.create_payment_intent(credit_card:, buyer_amount:, seller_amount:, merc
...
-      on_behalf_of: merchant_account[:external_id],
+      payment_method: order.credit_card[:external_id],
+      customer: order.credit_card[:customer_account][:external_id],
+      on_behalf_of: order.merchant_account[:external_id],
       transfer_data: {
-        destination: merchant_account[:external_id],
-        amount: seller_amount
+        destination: order.merchant_account[:external_id],
+        amount: order.seller_total_cents,
+        transfer_group: order.id
```
ashkan18: cc: @mdole this should help our upcoming effort of grouping transfer together for our FSA :)

lib/payment_service.rb
```diff
@@ -27,35 +34,21 @@ def self.capture_authorized_hold(payment_intent_id)
...
+    Transaction.new(external_id: @payment_intent_id, external_type: Transaction::PAYMENT_INTENT, transaction_type: Transaction::CANCEL, status: Transaction::SUCCESS, payload: payment_intent.to_h)
   rescue Stripe::StripeError => e
-    generate_transaction_from_exception(e, Transaction::REFUND, external_id: external_id, external_type: Transaction::PAYMENT_INTENT)
+    generate_transaction_from_exception(e, Transaction::CANCEL, external_id: @payment_intent_id, external_type: Transaction::PAYMENT_INTENT)
   end
 
-  def self.confirm_payment_intent(payment_intent_id)
-    payment_intent = Stripe::PaymentIntent.retrieve(payment_intent_id)
+  def confirm_payment_intent
+    payment_intent = Stripe::PaymentIntent.retrieve(@payment_intent_id)
```
joeyAghion: Lines like this are repeated a few times. You could refactor them into a helper method called `payment_intent` and remove them. (If safe, that method could memoize the result.)

ashkan18:
🙌 merging this so `OrderCancelation` refactor @sepans is doing can be build on top of this

PR #540, opened 2019-12-19
Upgrade to rails6

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-1609

# Things Learned
- `graphql-ruby` upgrade sets the default `max_depth` of queries to 10 which failed lot of tests since our schema introspection query is 13 level deep. Had to manually increase that in order to fix the tests.
- Our pagination tests started failing because `graphql-ruby` (for a good reason) now forces queries being sent to connections to have `sort` because without it cursor and pagination is meaningless, so I ended up setting default for our `orders` query.


ashkan18:
now that other changes are in prod for few days, going to merge this.

PR #541, opened 2019-12-19
Fix sending transfer_group in wrong spot

ashkan18:
# Problem
Integrity tests started failing after #538 with errors like
https://dashboard.stripe.com/test/logs/req_yfs5btC5dBIW5A

# Cause
We added `transfer_group` as part of #538 but it was added in wrong place under `transfer_data` while it should have been added in root level based on 
https://stripe.com/docs/api/payment_intents/create#create_payment_intent-transfer_group 

# Props
To everyone involved in integrity tests

### DIFF
--- a/lib/payment_service.rb
+++ b/lib/payment_service.rb
@@ -115,9 +115,9 @@ class PaymentService
       on_behalf_of: order.merchant_account[:external_id],
       transfer_data: {
         destination: order.merchant_account[:external_id],
-        amount: order.seller_total_cents,
-        transfer_group: order.id
+        amount: order.seller_total_cents
       },
+      transfer_group: order.id,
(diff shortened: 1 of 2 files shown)


joeyAghion:
I'd love for a regression test to be added to this repo that could capture this kind of thing. Might even be possible to add an assertion to an existing test about the shape of the data submitted to Stripe.

ashkan18:
assertion was already there (thats why its failing now) but the issue is we mock stripe our own way and stripe ruby mock gem we were using wasn't keeping up with Stripe changes. So it's kind of we testing our own mock basically.

ashkan18:
this happened mainly because of me guessing Stripe's API 🤦‍♂

PR #542, opened 2019-12-19
Fetch all partner locations from gravity

sweir27:
We found that, because we're only fetching 10 partner locations from gravity, we weren't correctly calculating tax for partners that have > 10 tax nexuses.

### DIFF
--- a/lib/gravity.rb
+++ b/lib/gravity.rb
@@ -37,7 +37,8 @@ module Gravity
     url = "/partner/#{partner_id}/locations"
     params = { private: true }
     params = params.merge(address_type: ['Business', 'Sales tax nexus']) if tax_only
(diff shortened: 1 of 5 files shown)


lib/gravity.rb
```diff
@@ -77,4 +78,19 @@ def self.get_user(user_id)
     Rails.logger.warn("Could not fetch user #{user_id} from gravity: #{e.message}")
     nil
   end
+
+  def self.fetch_all(url, params)
+    items = []
+    page = 1
+    size = 10
```
ashkan18: should we increase the default limit to 20 since its still decent for server to server calls?

ashkan18:
#mergeongreen

PR #548, opened 2019-12-24
[Dont Merge][Pure Experiment] Experiment with switching to event store

ashkan18:
# Total Experiment
New decade, new experiment to see how things would look like in an event source world. 

Here only trying to look at `createOrderWithArtwork` mutation and see how things would look like if we'd switched to event source. 

- Added https://railseventstore.org/ lib
  - setup `event_store` in `application.rb`
- Defined two _commands_ (events) for `OrderPlaced` and `OrderCreated` 
- Defined two handlers to act on these posted events
- Subscribed our _handlers_ to their events. keep in mind one handler can listen on multiple events. In this case i can imagine `OrderNotifier` care about more events.
- Updated only `createOrderWithArtwork` to:
  - instead of creating the order, create a new `OrderPlaced` event (stored in event store table)
  - ☝️ will kick of a synchronous `OrderCreator.call` 
- Updated `OrderCreator` to work with _event_

# So far
## Pros
- Very clear separation of what happens when based on different events
- Stored events on each command and perfect audit log

## Cons
- Sync response in GraphQL mutation still feels little weird
- Need to decide what happens in case of errors 
- The sync calling of the handler feels little too magical


app/graphql/mutations/create_order_with_artwork.rb
```diff
@@ -8,7 +8,9 @@ class Mutations::CreateOrderWithArtwork < Mutations::BaseMutation
   field :order_or_error, Mutations::OrderOrFailureUnionType, 'A union of success/failure', null: false
 
   def resolve(artwork_id:, edition_set_id: nil, quantity: 1)
-    order = OrderService.create_with_artwork!(
+    order_id = SecureRandom.uuid
```
ashkan18: we now have to generate the order id in advance to be able to find it later, this may not work in case we are looking for existing orders.

app/graphql/mutations/create_order_with_artwork.rb
```diff
@@ -17,7 +19,9 @@ def resolve(artwork_id:, edition_set_id: nil, quantity: 1)
       quantity: quantity,
       user_agent: context[:user_agent],
       user_ip: context[:user_ip]
-    )
+    })
+    Rails.configuration.event_store.publish(event)
+    order = Order.find(order_id)
```
ashkan18: between line 23 and 24 is were all magic happens. We expect by line 24 we've already created or failed creating order.

app/controllers/api/base_api_controller.rb
```diff
@@ -3,5 +3,16 @@ class BaseApiController < ApplicationController
...
+    before_action :use_request_metadata
+
+    private
+
+    def use_request_metadata(&block)
+      Rails.configuration.event_store.with_metadata(request_metadata, &block)
+    end
+
+    def request_metadata
+      { user_id: current_user[:id] }
```
ashkan18: we set `user_id` on all events `metadata`

config/application.rb
```diff
@@ -37,5 +37,22 @@ class Application < Rails::Application
     ]
 
     config.active_job.queue_adapter = :sidekiq
+
+    config.to_prepare do
+      Rails.configuration.event_store = RailsEventStore::Client.new(
+        request_metadata: -> (env) do
+          request = ActionDispatch::Request.new(env)
+          {
+            remote_ip:  request.remote_ip,
```
ashkan18: set `ip` and `user_agent` for all event metadata

PR #551, opened 2020-01-07
Only require phone number for pickup orders

lilyfromseattle:
__Problem:__
We were requiring phone number on pickup orders as well as ship orders, so we were getting validation errors when pickup orders were placed without phone number (phone number field has not been added to reaction yet for pickup orders)

__Solution:__
I changed the validation logic so it only requires phone number for pickup orders. I also added a test to confirm this.


ashkan18:
🙌 thanks

PR #552, opened 2020-01-08
#Hackthon2020 Add myOrders Query

ashkan18:
# Problem
We want to be able to show specific user's order in their dashboard.

# Solution
Add a new `myOrders` query which can be filtered down by:
- `state`
- `sellerId`
and can be sorted same way we currently can sort orders.

It will return a `OrderConnection` for the current user's orders.

A trusted app request will fail.

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -55,6 +55,7 @@ class Types::QueryType < Types::BaseObject
 
   def my_orders(params = {})
     raise ActiveRecord::RecordNotFound unless context[:current_user][:id]
+
     sort = params.delete(:sort)
     order_clause = sort_to_order[sort] || { state_updated_at: :desc }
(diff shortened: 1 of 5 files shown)


ashkan18:
for the sake of #hackthon2020 going to #mergeongreen this but please review and let me know how things look like.

app/graphql/types/query_type.rb
```diff
@@ -45,6 +53,14 @@ def orders(params = {})
     Order.where(params).order(order_clause)
   end
 
+  def my_orders(params = {})
+    raise ActiveRecord::RecordNotFound unless context[:current_user][:id]
+
+    sort = params.delete(:sort)
```
sepans: I am guessing it doesn't need additional authorization since if `current_user` is invalid the query returns nothing. idk what the convention is in exchange for such cases? returning authorization error or just empty result?

PR #554, opened 2020-01-14
Switch to new style for order cancelation methods

ashkan18:
# Change
We've switched most of our actions to new style of services where we have `OrderService` have the business logic of what happens during in each action and when needed depending on the logic, use other helpers/processors to handle the logic.

This follows the same pattern and is the last piece left to fully move all of our actions to this style and be more consistent:
https://artsyproduct.atlassian.net/browse/PURCHASE-902

# Extra
As part of this added few new `statsd` counters, because why not?


app/admin/orders.rb
```diff
@@ -41,7 +41,7 @@
   end
 
   member_action :refund, method: :post do
-    OrderCancellationService.new(resource).refund!
+    OrderServoce.refund!(resource)
```
sweir27:
```suggestion
OrderService.refund!(resource)
```

ashkan18: clearly there is not much test around admin... thanks for 👀

app/services/order_service.rb
```diff
@@ -123,5 +123,42 @@ def self.confirm_fulfillment!(order, user_id)
...
+    order_cancelation_processor.cancel_payment if order.mode == Order::BUY
+    order_cancelation_processor.queue_undeduct_inventory_jobs if order.mode == Order::BUY
+    order_cancelation_processor.notify
+    Exchange.dogstatsd.increment 'order.reject'
+  end
+
+  def self.refund!(order)
+    order.refund!
+    order_cancelation_processor = OrderCancellationProcessor.new(order)
+    order_cancelation_processor.refund_payment
```
sweir27: If someone mistakenly tries to `refund` on a BUY order, will this error in the way we expect?

lib/order_cancellation_processor.rb
```diff
@@ -0,0 +1,31 @@
+class OrderCancellationProcessor
```
sweir27: For consistency, can we name this `OrderCancelationProcessor`? 🙏

app/services/order_service.rb
ashkan18: in the admin, we only allow `approved` and `fulfilled` orders to be refunded. Also in the model we can only go to refund state from those two states so we should be safe
https://github.com/artsy/exchange/blob/master/app/models/order.rb#L262

app/services/order_service.rb
```diff
@@ -123,5 +123,42 @@ def self.confirm_fulfillment!(order, user_id)
...
+  end
+
+  def self.refund!(order)
+    order.refund!
+    order_cancelation_processor = OrderCancelationProcessor.new(order)
+    order_cancelation_processor.refund_payment
+    order_cancelation_processor.queue_undeduct_inventory_jobs
+    order_cancelation_processor.notify
+    Exchange.dogstatsd.increment 'order.refund'
+    Exchange.dogstatsd.count('order.money_refunded', order.buyer_total_cents)
```
iskounen: Should we have separate order.money_refunded buckets for different currencies?

ashkan18: good call! i don't think anyone is looking at this at the moment but let me update to include currency in bucket name.

ashkan18: @iskounen made that change, take a look when have a chance

ashkan18:
going to merge this to have it on staging for few hours and do some testing.

PR #555, opened 2020-01-17
Add fulfilled_by_admin_id to orders

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-1695
Artsy admins currently can mark orders as fulfilled to clean up the _un-completed_ orders. While that works and tracked in our papertrail logic. We want to capture the fact that this was handled by admins in a more first class way. This way we can make sure emails are not sent in this case and etc.

# Solution
Add `fulfilled_by_admin_id` to `Order` and when we mark things as fulfilled in admin, we call `confirm_order` and pass newly added `confirmed_by_admin` flag that makes sure `fulfilled_by_admin_id` is set on the `Order`.

Also added `fulfilled_by_admin_id` to `OrderEvent` so consumers can decide what to do based on this being approved by admin or actual user.

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -75,7 +75,7 @@ ActiveAdmin.register Order do
 
   member_action :confirm_fulfillment, method: :post do
     if resource.fulfillment_type == Order::SHIP
-      OrderService.confirm_fulfillment!(resource, current_user[:id])
+      OrderService.confirm_fulfillment!(resource, current_user[:id], true)
     end
     redirect_to resource_path, notice: "Fulfillment confirmed!"
   end
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -75,7 +75,7 @@ ActiveAdmin.register Order do
 
   member_action :confirm_fulfillment, method: :post do
     if resource.fulfillment_type == Order::SHIP
-      OrderService.confirm_fulfillment!(resource, current_user[:id], true)
+      OrderService.confirm_fulfillment!(resource, current_user[:id], fulfilled_by_admin: true)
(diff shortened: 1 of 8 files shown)


app/admin/orders.rb
```diff
@@ -75,7 +75,7 @@
 
   member_action :confirm_fulfillment, method: :post do
     if resource.fulfillment_type == Order::SHIP
-      OrderService.confirm_fulfillment!(resource, current_user[:id])
+      OrderService.confirm_fulfillment!(resource, current_user[:id], true)
```
sepans: Wondering what is the best way to make this boolean to be more explicitly set by the caller?

One way I can think of is to make `confirm_fulfillment!` method private and create 2 new public functions `confirm_fulfillment_by_admin!` and `confirm_fulfillment_by_seller!` which just call the `confirm_fulfillment!` with appropriate boolean. Maybe this is too much of hassle or ruby has a better solution for that butI totally see myself calling the function with the wrong boolean value in my code 😆

ashkan18: good point! since this `confirm_fulfillment` is a class method and not an instance method making the private version may look a bit weird in Ruby, but what about making that last param a keyword param meaning we have to call it this way for passing param
```ruby
OrderService.confirm_fulfillment!(resource, current_user[:id], by_admin: true)
```

sepans: 💯 I knew ruby has something for this!

ashkan18: done!

ashkan18:
@sepans updated with your comments.

sepans:
🚀

PR #559, opened 2020-01-26
Add Danger rule about PaperTrail documentation

Thrillberg:
Hi! 👋 This is my first contribution to an Artsy project! I saw this open issue, so I hope it's still relevant and my PR is helpful. I'm totally happy to make changes to my first stab at it here. One other thing I noticed is that the README is substantially lengthened by the section about `Papertrail Audit Logging` so maybe it could be split off into another markdown file and linked to from the README? Also, I think the numbered list is not exactly working since each item is listed as `1.`. I'd be happy to open issues for either or both of these and take them on, too, if that would be helpful! (I presume they should be numbers `1` through `5`) 😄 

EDIT: Sorry, almost forgot, two tests failed for me locally:
```
rspec ./spec/controllers/api/health_controller_spec.rb:6 # Api::HealthController @GET #api/health unauthorized returns healthy: true
rspec ./spec/system/visiting_order_details_spec.rb:13 # visit order details renders order details page
```
Is this expected? Or are they worth looking into further?

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -1,6 +1,7 @@
 import {danger, warn, fail} from "danger"
 
 const addedOrModified = danger.git.modified_files.concat(danger.git.created_files)
+const added = danger.git.created_files
 const fileInGraphQLFolder = addedOrModified.find(f => f.startsWith("app/graphql"))
 if (fileInGraphQLFolder) {
   warn("If you want these changes to be reflected in Metaphysics, you will need to [update the stored schema](https://github.com/artsy/exchange#did-you-change-graphql-schema).")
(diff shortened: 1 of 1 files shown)


ashkan18:
thanks @Thrillberg ! this is great and yes totally agree with your comment about separating `Papertrail` section to its own doc and point to it.  Maybe we need to add a `docs` folder and put all new docs for more specific topics in there.

As for your tests failures, thing should work fine locally, the second failure (`rspec ./spec/system/visiting_order_details_spec.rb)` is a [capybara based test](https://github.com/artsy/exchange/pull/260) with `js: true` flag. I think you need to make sure things are setup properly locally. But the health test should pass locally!

ashkan18:
merging! ❤️ 💙 💜 💛

Thrillberg:
Great! Thanks for getting to this so quickly @ashkan18! I'll write up an issue for reorganizing the docs shortly. I'll look into my local test failures again but I suppose if CI is passing then it's no big deal.

ashkan18:
awesome! thanks! I see the new issue 🙌 

I'm still curious why the health test is failing for you. I can't repro it locally 🤔

Thrillberg:
>I'm still curious why the health test is failing for you.

@ashkan18 Figured it out! I didn't have a `redis` server running. Is that something that can be added to the `Procfile.dev`? Or if not there, perhaps it's worth a line in the README.

ashkan18:
oh good call! yes! feel free to add it there! The reason it works on Circle is we use [Hokusai](https://github.com/artsy/hokusai) for running tests which does everything in docker with redis and everything setup.

PR #562, opened 2020-01-28
Add fraud review to exchange

zephraph:
This PR does two things

1. Adds the notion of a fraud review
2. Makes some minor admin notes UI improvements

## Fraud Review

Currently our fraud review process is highly manual. Our friends over in Consumer Marketplace want to be able to track the manual reviews that happen and report stats on them. 

The `fraud_reviews` model is _very_ similar to the `admin_notes` model. @ashkan18 and I have had a lot of conversations on whether we should consolidate the concepts. At this point, I'm explicitly _not_ combining the two models and accepting that this _may_ be technical debt. The explicit reason is because fraud review and what admin notes are used for are two separate workflows utilized by different people. It's been requested that we do a major overhaul to how admin notes currently is set up and functions. 

In the future we may consolidate these models if we can do it in a way that makes sense for our business partners.

## What's Next

This is the first PR in a series that's required. Next will be a PR to fulcrum to add ingestion of this table into redshift. Following that there will be a PR to APRd to have fast access to create a fraud review for our MSO pals. 

Phase II of this work (when we can get to it) is updating the APRd slack app to  have more interactivity. The idea is that we'd like to meet our customers (MSO) where they're at and enable them to leave fraud reviews without ever leaving slack. This will require setting up auth on APRd for reviewers and adding a graphql endpoint to exchange for leaving reviews. Also rebuilding some of the slack messaging setup...

### DIFF
--- a/app/admin/fraud_reviews.rb
+++ b/app/admin/fraud_reviews.rb
@@ -1,13 +1,14 @@
-ActiveAdmin.register AdminNote do
+ActiveAdmin.register FraudReview do
   belongs_to :order
-  permit_params :order_id, :admin_id, :note_type, :description
+  permit_params :order_id, :admin_id, :considered_fraudulent, :reason, :context
 
   form do |f|
     f.inputs do
       f.input :order_id, as: :string, input_html: {readonly: true, value: order.id}
       f.input :admin_id, input_html: {readonly: true, value: current_user[:id]}, wrapper_html: { style: "display:none" }
-      f.input :note_type, :as => :select, :collection => AdminNote::TYPES.collect { |type| [type[1].humanize, type[0]] }
-      f.input :description
+      f.input :context
+      f.input :considered_fraudulent
+      f.input :reason
     end
     f.actions
   end
--- a/app/admin/fraud_reviews.rb
+++ b/app/admin/fraud_reviews.rb
@@ -6,7 +6,6 @@ ActiveAdmin.register FraudReview do
     f.inputs do
       f.input :order_id, as: :string, input_html: {readonly: true, value: order.id}
       f.input :admin_id, input_html: {readonly: true, value: current_user[:id]}, wrapper_html: { style: "display:none" }
-      f.input :context
       f.input :considered_fraudulent
       f.input :reason
     end
--- a/app/admin/fraud_reviews.rb
+++ b/app/admin/fraud_reviews.rb
@@ -5,7 +5,7 @@ ActiveAdmin.register FraudReview do
 
   form do |f|
     f.inputs do
-      f.input :order_id, as: :string, input_html: {readonly: true, value: order.id}
+      f.input :order_id, as: :string, input_html: {readonly: true, value: order.id}, wrapper_html: { style: "display:none" }
       f.input :admin_id, input_html: {readonly: true, value: current_user[:id]}, wrapper_html: { style: "display:none" }
       f.input :considered_fraudulent
       f.input :reason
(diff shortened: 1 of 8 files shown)


app/admin/orders.rb
```diff
@@ -390,11 +390,27 @@
...
         column "Note Type" do |admin_note|
           admin_note.note_type.to_s.humanize
         end
         column :description
       end
     end
+
+    panel "Fraud Review" do
+      h5 link_to("Add fraud review",  new_admin_order_fraud_review_path(order), class: :button)
+      table_for(order.fraud_reviews) do
```
ashkan18:
```suggestion
      table_for(order.fraud_reviews.order(created_at: :desc) do
```

app/admin/fraud_reviews.rb
```diff
@@ -0,0 +1,15 @@
+ActiveAdmin.register FraudReview do
```
ashkan18: would be useful to follow https://stackoverflow.com/questions/9047848/rails-activeadmin-change-the-after-update-redirect-to to redirect to order page after successful submission.

app/models/fraud_review.rb
```diff
@@ -0,0 +1,3 @@
+class FraudReview < ApplicationRecord
+  belongs_to :order
```
ashkan18: im generally not a big fan of `default_scope` but it may be useful in this case that we always get the most recent one first you could do 
```ruby
default_scope { order(created_at: :desc) }
```

db/schema.rb
```diff
@@ -23,8 +23,8 @@
     t.uuid "resource_id"
     t.string "author_type"
     t.uuid "author_id"
-    t.datetime "created_at", precision: 6, null: false
```
ashkan18: hmm... wondering where this is coming from, most likely a missed schema update from the past?

app/admin/orders.rb
```diff
@@ -390,11 +390,27 @@
       h5 link_to("Add note", new_admin_order_admin_note_path(order), class: :button)
       table_for(order.admin_notes) do
         column :created_at
+        column "Admin" do |fraud_review|
+          Gravity.get_user(fraud_review.admin_id)[:name]
```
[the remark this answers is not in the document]
zephraph: Actually @mzikherman and I talked about potentially adding memorization to this method. What do you think?

config/routes.rb
```diff
@@ -7,6 +7,7 @@
     post '/webhooks/stripe', to: 'webhooks#stripe'
   end
   resources :admin_notes
+  resources :fraud_reviews
```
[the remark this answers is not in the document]
zephraph: I have no idea. I was basically copying the pattern.

app/models/fraud_review.rb
zephraph: Hmmm, I don't understand. What would this ultimately do?

db/schema.rb
zephraph: Ummm, I'm not sure. Should I revert this part?

app/admin/fraud_reviews.rb
zephraph: I tried adding this to the end of the file:

```ruby
  controller do
    def update
      update! do |format|
        format.html { redirect_to admin_orders_path }
      end
    end
  end
```

but it didn't seem to work. I replaced `admin_orders_path` with `admin_order_path` and `orders_path`, but still nothing. 

Thoughts?

app/admin/orders.rb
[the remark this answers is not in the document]
zephraph: @ashkan18 and I discussed this in person. Given that this is only viewed by admins and there should never be more than a few in this view at any given time we're not going to optimize it further.

zephraph:
@ashkan18 when you're ready, I believe this is good to go.

ashkan18:
perfect! let's do this! 🕺

PR #563, opened 2020-01-28
Move PaperTrail documentation to separate file

Thrillberg:
Oops! I need a description! This moves some documentation away from the README, linking to it instead, in the interest of keeping the README doc itself a little shorter.


ashkan18:
Nice! sorry for late response! lot cleaner! 🙌

PR #564, opened 2020-01-28
Commission exemption: orders [GALL-2498]

mdole:
This PR adds code that calls the [new Gravity GraphQL endpoint](https://github.com/artsy/gravity/pull/12725) to not charge a partner commission on BNMO sales when appropriate.

For this to work, we add a call to Gravity during the `approve!` step of order processing. When a partner approves an order, we send a call to `debitCommissionExemption`, which will return an amount of GMV on which to not charge commission, or it will return 0 if we should not exempt any of the order from commission. We then update the `data_transfer` --> `amount` on the Stripe hold, which is the final amount of money transferred to the partner, at the same time we `capture` the hold.

**Note:** After merging this, we should make sure we QA thoroughly before deploying to production, since this impacts the final amount of money being sent to partners (and also the amount Artsy pockets). Before merging and deploying to staging, let's make sure no one is going to be held up if this is up for a few days.

Jira Ticket: [GALL-2498](https://artsyproduct.atlassian.net/browse/GALL-2498)


app/services/order_service.rb
```diff
@@ -69,16 +69,19 @@ def self.submit!(order, user_id)
     order
   rescue StandardError => e
     # catch all
+    order_processor.revert_reason = e.message
```
iskounen: - Do we also want to log the reason for a revert on line 54 when there is insufficient inventory? If so, we should consider adding an optional `reason` parameter to the `revert!` method in `OrderProcessor` so that method can handle all actions that need to happen during a revert.
- I can't find a `revert_reason` field in `OrderProcessor`. Is this supposed to update the `@reversion_reason` field?

spec/lib/order_processor_spec.rb
```diff
@@ -150,6 +151,15 @@
...
       expect(order_processor.instance_variable_get(:@state_changed)).to eq false
     end
+    it 'it reverts debit commission exemption' do
+      order.submit!
+      order.approve!
+      order_processor.instance_variable_set(:@reversion_reason, 'super ugly artwork')
+      order_processor.instance_variable_set(:@exempted_commission, true)
+
+      expect(Gravity).to receive(:credit_commission_exemption).with(order.seller_id, order.items_total_cents, order.currency_code, order.id, 'super ugly artwork')
+      order_processor.revert!
```
iskounen: It looks like we need to continue this test to ensure that the debit commission exemption is reversed. Is that correct?

lib/gravity.rb
```diff
@@ -93,4 +93,32 @@ def self.fetch_all(url, params)
...
+    mutation_args = {
+      input: {
+        partnerId: partner_id,
+        exemption: { amountMinor: amount_minor, currencyCode: currency_code },
+        referenceId: reference_id,
+        notes: notes
+      }
+    }
+    response = GravityGraphql.authenticated.debit_commission_exemption(mutation_args).to_h
+    response['data']['debitCommissionExemption']['amountOfExemptGmvOrError']
```
mdole: Is this safe? Will `GravityGraphql` return `nil` sometimes, which would cause this to break?

app/services/order_service.rb
mdole: good catch on `revert`/`reversion` reason - there was already a mismatch between those naming conventions, and I thought I did all the cleanup but clearly missed some!

also like the idea of adding that param; done in be15c4f

spec/lib/order_processor_spec.rb
mdole: hm, yeah - I think it's a question of "what level of testing is appropriate when?"

I'm of the opinion that this is the appropriate level of test for this spec, because this is all `order_processor` is responsible for. Feels to me like we should add a test to `gravity_spec` ensures we're making the correct call through Artemis. Does that sound reasonable?

app/operations/gravity_graphql/credit_commission_exemption.graphql
```diff
@@ -0,0 +1,13 @@
+mutation($input: CreditCommissionExemptionInput!) {
+  creditCommissionExemption(input: $input) {
+    totalRemainingGmvOrError {
+      ... on Money {
+        amountMinor
```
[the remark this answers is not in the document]
mdole: good point, added!

spec/lib/order_processor_spec.rb
iskounen: I naively pointed this out because I was expecting some kind of assertion after the call to `revert!`. Is the line before that somehow asynchronously checking that the creditCommissionExemption mutation is requested as a result of the `revert!`?

lib/order_processor.rb
```diff
@@ -116,4 +120,42 @@ def on_success
...
+      else
+        commission_cents = (li.list_price_cents - exemption_running_total) * order.commission_rate
+        exemption_running_total = 0
+        li.update!(commission_fee_cents: commission_cents)
+      end
+    end
+
+    # Update the toplevel commission on the order
+    order.update!(
+      commission_fee_cents: @order.line_items.map(&:commission_fee_cents).sum
```
iskounen: Does this update need to happen before the update on line 154, or can the happen at the same time?

lib/order_processor.rb
```diff
@@ -116,4 +120,42 @@ def on_success
     ReminderFollowUpJob.set(wait_until: order.state_expiration_reminder_time).perform_later(order.id, order.state)
     Exchange.dogstatsd.increment "order.#{order.state}"
   end
+
+  # Call Gravity to check if the partner should be charged commission on this order and apply it if so
+  def debit_commission_exemption(notes)
+    gmv_to_exempt_and_currency_code = Gravity.debit_commission_exemption(order.seller_id, order.items_total_cents, order.currency_code, order.id, notes)
+    apply_commission_exemption(gmv_to_exempt_and_currency_code[:amount_minor])
+  rescue Errors::InternalError
+    nil
```
iskounen: What is this exception handler protecting us from?

lib/gravity.rb
```diff
@@ -93,4 +93,32 @@ def self.fetch_all(url, params)
...
+  def self.debit_commission_exemption(partner_id, amount_minor, currency_code, reference_id, notes)
+    mutation_args = {
+      input: {
+        partnerId: partner_id,
+        exemption: { amountMinor: amount_minor, currencyCode: currency_code },
+        referenceId: reference_id,
+        notes: notes
+      }
+    }
+    response = GravityGraphql.authenticated.debit_commission_exemption(mutation_args).to_h
```
iskounen: I would like to learn more about the mechanics of making GraphQL calls. How does the `creditCommissionExemption` mutation get requested as a result of this method call?

spec/lib/order_processor_spec.rb
mdole: yeah! `.to receive` is super cool - you call it _before_ the thing you expect to happen, and then it will pass if the appropriate thing (`Gravity` in this case) receives the appropriate call (`credit_commission_exemption`).

lib/gravity.rb
mdole: guessing you were referring to `debitCommissionExemption` 😄 

this is a bit of Artemis magic - it knows to look for the `debit_commission_exemption.graphql` file adjacent to `gravity_graphql.rb`!

lib/gravity.rb
mdole: good point! I'll investigate that tomorrow

lib/gravity.rb
```diff
@@ -93,4 +93,32 @@ def self.fetch_all(url, params)
...
+    response['data']['debitCommissionExemption']['amountOfExemptGmvOrError']
+  rescue GravityGraphql::GraphQLError
+    raise Errors::InternalError.new(:gravity, message: e.message)
```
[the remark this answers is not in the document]
mdole: I don't believe so. I'll double check the business logic here, but I believe we would prefer for the hold to be captured even if the exemption doesn't go through for some reason - my thinking is that it's more important for the gallery to process the order and get the work to the collector as quickly as possible than it is for us to get the exemption right (since we can also figure out another way to compensate them, I imagine).

Def worth double checking though!

lib/order_processor.rb
mdole: I _think_ I set up `Gravity.debit_commission_exemption` to throw an `InternalError` if something goes wrong...but yeah I'm honestly not sure if this is the right way to handle that issue (or even if that would be the way things shake out). thoughts/suggestions? I can also take a closer look tomorrow

lib/order_processor.rb
mdole: Good question - I just double checked and this does need to happen before the call to `BuyOrderTotals`. It's because `seller_total_cents` is dependent upon `commission_fee_cents`, which would not be correct when we call `BuyOrderTotals` if we hadn't already updated the line items

lib/gravity.rb
```diff
@@ -93,4 +93,32 @@ def self.fetch_all(url, params)
     end
     items
   end
+
+  def self.debit_commission_exemption(partner_id, amount_minor, currency_code, reference_id, notes)
```
jpotts244: Ruby's keyword arguments might be a nice idea here! https://thoughtbot.com/blog/ruby-2-keyword-arguments

I've found these very helpful in methods that take more than a two arguments - it helps callers of this know what they are explicitly passing into the method and we dont have to solely rely on the ordering of the arguments.

It might look something like:
```ruby
def self.debit_commission_exemption(partner_id:, amount_minor:, currency_code:, reference_id:, notes:)
   // super cool code here
```

and we can call this method in other parts of the code like this:
```ruby
  Gravity .debit_commission_exemption(partner_id: "id-here", amount_minor: 0, currency_code: "USD", reference_id: "im-a-reference-id", notes: "ship it")
```

Let me know what you think! `credit_commission_exemption` is also another good candidate

mdole: oh yeah! I think that's a great idea! helps provide a little more structure and safety when calling the method, and I think it's more readable too.

thanks for the nudge - I'd run across that style of method declaration before but I didn't know what it was called or really how to think about it.

lib/gravity.rb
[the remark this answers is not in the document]
mdole: double checked this and seems like we're comfortable with letting the order go through even if the exemption fails, and we would like to figure out a way to notify on exemption failure to do a manual reimbursement. I have a pairing session with Ashkan tomorrow and will talk with him about notifying then! https://artsy.slack.com/archives/C9ZJYFDNF/p1581626381208600

lib/order_processor.rb
mdole: We talked IRL - made some tweaks to error handling in 1f75764 and are planning to follow up by having this trigger some kind of notification about commission exemption failing

lib/gravity.rb
mdole: updated to use dig in 459ec50, and we did some thinking and poking around - we don't think this would ever return `nil`. Should always either raise an error (if it can't access Gravity, for example) or return a GQL response. Waiting to hear back in #graphql in case anyone knows of something we're missing, but should be safe

app/services/order_service.rb
```diff
@@ -51,7 +51,7 @@ def self.submit!(order, user_id)
 
     order_processor.advance_state(:submit!)
     unless order_processor.deduct_inventory
-      order_processor.revert!
+      order_processor.revert!('Insufficient inventory')
```
ashkan18: instead of string lets use symbol so `:insufficient_inventory`

app/services/order_service.rb
```diff
@@ -69,16 +69,18 @@ def self.submit!(order, user_id)
...
-    order_processor&.revert!
+    order_processor&.revert!(e.message)
     raise e
   end
 
   def self.approve!(order, user_id)
     raise Errors::ValidationError.new(:unsupported_payment_method, order.payment_method) unless order.payment_method == Order::CREDIT_CARD
 
     payment_service = PaymentService.new(order)
+    order_processor = OrderProcessor.new(order, user_id)
```
ashkan18: lets move the payment part into `OrderProcessor` and use the `payment_service` already in that processor. so something like
```ruby
order_processor.debit_commission_exeption...
order_processor.capture_hold
```
and add new `capture_hold` to `OrderProcessor`

app/services/order_service.rb
```diff
@@ -69,16 +69,18 @@ def self.submit!(order, user_id)
...
     transaction = nil
     order.approve! do
+      order_processor.debit_commission_exemption('Exchange debit')
```
ashkan18: if this note is going to be the same for everyone i wouldn't pass it for now and leave it for the cases that are needed for something.

lib/order_processor.rb
```diff
@@ -13,13 +13,17 @@ def initialize(order, user_id, offer = nil)
...
   end
 
-  def revert!
+  def revert!(reversion_reason = nil)
     undeduct_inventory! if @deducted_inventory.any?
     reset_totals! if @totals_set
+    revert_debit_exemption(@reversion_reason) if @exempted_commission
     return unless @state_changed
 
+    @reversion_reason = reversion_reason
```
ashkan18: looks like we can remove this one since its already passed in the call to `revert_debit_exemption`

lib/order_processor.rb
```diff
@@ -116,4 +120,46 @@ def on_success
...
   end
+
+  # Call Gravity to check if the partner should be charged commission on this order and apply it if so
+  def debit_commission_exemption(notes)
+    gmv_to_exempt_and_currency_code = Gravity.debit_commission_exemption(partner_id: order.seller_id,
+                                                                         amount_minor: order.items_total_cents,
+                                                                         currency_code: order.currency_code,
+                                                                         reference_id: order.id,
+                                                                         notes: notes)
+    apply_commission_exemption(gmv_to_exempt_and_currency_code[:amount_minor])
```
ashkan18: our mutation response is a union that could be `Money` or `Error` we should check if `:amount_minor` exists on the map since its possible to not get this when mutation ends in `Error` , `gmv_to_exempt_and_currency_code[:amount_minor]` will be `nil`.

lib/order_processor.rb
```diff
@@ -116,4 +120,46 @@ def on_success
...
+  rescue GravityGraphql::GraphQLError
+    nil
```
ashkan18: lets log this case here for now and later we could post an event.

lib/order_processor.rb
```diff
@@ -116,4 +120,46 @@ def on_success
...
+    order.update!(
+      commission_fee_cents: @order.line_items.map(&:commission_fee_cents).sum
+    )
+    totals = order.mode == Order::BUY ? BuyOrderTotals.new(@order) : OfferOrderTotals.new(@offer)
+    order.update!(seller_total_cents: totals.seller_total_cents)
+    @exempted_commission = true
+  end
+
+  def revert_debit_exemption(reversion_reason)
+    Gravity.credit_commission_exemption(partner_id: order.seller_id, amount_minor: order.items_total_cents, currency_code: order.currency_code, reference_id: order.id, notes: reversion_reason)
```
ashkan18:
```ruby
Gravity.credit_commission_exemption(partner_id: order.seller_id, amount_minor: order.items_total_cents, currency_code: order.currency_code, reference_id: order.id, notes: reversion_reason) if @exempted_commission
```

app/services/order_service.rb
mdole: talked via slack - we actually think this wouldn't do much for consistency since this is the only case we're doing a credit with a reason for right now. if we start doing a lot more credits for different reasons, then we should revisit and consider an enum or something similar. for now, changed to `insufficient_inventory`

lib/order_processor.rb
mdole: looks like the scenario in which we return an error is when we try to debit with a negative amount - will do some testing and make sure we're covered from the Exchange side if that happens somehow

mdole:
going to merge this and follow up with @ashkan18's last couple items in a separate PR so this can be tested by Cypress over the weekend!

lib/order_processor.rb
mdole: addressed in https://github.com/artsy/exchange/pull/575

PR #566, opened 2020-01-31
Fix Stripe capability issue and regression from order cancelation

ashkan18:
# Problem(s)
We've started seeing failures on staging for our cypress integrity tests where we get following error on Stripe:
```
Your destination account needs to have at least one of the following capabilities enabled: transfers, legacy_payments
```
This is a new set of errors that we haven't seen in the past and looks like for these since there is no charge we can't store transactions. This hasn't happened on prod yet but worth adding support for it:

https://sentry.io/organizations/artsynet/issues/1485964863/?project=1275269&query=is%3Aunresolved

# Regression
As part of #554 we seem to have missed 2 admin calls in admin which lead to 
https://sentry.io/organizations/artsynet/issues/1477952941/?project=1275271&query=is%3Aunresolved
and 
https://sentry.io/organizations/artsynet/issues/1477975587/?project=1275271&query=is%3Aunresolved

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -41,17 +41,17 @@ ActiveAdmin.register Order do
   end
 
   member_action :refund, method: :post do
-OrderService.refund!(resource)
+    OrderService.refund!(resource)
     redirect_to resource_path, notice: "Refunded!"
   end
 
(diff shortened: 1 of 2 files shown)


sepans:
🙌 and 😌

PR #570, opened 2020-02-04
Use Simplecov and Codecov for coverage reporting

eessex:
- Replaces `coveralls` with `simplecov` gem, and updates reporting configuration
- Uploads reports after CI test runs via `codecov` orb

Coverage reports for this app available here: https://codecov.io/gh/artsy/exchange

Related to https://artsyproduct.atlassian.net/browse/PLATFORM-2072

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,7 @@
 version: 2.1
 
 orbs:
+  codecov: codecov/codecov@1.0.5
   hokusai: artsy/hokusai@0.7.2
 
 not_staging_or_release: &not_staging_or_release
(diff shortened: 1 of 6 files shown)


ashkan18:
oh nice! this def looks simpler :) thanks! we also had this issue in the past with codecov where it wasn't uploading coverage from `master` when running in circle in docker hopefully this fixes it.

eessex:
This matches the setup we've used for CI on other apps, so I'd expect to see coverage for master once the staging deploy finishes running.

eessex:
Looks like master is there now - https://codecov.io/gh/artsy/exchange/branches

ashkan18:
niiiice!

PR #572, opened 2020-02-14
fix failing CI build

mdole:
See https://github.com/artsy/rosalind/pull/349 and https://artsy.slack.com/archives/CP9P4KR35/p1581717280251400


ashkan18:
Nice! merging

PR #575, opened 2020-02-20
convert debit return to snake case and add tests

mdole:
Fixing a couple issues that popped up while QAing commission exemption. First, our `dig` statement was always returning `nil` because we were using strings instead of symbols, and second, the object we were trying to return would have had camelCased keys (in `order_processor.rb` we [expect snake_case](https://github.com/artsy/exchange/blob/a22bddbb521c81cf9b3d2d4af42046acb6a1fd53/lib/order_processor.rb#L129)).

### DIFF
--- a/lib/gravity.rb
+++ b/lib/gravity.rb
@@ -104,7 +104,9 @@ module Gravity
       }
     }
     response = GravityGraphql.authenticated.debit_commission_exemption(mutation_args).to_h
-    response.dig('data', 'debitCommissionExemption', 'amountOfExemptGmvOrError')
+    gmv_or_error = response.dig(:data, :debitCommissionExemption, :amountOfExemptGmvOrError)
+    # Convert hash to snake case
+    gmv_or_error.transform_keys { |key| key.to_s.underscore.to_sym }
   end
 
   def self.credit_commission_exemption(partner_id:, amount_minor:, currency_code:, reference_id:, notes:)
--- a/lib/gravity.rb
+++ b/lib/gravity.rb
@@ -104,9 +104,13 @@ module Gravity
       }
     }
     response = GravityGraphql.authenticated.debit_commission_exemption(mutation_args).to_h
-    gmv_or_error = response.dig(:data, :debitCommissionExemption, :amountOfExemptGmvOrError)
-    # Convert hash to snake case
-    gmv_or_error.transform_keys { |key| key.to_s.underscore.to_sym }
+    gmv_or_error = response.dig('data', 'debitCommissionExemption', 'amountOfExemptGmvOrError')
+    if gmv_or_error.nil?
+      nil
+    else
+      # Convert hash to snake case
+      gmv_or_error.transform_keys { |key| key.to_s.underscore.to_sym }
+    end
   end
 
   def self.credit_commission_exemption(partner_id:, amount_minor:, currency_code:, reference_id:, notes:)
--- a/spec/lib/gravity_spec.rb
+++ b/spec/lib/gravity_spec.rb
@@ -214,7 +214,11 @@ describe Gravity, type: :services do
 
   describe '#debit_commission_exemption' do
     let(:parameters) { { partner_id: seller_id, amount_minor: 100, currency_code: 'USD', reference_id: 'order123', notes: 'hi' } }
-    # Full response looks like: #<GraphQL::Client::Response:0x00007fbad5a3f650 @original_hash={"data"=>{"debitCommissionExemption"=>{"amountOfExemptGmvOrError"=>{"__typename"=>"Money", "amountMinor"=>490, "currencyCode"=>"USD"}}}}, @data=#< debitCommissionExemption=...>, @errors=#<GraphQL::Client::Errors @messages={} @details={}>, @extensions=nil> }
+    # Full response looks like:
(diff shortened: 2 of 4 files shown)


spec/lib/gravity_spec.rb
```diff
@@ -211,4 +211,22 @@
       expect(response.length).to eq 0
     end
   end
+
+  describe '#debit_commission_exemption' do
```
iskounen: I think that a test for this method should verify that it returns whatever value is assigned to `amountOfExemptGmvOrError` in the mutation's response body, or `nil` if the response body doesn't have a `amountOfExemptGmvOrError`.

What I'm thinking is something like:
```
it 'returns the value of amountOfExemptGmvOrError if it is included in the response body' do
  debit_commission_exemption_mutation.to_return(status: 200, body: "{ 'data': { 'debitCommissionExemption': { 'amountOfExemptGmvOrError': 'foo' } } }", headers: {})
  response = Gravity.debit_commission_exemption(partner_id: order.seller_id, amount_minor: order.items_total_cents, currency_code: order.currency_code, reference_id: order.id, notes: 'test')
  expect(response).to eq 'foo'
end
it 'returns nil if amountOfExemptGmvOrError is not included in the response body' do
  debit_commission_exemption_mutation.to_return(status: 200, body: '{}', headers: {})
  response = Gravity.debit_commission_exemption(partner_id: order.seller_id, amount_minor: order.items_total_cents, currency_code: order.currency_code, reference_id: order.id, notes: 'test')
  expect(response).to be_nil
end
```

I see value in adding a test to make sure that the keys in the response body are transformed to snake_case, but I think it is misleading to expect the return value to be `{ amount_minor: 10, currency_code: 'USD' }` on a successful mutation.

jpotts244: +1, adding tests in this format will also expose the bug you found with the symbol vs string look ups on line 107 since it will be testing the method receiving true JSON!

mdole: good suggestion, thanks both! I did a little poking around in a rails console, and I actually think I was wrong about having `dig` use symbols instead of strings. An excerpt (placed a `byebug` in `Gravity.debit_commission_exemption`):
```
dotenv bundle exec rails c
response2 = Gravity.debit_commission_exemption(partner_id: '581b45e4cd530e658b000124', amount_minor: 500, currency_code: 'USD', reference_id: '123orderid1', notes: 'tktk')
>> #<GraphQL::Client::Response:0x00007fbad5a3f650 @original_hash={"data"=>{"debitCommissionExemption"=>{"amountOfExemptGmvOrError"=>{"__typename"=>"Money", "amountMinor"=>490, "currencyCode"=>"USD"}}}}, @data=#< debitCommissionExemption=...>, @errors=#<GraphQL::Client::Errors @messages={} @details={}>, @extensions=nil>
response2.to_h
>> {"data"=>{"debitCommissionExemption"=>{"amountOfExemptGmvOrError"=>{"__typename"=>"Money", "amountMinor"=>490, "currencyCode"=>"USD"}}}}
response2.to_h.dig(:data, :debitCommissionExemption, :amountOfExemptGmvOrError)
>> nil
response2.to_h.dig('data','debitCommissionExemption', 'amountOfExemptGmvOrError')
>> {"__typename"=>"Money", "amountMinor"=>490, "currencyCode"=>"USD"}
```

So now I'm trying to figure out how to have `GravityGraphql.debit_commission_exemption` return that whole `#<GraphQL::Client::Response...` thing, since it looks to be what's actually coming back from Gravity. Does that feel right?

iskounen: What will you be able to do once you have access to the `GraphQL::Client::Response` object?

mdole: yeah, I think you're totally right - I was just feeling nervous about mocking it wrong tbh. I ended up just using the response hash. I did leave a comment in there with the full response...probably unnecessary, just felt nice to have a record of what that response looks like

mdole: (comment removed! marking as resolved)

lib/gravity.rb
```diff
@@ -104,7 +104,13 @@ def self.debit_commission_exemption(partner_id:, amount_minor:, currency_code:,
       }
     }
     response = GravityGraphql.authenticated.debit_commission_exemption(mutation_args).to_h
-    response.dig('data', 'debitCommissionExemption', 'amountOfExemptGmvOrError')
+    gmv_or_error = response.dig('data', 'debitCommissionExemption', 'amountOfExemptGmvOrError')
+    if gmv_or_error.nil?
+      nil
+    else
+      # Convert hash to snake case
+      gmv_or_error.transform_keys { |key| key.to_s.underscore.to_sym }
```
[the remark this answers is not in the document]
mdole: yeah! I'd like to make that a follow-up to keep this moving forwards, but I'm def up for doing that refactor

spec/lib/order_processor_spec.rb
```diff
@@ -466,5 +466,21 @@
         expect(order.commission_fee_cents).to eq 800_00
       end
     end
+
+    context 'with successful query but return other than amount_minor from Gravity.debit_commission_exemption' do
```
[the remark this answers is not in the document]
mdole: rearranged in f6d9779!

PR #571, opened 2020-02-05
GALL-2588 Commission exemption credit

ansor4:
Requires https://github.com/artsy/exchange/pull/564

This PR calls gravity to credit commission exemption in cases where we refunded orders or orders were reverted.

I use `stub_request` because I couldn't get the Artemis dsl `stub_graphql` to work.

PR #576, opened 2020-02-20
Credit a Partner's Commission Exemption When an Order Is Refunded (GALL-2588)

iskounen:
This PR updates Exchange to request a credit to a partner's commission exemption balance in Gravity when an order is refunded in the admin UI.

This work was started in #571. Instead of rebasing that feature branch on master after #564 was merged, I decided to open a new feature branch and manually port over the relevant changes to the refund flow, since there were too many merge conflicts.

### DIFF
--- a/lib/order_cancelation_processor.rb
+++ b/lib/order_cancelation_processor.rb
@@ -19,6 +19,9 @@ class OrderCancelationProcessor
     transaction = @payment_service.refund
     @order.transactions << transaction
     raise Errors::ProcessingError.new(:refund_failed, transaction.failure_data) if transaction.failed?
+
+    # Only credit commission exemption for refunds, cancelations never deducted from commission exemption total
+    Gravity.credit_commission_exemption(partner_id: @order.seller_id, amount_minor: @order.items_total_cents, currency_code: @order.currency_code, reference_id: @order.id, notes: 'refund')
   end
 
   def cancel_payment
--- a/lib/order_processor.rb
+++ b/lib/order_processor.rb
@@ -144,6 +144,7 @@ class OrderProcessor
   end
 
   def revert_debit_exemption(reversion_reason)
-    Gravity.credit_commission_exemption(partner_id: order.seller_id, amount_minor: order.items_total_cents, currency_code: order.currency_code, reference_id: order.id, notes: reversion_reason) if @exempted_commission
+    Gravity.credit_commission_exemption(partner_id: order.seller_id, amount_minor: order.items_total_cents, currency_code: order.currency_code, reference_id: order.id, notes: reversion_reason)
+    @exempted_commission = false
   end
 end
--- a/spec/lib/gravity_spec.rb
+++ b/spec/lib/gravity_spec.rb
@@ -227,4 +227,13 @@ describe Gravity, type: :services do
       expect(return_value).to be nil
     end
   end
+
+  describe '#credit_commission_exemption' do
+    it 'requests the credit commission exemption mutation and returns nil' do
+      mutation = stub_request(:post, Rails.application.config_for(:graphql)[:gravity_graphql][:url]).to_return(status: 200, body: { foo: { bar: 'baz' } }.to_json)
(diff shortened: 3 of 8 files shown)


spec/lib/gravity_spec.rb
```diff
@@ -227,77 +227,20 @@
...
-  #     expect(res).to eq({"totalRemainingGmvOrError"=>{"__typename"=>"Money", "amountMinor"=>1000, "currencyCode" => "USD"}})
-  #   end
-  # end
+  describe '#credit_commission_exemption' do
+    let(:parameters) { { partner_id: seller_id, amount_minor: 100, currency_code: 'USD', reference_id: 'order123', notes: 'hi' } }
+    let(:response) { { 'data' => { 'creditCommissionExemption' => 'foo' } } }
+    it 'returns snake-cased creditCommissionExemption if it is included in the response body' do
+      allow(GravityGraphql).to receive_message_chain(:authenticated, :credit_commission_exemption).and_return(response)
+      return_value = Gravity.credit_commission_exemption(parameters)
+      expect(return_value).to eq 'foo'
```
iskounen: @mdole I added this test so that `debit_commission_exemption` has a companion. However, I noticed that the return value for `credit_commission_exemption` doesn't seem to be used by the methods that call it. Do you know why we are returning the result of `response.dig('data', 'creditCommissionExemption')` for that method?

mdole: hm, good catch - no, I don't believe that return value serves any purpose at the moment. maybe we just return `nil` or `true`?

mdole: not sure if we'd have anything worth testing at this level in that case. we could test whether `GravityGraphql.authenticated.credit_commission_exemption` is being called with the appropriate params, but that doesn't feel like a very meaningful test

lib/order_processor.rb
```diff
@@ -161,5 +161,6 @@ def apply_commission_exemption(exemption_amount_cents)
 
   def revert_debit_exemption(reversion_reason)
     Gravity.credit_commission_exemption(partner_id: order.seller_id, amount_minor: order.items_total_cents, currency_code: order.currency_code, reference_id: order.id, notes: reversion_reason) if @exempted_commission
```
mdole: #minor thoughts on getting rid of the `if @exempted_commission` in this line? we already check that before calling this method during `revert!`, so doesn't feel warranted here

mdole:
So I think this is looking good, and also I'm trying to figure out how we could improve test coverage. It never feels good to go "well I guess we'll just test it on staging", but struggling to figure out how else we could do it

lib/order_cancelation_processor.rb
```diff
@@ -19,6 +19,9 @@ def refund_payment
     transaction = @payment_service.refund
     @order.transactions << transaction
     raise Errors::ProcessingError.new(:refund_failed, transaction.failure_data) if transaction.failed?
+
+    # Only credit commission exemption for refunds, cancelations never deducted from commission exemption total
```
iskounen: @ansor4 this comment is a carry-over from #571, and I would like to double-check that it is valid. Do you know why we don't want to credit commission exemption for canceled payments?

spec/lib/gravity_spec.rb
iskounen: I updated the test to check that the mutation was requested and that this method returns `nil`. As you say, it's not a very meaningful test, but at least it describes what this method does, which may be useful for future engineers who wish to refactor this in the future.

mdole: makes sense! smart thinking

lib/order_cancelation_processor.rb
mdole: :ceiling-cat: I can answer this one - we run `debit_commission_exemption` at order _approval_ time. orders are `canceled` if the partner rejects them or if they don't approve them within the 72hr window; in that case, we never hit Gravity to record an exemption so we don't have to worry about reverting one

iskounen: Got it. I got my BNMO terminology mixed up. Thanks for the clarification!

PR #578, opened 2020-02-21
Commission exemption: offers [GALL-2589]

mdole:
This PR continues the work started in https://github.com/artsy/exchange/pull/564, adding code that checks for exemptions when an offer is approved. 

With help from @ashkan18, I also moved the logic for calculating exemption to the `[Buy/Offer]OrderTotals` classes. We realized that it made much more sense for the logic determining the final commission after exemption to live with the logic that was responsible for setting commission amounts in general. Plus, this keeps the `order_processor` nice n' streamlined.

This PR also adds simple integration tests for both order and offer - just working on covering our bases here.

[Jira ticket](https://artsyproduct.atlassian.net/browse/GALL-2589)

### DIFF
--- a/spec/lib/order_processor_spec.rb
+++ b/spec/lib/order_processor_spec.rb
@@ -28,7 +28,6 @@ describe OrderProcessor, type: :services do
   let(:stub_artwork_request) { stub_request(:get, "#{Rails.application.config_for(:gravity)['api_v1_root']}/artwork/a-1").to_return(status: 200, body: artwork.to_json) }
   let!(:line_item1) { Fabricate(:line_item, order: order, quantity: 1, list_price_cents: 1000_00, artwork_id: 'a-1', artwork_version_id: '1', sales_tax_cents: 0, shipping_total_cents: 0) }
   let(:line_item2) { Fabricate(:line_item, order: order, artwork_id: 'a2', quantity: 2) }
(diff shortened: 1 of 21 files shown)


spec/lib/order_processor_spec.rb
```diff
@@ -380,65 +379,37 @@
...
-        order.reload
-        order_processor.apply_commission_exemption(2500_00)
-        expect(order_processor.instance_variable_get(:@exempted_commission)).to be true
-        expect(order.commission_fee_cents).to eq 400_00
-        expect(order.seller_total_cents).to eq 2512_70
+    context 'with > 0 commission exemption' do
+      OrderTotal = Struct.new(:seller_total_cents, :commission_fee_cents)
+      before do
+        allow(BuyOrderTotals).to receive(:new).and_return(OrderTotal.new(5, 10))
+        allow(OfferOrderTotals).to receive(:new).and_return(OrderTotal.new(20, 30))
```
mdole: this felt like kind of an icky way to handle this (defining a whole `struct`), but I wasn't sure how else to get an object/hash/whatever that I could run `totals.seller_total_cents` on ¯\_(ツ)_/¯

mdole:
going to go ahead and merge this w/o review so I can do some tests & QA on staging - would still very much appreciate PR feedback next week when folks are working again! I'm happy to make revisions and adjust, just want to see if it's working and get as close as possible to ready for launch by Monday :)

PR #580, opened 2020-02-25
Support Orders for Artworks Listed in Euros (GALL-2642)

iskounen:
This PR adds the Euro as a supported currency for orders.


mdole:
We basically copy/pasted Volt's chromedriver config so that we could run the browser in non-headless mode locally, but when those changes caused CI to fail, we decided to put them in a separate PR for posterity and close for the moment. If we run into issues with drivers again, we can reopen and invest some time in a fix (or if we get curious; could also be a great learning experience) https://github.com/artsy/exchange/pull/588

PR #581, opened 2020-02-27
GALL-2637 Add eu country codes and check for that when calculating shipping

ansor4:
I think this is the easiest way to add this check to this app, and it doesn't allow for bad inputs from the mutation (ie. Sending a shipping address with country 'US' and also a flag `is_eu_shipping destination`)

- [x] Add tests

### DIFF
--- a/lib/address.rb
+++ b/lib/address.rb
@@ -1,6 +1,11 @@
 class Address
   attr_reader :country, :region, :city, :street_line1, :street_line2, :postal_code
   UNITED_STATES = Carmen::Country.coded('US')
+  COUNTRY_CODES_EU_SHIPPING = %w[AT BE CH DE DK EE ES FI FR GR IE IT LV LT LU NL PL PT SK SI SE].freeze
+  # Not sure why we need this Carmen gem, but to keep with the pattern,
+  # let's run the list of country codes through that and convert to set
+  COUNTRY_CODES_EU_SHIPPING_SET = COUNTRY_CODES_EU_SHIPPING.map { |country| Carmen::Country.coded(country)&.code }.compact.to_set
+
   def initialize(address)
     @address = parse(address)
     @country = @address[:country]
@@ -29,6 +34,10 @@ class Address
     united_states? &&
       @region != UNITED_STATES.subregions.coded('HI').code &&
       @region != UNITED_STATES.subregions.coded('AK').code
+  end
+
+  def eu_local_shipping?
+    COUNTRY_CODES_EU_SHIPPING_SET.include? @country
   end
 
   private
--- a/lib/address.rb
+++ b/lib/address.rb
@@ -1,10 +1,13 @@
 class Address
   attr_reader :country, :region, :city, :street_line1, :street_line2, :postal_code
   UNITED_STATES = Carmen::Country.coded('US')
-  COUNTRY_CODES_EU_SHIPPING = %w[AT BE CH DE DK EE ES FI FR GR IE IT LV LT LU NL PL PT SK SI SE].freeze
-  # Not sure why we need this Carmen gem, but to keep with the pattern,
-  # let's run the list of country codes through that and convert to set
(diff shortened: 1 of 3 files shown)


lib/address.rb
```diff
@@ -1,6 +1,11 @@
 class Address
   attr_reader :country, :region, :city, :street_line1, :street_line2, :postal_code
   UNITED_STATES = Carmen::Country.coded('US')
+  COUNTRY_CODES_EU_SHIPPING = %w[AT BE CH DE DK EE ES FI FR GR IE IT LV LT LU NL PL PT SK SI SE].freeze
+  # Not sure why we need this Carmen gem, but to keep with the pattern,
+  # let's run the list of country codes through that and convert to set
+  COUNTRY_CODES_EU_SHIPPING_SET = COUNTRY_CODES_EU_SHIPPING.map { |country| Carmen::Country.coded(country)&.code }.compact.to_set
```
oxaudo: Hm... Wondering what is up with that? Would not that be a problem if this `carmen` thing can not convert one of the country codes to a valid code? I almost feel like we need to verify the `COUNTRY_CODES_EU_SHIPPING` and `COUNTRY_CODES_EU_SHIPPING_SET` are the same length and warn if they are not...

lib/shipping_helper.rb
```diff
@@ -3,7 +3,7 @@ def self.calculate(artwork, fulfillment_type, shipping_address = nil)
     return 0 if fulfillment_type == Order::PICKUP
     raise Errors::ValidationError.new(:missing_artwork_location, artwork_id: artwork[:_id]) if artwork[:location].blank?
 
-    if domestic?(artwork, shipping_address)
+    if domestic?(artwork, shipping_address) || eu_local_shipping?(artwork, shipping_address)
```
oxaudo: Nice!

oxaudo:
Nice and simple! Except for the triple time array duplication here:) But 🤷‍♀ !

Also wondering if we need to actually understand what this `Carmen` thing does or worry about the fact that some of the countries might not be converted?

oxaudo:
Also let me cc @ashkan18 and @sweir27 for a second opinion here if the changes make sense to them. Thanks guys! I just not sure I have  enough of an exchange understanding to be able to anticipate problems.

ansor4:
@mdole please take a look if you have a chance 🙏

oxaudo:
This PR should probably be merged roughly the same time as https://github.com/artsy/metaphysics/pull/2216 and https://github.com/artsy/reaction/pull/3222 since all of those needed to basically consider local shipping for EU to include all EU country. Not sure what is the best way to deploy all that in order so that we do not misrepresent what is going to happen to the users. Volt also need to be updated to display what 'domestic' means for Euro countries. Thoughts @ansor4 ?

lib/address.rb
```diff
@@ -1,6 +1,11 @@
 class Address
   attr_reader :country, :region, :city, :street_line1, :street_line2, :postal_code
   UNITED_STATES = Carmen::Country.coded('US')
+  COUNTRY_CODES_EU_SHIPPING = %w[AT BE CH DE DK EE ES FI FR GR IE IT LV LT LU NL PL PT SK SI SE].freeze
+  # Not sure why we need this Carmen gem, but to keep with the pattern,
```
sweir27: A little bit of sleuthing reveals: https://github.com/artsy/exchange/pull/141

We added the `carmen` gem to coerce text-based regions into proper country codes (and so that we are comparing apples to apples across Exchange).

Running these strings through the gem will ensure that they match elsewhere where we may compare them. I might get rid of the `COUNTRY_CODES_EU_SHIPPING` constant to make it more concise (or update the comment to explain the logic vs. the confusion). Or, decide that it's no longer needed!

sweir27:
In terms of shipping-- unless partners are able to set EU artworks for BNMO (I don't _think_ they can yet?) it shouldn't matter too much, right?

oxaudo:
@sweir27 - surprisingly since this is more about where the artwork ships from - we already have bunch of BNMO artworks in production that are set to ship from EU countries. For now we apply domestic shipping only to shipping within the same country and international to all other countries and this PR will change that. We are discussing the deploy plan now with @joanna20

PR #582, opened 2020-02-28
Switch from using credit_... to refund_... endpoint

mdole:
Following up on https://github.com/artsy/gravity/pull/12796, this PR switches from consuming the now-nonexistent `credit_commision_exemption` Gravity endpoint to `refund_commission_exemption`. Also grabs the new schema from Gravity using `dotenv bundle exec rake graphql:schema:update`, a handy rake task provided by Artemis


mdole:
thanks for reviewing @ansor4 ! I'm going to self-assign and merge so I can do some QA on staging later today - feels relatively safe since this is just updating to match already-merged Grav behavior

PR #595, opened 2020-03-19
[Merge close to March 26] - adjust stripe fee to the new value

oxaudo:
The exchange side of https://artsyproduct.atlassian.net/browse/GALL-2664

@iskounen @ansor4 @mdole - I left the Money part as part of the functions because one of the benefits mentioned: `Represents monetary values as integers, in cents. This avoids floating point rounding errors.`. Seemed important.

The other part of the issue is to update Volt with proper message.


spec/integration/buy_now_sca_path.rb
```diff
@@ -141,7 +141,7 @@
         tax_total_cents: 100_00,
         commission_fee_cents: 100_00,
         transaction_fee_cents: 38_00,
```
mdole: I'm surprised this line isn't failing - shouldn't our transaction fee now be `51_00`  (1300 order total including shipping etc. x .039 + .30 = 51.00)?

oxaudo: Good question. It has to be updated I've just missed it. But really not sure why it was not failing...

PR #599, opened 2020-04-17
Upgrade rubocop and switch gem

jonallured:
Exchange had fallen behind on RuboCop so this PR updates it to latest and complies with the new rules. I also switched to the `rubocop-rails` gem which is the current best-practice.

So mostly this is just churn, but I did have to refactor the `ShippingHelper.calculate` method so that I could comply but also so I could better understand what that method did - hopefully people find this easier to change in the future!

### DIFF
--- a/app/events/order_event.rb
+++ b/app/events/order_event.rb
@@ -51,7 +51,7 @@ class OrderEvent < Events::BaseEvent
   end
 
   def properties
-    PROPERTIES_ATTRS.map { |att| [att, @object.send(att)] }.to_h.merge(line_items: line_items_details, last_offer: last_offer)
+    PROPERTIES_ATTRS.index_with { |att| @object.send(att) }.merge(line_items: line_items_details, last_offer: last_offer)
   end
 
   private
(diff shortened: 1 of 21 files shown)


app/events/order_event.rb
```diff
@@ -51,7 +51,7 @@ def subject
   end
 
   def properties
-    PROPERTIES_ATTRS.map { |att| [att, @object.send(att)] }.to_h.merge(line_items: line_items_details, last_offer: last_offer)
+    PROPERTIES_ATTRS.index_with { |att| @object.send(att) }.merge(line_items: line_items_details, last_offer: last_offer)
```
ansor4, line 54: this is cool!

lib/shipping_helper.rb
```diff
@@ -1,13 +1,19 @@
...
+    elsif domestic?(artwork, shipping_address) || eu_local_shipping?(artwork, shipping_address)
+      exception = Errors::ValidationError.new(:missing_domestic_shipping_fee)
+      cents = artwork[:domestic_shipping_fee_cents]
     else
-      artwork[:international_shipping_fee_cents] || raise(Errors::ValidationError.new(:unsupported_shipping_location, failure_code: :domestic_shipping_only))
+      exception = Errors::ValidationError.new(:unsupported_shipping_location, failure_code: :domestic_shipping_only)
+      cents = artwork[:international_shipping_fee_cents]
     end
+
+    cents || raise(exception)
```
ansor4, line 16: 👍 nice!

PR #602, opened 2020-04-20
Switch to webdrivers gem

jonallured:
The current best-practice is to use the webdrivers gem rather than the older selenium-webdriver gem so this PR moves us to the new one. Let's see what circle thinks of it!


jonallured:
Hrm, this didn't work the way I expected so I'm just going to close for now and we can revisit if/when it's worth it.

PR #606, opened 2020-04-21
CSGN-121: Add more artwork details to order page

jonallured:
This PR makes some changes to the order page in the admin interface so that we can see if an order came through as the result of a consignment. Here's some before/after screenshots:

<details>
Screen Shot 2020-04-21 at 1 41 31 PM
Screen Shot 2020-04-21 at 1 41 35 PM
Screen Shot 2020-04-21 at 1 41 40 PM
Screen Shot 2020-04-21 at 1 41 42 PM

</details>

Note, there was a decent amount of churn in this active admin config file that had nothing to do with my changes. I have broken this out in the individual commits, so might be easier to step through so that the more relevant stuff stands out.

## Bonus scripts

In order to boot up Exchange locally with some data, I wanted to be able to use some of the scripts that I have written over on Convection, so this PR also includes a bunch of those: 

* bin/console
* bin/pull_data
* bin/server
* bin/worker

Probably the more interesting one is that pull data script - this will find the staging db and pull it down locally. 😎 

https://artsyproduct.atlassian.net/browse/CSGN-121

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -5,9 +5,9 @@ ActiveAdmin.register Order do
 
   scope :all
   scope('Submitted', default: true) { |scope| scope.where(state: Order::SUBMITTED) }
-  scope('Active', default: true) { |scope| scope.active }
-  scope('Pickup') { |scope| scope.where(state: [ Order::APPROVED, Order::FULFILLED, Order::SUBMITTED ], fulfillment_type: Order::PICKUP ) }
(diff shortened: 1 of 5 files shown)


jonallured:
@damassi good question and thanks for keeping me honest! 😝 

So I just looked and the only spec that looks related is this one:

https://github.com/artsy/exchange/blob/master/spec/system/visiting_order_details_spec.rb

But this is more of a sanity-check type of test, it doesn't assert about more than just the surface-level of the order detail page in admin. I guess that means we don't really cover the admin interface in spec coverage. 🤷

PR #607, opened 2020-04-23
CSGN-120: Bubble up consignment flag to stripe

jonallured:
This PR adds a new key to the metadata sent to Stripe for an order. This new field is called `any_consignment_works` and is set to true if any of the order's line items link to an artwork that has an `import_source` of `convection`.

https://artsyproduct.atlassian.net/browse/CSGN-120

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,7 +15,7 @@ Metrics/ParameterLists:
   Max: 11
 
 Metrics/ClassLength:
-  Max: 190
+  Max: 1000000
 
 Metrics/AbcSize:
   Max: 61
--- a/lib/payment_service.rb
+++ b/lib/payment_service.rb
@@ -138,6 +138,8 @@ class PaymentService
   end
 
   def metadata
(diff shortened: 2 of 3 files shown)


.rubocop.yml
```diff
@@ -15,7 +15,7 @@ Metrics/ParameterLists:
   Max: 11
 
 Metrics/ClassLength:
-  Max: 190
+  Max: 1000000
```
pepopowitz, line 18:
```suggestion
  Max: 1000001
```

:trollface:

starsirius: lol `z-index: 99999;` Maybe we just disable this cop.

damassi: @pepopowitz - out of curiosity is the thinking here that by starting at 1 one can see that there was a more deliberate attempt to increment, so if theres a problem one could then go to 2, etc, versus some other new arbitrary number?

pepopowitz: No, I was completely trolling. I agree with @starsirius that if we need to set this number to something totally unreasonable, we should just get rid of this rule.

damassi: oh hahaha 😄

PR #609, opened 2020-04-23
Add rake task to clean cypress card data

eessex:
Adds a rake task to run the exchange cleanup for cypress+test@example.com credit cards, as described here: https://github.com/artsy/integrity#too-many-credit-cards-for-a-customer

Related to https://artsyproduct.atlassian.net/browse/INTGRTY-14

Related to https://github.com/artsy/gravity/pull/12985

### DIFF
--- a/lib/tasks/clean_cypress_data.rake
+++ b/lib/tasks/clean_cypress_data.rake
@@ -1,0 +1,18 @@
+# Delete credit cards for Cypress test user
+# Cards are automatically added by Integrity builds,
+# but must be periodically cleaned.
+# Prevents Stripe errors for "too many cards"
+# https://github.com/artsy/integrity#too-many-credit-cards-for-a-customer
+
+namespace :cypress do
+  desc 'Delete credit cards for Cypress test user.'
+  task clean_credit_cards: :environment do
+    logger.info("[#{Time.zone.now}] Cleaning credit cards for Integrity test user...")
+    first_order = Order.where(buyer_id: '5d97bb95d845aa001137da26').first
+    payment_intent = Stripe::PaymentIntent.retrieve(first_order.external_charge_id)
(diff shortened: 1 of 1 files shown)


lib/tasks/clean_cypress_data.rake
```diff
@@ -0,0 +1,18 @@
...
+
+namespace :cypress do
+  desc 'Delete credit cards for Cypress test user.'
+  task clean_credit_cards: :environment do
+    logger.info("[#{Time.zone.now}] Cleaning credit cards for Integrity test user...")
+    first_order = Order.where(buyer_id: '5d97bb95d845aa001137da26').first
+    payment_intent = Stripe::PaymentIntent.retrieve(first_order.external_charge_id)
+    customer = Stripe::Customer.retrieve(payment_intent.customer)
+    sources = Stripe::Customer.list_sources(customer.id, { limit: 100 })
+    sources.map(&:id).each { |id| Stripe::Customer.delete_source(customer.id, id) }
```
joeyAghion, line 15: It doesn't look like any Exchange data is being cleaned up in this task. If that's the case, why not just delete the cards from [the Gravity task](https://github.com/artsy/gravity/pull/12985)? I believe they're all stored in Gravity anyway, and that would avoid hard-coding an the test user's id here.

eessex: The part I am unsure how to retrieve from gravity is `Order`

joeyAghion: But do you need to retrieve orders? It looks like you're just using them to eventually retrieve the Stripe cards. This same `Stripe::Customer` is identified by Gravity's `user.stripe_customer.external_id`, so the iterating through sources and deleting can happen there just as easily as here.

eessex:
This logic has been moved to artsy/gravity#12985, closing.

PR #610, opened 2020-05-01
[Toolchain] Add Guard for watching tests

damassi:
Follows [Convection PR](https://github.com/artsy/convection/pull/505) in adding `Guard`, which will watch tests for changes and re-execute, without having to restart command. Additionally adds the ability to focus on specific tests via the `focus: true` param.  

guard

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,7 @@ gem 'taxjar-ruby', require: 'taxjar'
 group :development, :test do
   gem 'byebug'
   gem 'graphlient'
+  gem 'guard-rspec', require: false
   gem 'rspec-rails'
   gem 'rubocop-rails', require: false
 end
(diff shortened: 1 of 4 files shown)


jonallured:
Woot! I know this PR is about the rspec watching, but I think there's a couple more things to do here to get that livereload stuff working:

https://github.com/artsy/convection/blob/master/config/environments/development.rb#L43-L45
https://github.com/artsy/convection/blob/master/Gemfile#L55

I _think_ that's it.

damassi:
@jonallured I copied and pasted that from Convection -- what does the live-reload bit do? Rails already automatically reloads on changes, no?

jonallured:
@damassi I don't think it does it automatically, I think the config I linked to is required but I confess that I did not actually try it on this branch so it's certainly possible I'm wrong!

damassi:
Since this app has little to no UI, updated to remove the live-reload bits in favor of focusing specifically on guard for tests (also added a screen-cap up above).

jonallured:
Nice one! 🐎

PR #611, opened 2020-05-02
[Consignments] If consignment only use artwork location and buyer location for calculating tax

damassi:
Addresses 
- https://artsyproduct.atlassian.net/browse/CSGN-119 
- https://artsyproduct.atlassian.net/browse/CSGN-181

Today our sales tax on BNMO works is calculated by sending taxjar all locations associated with a partner. For direct listings, we will be using one partner account so we only want sales tax to be calculated based on the artwork listing and the buyers address. 

Additionally adds some better logging around our graphql queries via [`graphql-rails_logger`](https://github.com/jetruby/graphql-rails_logger).  

### Artsy

With Tax: 

Screen Shot 2020-05-01 at 11 46 52 PM

Without Tax: 

Screen Shot 2020-05-01 at 11 38 43 PM

### Exchange: 

With Tax: 

Screen Shot 2020-05-02 at 12 23 50 AM

Without Tax: 

Screen Shot 2020-05-02 at 12 24 04 AM

</details>

### DIFF
--- a/app/jobs/record_sales_tax_job.rb
+++ b/app/jobs/record_sales_tax_job.rb
@@ -5,19 +5,12 @@ class RecordSalesTaxJob < ApplicationJob
     line_item = LineItem.find(line_item_id)
     return unless line_item.should_remit_sales_tax?
 
-    artwork = Gravity.get_artwork(line_item.artwork_id)
-    consignment = artwork[:import_source] == 'convection'
+    artwork = line_item.order.artwork
     artwork_address = Address.new(artwork[:location])
+    order = line_item.order
+    nexus_addresses = order.nexus_addresses
 
-    # If the artwork originated from a consignment, the seller location
-    # corresponds to the artwork location for tax purposes.
-    seller_addresses = if consignment
-      [artwork_address]
-    else
-      Gravity.fetch_partner_locations(line_item.order.seller_id, tax_only: true)
-    end
-
-    service = Tax::CollectionService.new(line_item, artwork_address, seller_addresses)
+    service = Tax::CollectionService.new(line_item, artwork_address, nexus_addresses)
     service.record_tax_collected
     line_item.update!(sales_tax_transaction_id: service.transaction.transaction_id) if service.transaction.present?
   end
--- a/app/jobs/record_sales_tax_job.rb
+++ b/app/jobs/record_sales_tax_job.rb
@@ -6,8 +6,15 @@ class RecordSalesTaxJob < ApplicationJob
     return unless line_item.should_remit_sales_tax?
 
     artwork = Gravity.get_artwork(line_item.artwork_id)
+    consignment = artwork[:import_source] == 'convection'
+
     artwork_address = Address.new(artwork[:location])
-    seller_addresses = Gravity.fetch_partner_locations(line_item.order.seller_id, tax_only: true)
+    seller_addresses = if consignment
+      artwork_address
+    else
+      Gravity.fetch_partner_locations(line_item.order.seller_id, tax_only: true)
+    end
+
     service = Tax::CollectionService.new(line_item, artwork_address, seller_addresses)
     service.record_tax_collected
     line_item.update!(sales_tax_transaction_id: service.transaction.transaction_id) if service.transaction.present?
--- a/app/jobs/record_sales_tax_job.rb
+++ b/app/jobs/record_sales_tax_job.rb
@@ -7,10 +7,12 @@ class RecordSalesTaxJob < ApplicationJob
 
     artwork = Gravity.get_artwork(line_item.artwork_id)
     consignment = artwork[:import_source] == 'convection'
-
     artwork_address = Address.new(artwork[:location])
+
+    # If the artwork originated from a consignment, the seller location
+    # corresponds to the artwork location for tax purposes.
     seller_addresses = if consignment
-      artwork_address
+      [artwork_address]
     else
       Gravity.fetch_partner_locations(line_item.order.seller_id, tax_only: true)
     end
--- a/lib/order_helper.rb
+++ b/lib/order_helper.rb
@@ -15,6 +15,14 @@ module OrderHelper
     @artworks ||= line_items.map(&:artwork)
   end
 
+  def artwork
+    @artwork ||= begin
+      # This is with assumption of Offer order only having one line_item. Follows
+      # similar pattern as in offer_totals.
+      line_items.first&.artwork
+    end
+  end
+
   def artists
     @artists ||= artworks.map { |a| a[:artist] }
   end
@@ -32,7 +40,16 @@ module OrderHelper
   end
 
   def seller_locations
-    @seller_locations ||= Gravity.fetch_partner_locations(seller_id, tax_only: true)
+    consignment = artwork[:import_source] == 'convection'
+    artwork_address = Address.new(artwork[:location])
+
+    # If the artwork originated from a consignment, the seller location
+    # corresponds to the artwork location for tax purposes.
+    @seller_locations ||= if consignment
+      [artwork_address]
+    else
+      Gravity.fetch_partner_locations(seller_id, tax_only: true)
+    end
   end
 
   def inventory?
--- a/lib/order_helper.rb
+++ b/lib/order_helper.rb
@@ -39,12 +39,12 @@ module OrderHelper
     @merchant_account ||= Gravity.get_merchant_account(seller_id)
   end
 
-  def seller_locations
+  def nexus_addresses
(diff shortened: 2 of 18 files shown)


lib/shipping_helper.rb
```diff
@@ -1,6 +1,9 @@
 class ShippingHelper
   def self.calculate(artwork, fulfillment_type, shipping_address = nil)
-    return 0 if fulfillment_type == Order::PICKUP
+    consignment = artwork[:import_source] == 'convection'
+    pickup = fulfillment_type == Order::PICKUP
+
+    return 0 if consignment || pickup
```
damassi, line 6: @jonallured - I could be misunderstanding, but for these direct listings we're not factoring in shipping, correct?

spec/controllers/api/requests/offers/offer_set_shipping_mutation_request_spec.rb
```diff
@@ -321,7 +322,7 @@
           end
           context 'with PICKUP as fulfillment type' do
             before do
-              expect(Adapters::GravityV1).to receive(:get).with('/artwork/a-1').and_return(artwork1)
+              expect(Adapters::GravityV1).to receive(:get).twice.with('/artwork/a-1').and_return(artwork1)
```
damassi, line 325: This test had to be modified due [to this line](https://github.com/artsy/exchange/pull/611/files#diff-5db1509729f559a85ed75eca6842595eR43):
```
consignment = artwork[:import_source] == 'convection'
```
which performed a lookup on the artwork.

lib/order_helper.rb
```diff
@@ -32,7 +40,15 @@ def merchant_account
   end
 
   def seller_locations
-    @seller_locations ||= Gravity.fetch_partner_locations(seller_id, tax_only: true)
+    consignment = artwork[:import_source] == 'convection'
+
+    # If the artwork originated from a consignment, the seller location
+    # corresponds to the artwork location for tax purposes.
+    @seller_locations ||= if consignment
+      [Address.new(artwork[:location])]
```
starsirius, line 48: This could be surprising in the future that `seller_locations` returns `[artwork_location]`. Looks like `seller_locations` it's only used as nexus addresses now but folks could be tempted to use it for other purposes.

Maybe we can rename it to be more explicit or push the condition down to where we specify nexus addresses, like [here](https://github.com/artsy/exchange/blob/d78bbf36c24e73575d959bb03ad933fd2a7a665e/lib/offer_totals.rb#L41), and others.

lib/order_helper.rb
```diff
@@ -15,6 +15,14 @@ def artworks
     @artworks ||= line_items.map(&:artwork)
   end
 
+  def artwork
+    @artwork ||= begin
+      # This is with assumption of Offer order only having one line_item. Follows
+      # similar pattern as in offer_totals.
+      line_items.first&.artwork
```
starsirius, line 22: #Minor. Can probably reuse the `artworks` helper above.

lib/tax/calculator_service.rb
```diff
@@ -43,15 +42,15 @@ def artsy_should_remit_taxes?
...
     UnitConverter.convert_dollars_to_cents(
-      @tax_client.tax_for_order(
-        construct_tax_params(
-          line_items: [{
-            unit_price: UnitConverter.convert_cents_to_dollars(@unit_price_cents),
-            quantity: @quantity
-          }]
-        )
-      ).amount_to_collect
+      @tax_client.tax_for_order(tax_params).amount_to_collect
```
jonallured, line 53: Love it, great job extracting a local to make this more readable!!! ❤️ 🏆 🎉

lib/order_helper.rb
jonallured, line 22: Since there's already a method for finding the artworks, my suggestion would be to reuse it and take the first one.

```suggestion
    artworks.first
```

app/jobs/record_sales_tax_job.rb
```diff
@@ -6,8 +6,17 @@ def perform(line_item_id)
...
     artwork_address = Address.new(artwork[:location])
-    seller_addresses = Gravity.fetch_partner_locations(line_item.order.seller_id, tax_only: true)
+
+    # If the artwork originated from a consignment, the seller location
+    # corresponds to the artwork location for tax purposes.
+    seller_addresses = if consignment
+      [artwork_address]
+    else
+      Gravity.fetch_partner_locations(line_item.order.seller_id, tax_only: true)
+    end
```
starsirius, line 18: This can probably reuse the `seller_locations` (or whatever a more explicit name) helper. Something like
```ruby
# ...

order = line_item.order
nexus_addresses = order.seller_locations
service = Tax::CollectionService.new(line_item, artwork_address, nexus_addresses)

# ...
```

lib/shipping_helper.rb
jonallured: I think this is correct but maybe we can confirm with @joanna20?

app/jobs/record_sales_tax_job.rb
starsirius: Unrelated to this PR, looks like there is opportunity to DRY up many duplication in `Tax::CalculatorService` and `Tax::CollectionService`.

spec/lib/shipping_helper_spec.rb
```diff
@@ -68,12 +68,19 @@
...
   describe '#calculate' do
     context 'with pickup fulfillment type' do
       it 'returns 0' do
         expect(ShippingHelper.calculate(artwork, Order::PICKUP)).to eq 0
       end
     end
+    context 'when artwork is consigned' do
+      it 'returns 0' do
+        artwork[:import_source] = 'convection'
+        expect(ShippingHelper.calculate(artwork, Order::PICKUP)).to eq 0
```
jonallured, line 81: I think this test would pass even if the logic you added was not there. I think if we used `Order::SHIP` it would keep us more honest here.

spec/models/order_spec.rb
```diff
@@ -406,4 +406,25 @@
       end
     end
   end
+
+  describe '#seller_locations' do
```
jonallured, line 410: These specs look fantastic, thanks for taking the time to write them up!! ❤️

app/jobs/record_sales_tax_job.rb
```diff
@@ -6,8 +6,17 @@ def perform(line_item_id)
     return unless line_item.should_remit_sales_tax?
 
     artwork = Gravity.get_artwork(line_item.artwork_id)
+    consignment = artwork[:import_source] == 'convection'
     artwork_address = Address.new(artwork[:location])
-    seller_addresses = Gravity.fetch_partner_locations(line_item.order.seller_id, tax_only: true)
+
+    # If the artwork originated from a consignment, the seller location
+    # corresponds to the artwork location for tax purposes.
+    seller_addresses = if consignment
```
jonallured, line 14: I know this goes against what we originally talked about, but having seen the entire diff here, I think it would be cool to get this business logic isolated just in the OfferHelper. Can we do something like this here instead:

```ruby
order = line_item.order
seller_addresses = order.seller_addresses
```

That way we don't have to duplicate this logic and there's one source of truth about which addresses are the seller addresses. Does that seem like it would work?

damassi: I played with this a bit before when trying to access an `artwork` from the `line_item` but for some reason it seems like this Job doesn't have access to helper logic (`artwork` undefined, `seller_addresses` undefined, etc). Does something need to be included at this level in order to register the helper?

lib/order_helper.rb
damassi: Updated from `seller_locations` -> `nexus_addresses`.

damassi:
This is ready for final feedback!

spec/controllers/api/requests/set_shipping_mutation_request_spec.rb
```diff
@@ -240,6 +240,7 @@
           it 'returns an error' do
             allow(Adapters::GravityV1).to receive(:get).with("/partner/#{seller_id}/locations", params: { private: true, address_type: ['Business', 'Sales tax nexus'], page: 1, size: 20 }).and_return([{ country: 'US', state: 'NY' }])
             allow(Adapters::GravityV1).to receive(:get).with('/artwork/a-1').and_return(id: 'missing-location')
+            allow(Adapters::GravityV1).to receive(:get).with('/artwork/a-2').and_return(id: 'missing-location')
```
damassi, line 243: Noticed this test was failing with this required to get it to pass. I'm a bit uncertain why exactly this _wasn't_ here before, as all of the tests preceding and following (minus the one just below) have a similar setup.

PR #636, opened 2020-08-31
[PURCHASE-1715] Add 'Confirm offline sale' button

sepans:
Fixes [PURCHASE-1715](https://artsyproduct.atlassian.net/browse/PURCHASE-1715)

Problem:
MSO team is not able to report offline sales in exchange. This involves changing the state from `canceled` or `abandoned` (why? [discussion here](https://artsyproduct.atlassian.net/browse/PURCHASE-1715?focusedCommentId=31124)) to `fulfilled` as well as editing prices and the fulfillment date. There is no need to capture credit card or send emails since the sale has happened off-platform.

Solution:
Admins now have the ability to confirm offline sale on canceled and abandoned orders and edit prices/date. This also automatically adds an 'Admin note'.

The button (only appears on abandoned/canceled orders):

Screen Shot 2020-08-31 at 4 52 41 PM

This form shows up after pressing the button:

Screen Shot 2020-08-31 at 4 53 16 PM

Changes to order state:

Screen Shot 2020-08-31 at 4 54 01 PM

</details>

~To verify:
I checked mailtrap and the `Confirm fulfillment` button doesn't send emails to the collector so this shouldn't send it either but need to verify that on staging. If you think that is not the case let me know.~  see https://github.com/artsy/exchange/pull/636#issuecomment-685780756

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -83,18 +83,6 @@ ActiveAdmin.register Order do
     redirect_to resource_path, notice: 'toggled assisted flag!'
   end
 
-  member_action :confirm_offline_sale, method: :post do
-    if resource.fulfillment_type == Order::PICKUP
-      OrderService.confirm_pickup!(resource, current_user[:id])
-    elsif  resource.fulfillment_type == Order::SHIP
-      OrderService.confirm_fulfillment!(resource, current_user[:id], fulfilled_by_admin: true)
-    else
-      raise Errors::ValidationError, :wrong_fulfillment_type
-    end
-    redirect_to edit_admin_order_path(resource[:id]), notice: 'Offline sale created. you can now edit the price'
-  end
-
-
   action_item :refund, only: :show do
     link_to 'Refund', refund_admin_order_path(order), method: :post, data: { confirm: 'Are you sure you want to refund this order?' } if [Order::APPROVED, Order::FULFILLED].include? order.state
   end
@@ -128,7 +116,7 @@ ActiveAdmin.register Order do
   end
 
   action_item :confirm_offline_sale, only: :show do
-    link_to 'Confirm Offline Sale', confirm_offline_sale_admin_order_path(order), method: :post if [Order::ABANDONED, Order::CANCELED].include?(order.state)
+    link_to 'Confirm Offline Sale', edit_admin_order_path(order[:id]) if [Order::ABANDONED, Order::CANCELED].include?(order.state)
   end
 
   sidebar :artwork_info, only: :show do
@@ -426,6 +414,15 @@ ActiveAdmin.register Order do
     def update
       offline_sale_date = params[:order].delete(:offline_sale_date)
       admin_note_description = params[:order].delete(:admin_note_description)
+
+      if resource.fulfillment_type == Order::PICKUP
+        OrderService.confirm_pickup!(resource, current_user[:id])
+      elsif  resource.fulfillment_type == Order::SHIP
+        OrderService.confirm_fulfillment!(resource, current_user[:id], fulfilled_by_admin: true)
+      else
+        raise Errors::ValidationError, :wrong_fulfillment_type
+      end
+
       resource.state_histories.last.update_attributes(:created_at=> offline_sale_date, :updated_at=> offline_sale_date)
       resource.admin_notes.create!(note_type: AdminNote::TYPES[:offline_sale], admin_id: current_user[:id], description: admin_note_description)
 
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -417,7 +417,10 @@ ActiveAdmin.register Order do
 
       OrderService.confirm_fulfillment!(resource, current_user[:id], fulfilled_by_admin: true)
 
-      resource.state_histories.last.update_attributes(:created_at=> offline_sale_date, :updated_at=> offline_sale_date)
+      # update fulfilled state change timestamp to the `offline_sale_date` provided in the form
+      fulfillment_state = resource.state_histories.where(state: Order::FULFILLED).order(created_at: :asc).last
+      fulfillment_state.update!(created_at: offline_sale_date) if fulfillment_state
+
       resource.admin_notes.create!(note_type: AdminNote::TYPES[:offline_sale], admin_id: current_user[:id], description: admin_note_description)
 
       update! do |format|
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -69,7 +69,7 @@ ActiveAdmin.register Order do
   end
 
   member_action :confirm_pickup, method: :post do
-    OrderService.confirm_pickup!(resource, current_user[:id]) if resource.fulfillment_type == Order::PICKUP
+    OrderService.confirm_fulfillment!(resource, current_user[:id], fulfilled_by_admin: true) if resource.fulfillment_type == Order::PICKUP
     redirect_to resource_path, notice: 'Fulfillment confirmed!'
   end
 
@@ -415,13 +415,7 @@ ActiveAdmin.register Order do
       offline_sale_date = params[:order].delete(:offline_sale_date)
       admin_note_description = params[:order].delete(:admin_note_description)
 
-      if resource.fulfillment_type == Order::PICKUP
-        OrderService.confirm_pickup!(resource, current_user[:id])
-      elsif  resource.fulfillment_type == Order::SHIP
-        OrderService.confirm_fulfillment!(resource, current_user[:id], fulfilled_by_admin: true)
-      else
-        raise Errors::ValidationError, :wrong_fulfillment_type
-      end
+      OrderService.confirm_fulfillment!(resource, current_user[:id], fulfilled_by_admin: true)
 
       resource.state_histories.last.update_attributes(:created_at=> offline_sale_date, :updated_at=> offline_sale_date)
       resource.admin_notes.create!(note_type: AdminNote::TYPES[:offline_sale], admin_id: current_user[:id], description: admin_note_description)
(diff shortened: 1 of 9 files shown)


app/admin/orders.rb
```diff
@@ -82,6 +83,18 @@
     redirect_to resource_path, notice: 'toggled assisted flag!'
   end
 
+  member_action :confirm_offline_sale, method: :post do
+    if resource.fulfillment_type == Order::PICKUP
+      OrderService.confirm_pickup!(resource, current_user[:id])
+    elsif  resource.fulfillment_type == Order::SHIP
+      OrderService.confirm_fulfillment!(resource, current_user[:id], fulfilled_by_admin: true)
+    else
+      raise Errors::ValidationError, :wrong_fulfillment_type
```
sepans, line 92: `fulfillment_type` can be `nil` if users have abandoned orders before getting to the shipment page.

app/admin/orders.rb
```diff
@@ -391,4 +407,31 @@
...
+      f.input :seller_total_cents, as: :number, label: "Seller Payout (#{order.currency_code} cents)"
+    end
+    f.actions
+  end
+
+  controller do
+    def update
+      offline_sale_date = params[:order].delete(:offline_sale_date)
+      admin_note_description = params[:order].delete(:admin_note_description)
+      resource.state_histories.last.update_attributes(:created_at=> offline_sale_date, :updated_at=> offline_sale_date)
```
sepans, line 429: Updating `created_at` and `updated_at` is not a great idea but not sure if there is a better way of setting the 'sale' date 🤔

app/admin/orders.rb
```diff
@@ -82,6 +83,18 @@
     redirect_to resource_path, notice: 'toggled assisted flag!'
   end
 
+  member_action :confirm_offline_sale, method: :post do
+    if resource.fulfillment_type == Order::PICKUP
+      OrderService.confirm_pickup!(resource, current_user[:id])
+    elsif  resource.fulfillment_type == Order::SHIP
+      OrderService.confirm_fulfillment!(resource, current_user[:id], fulfilled_by_admin: true)
```
starsirius, line 90: Looks like there is opportunity to refactor and consolidate `OrderService.confirm_pickup!` and `OrderService.confirm_fulfillment!`. It feels a bit strange to have `confirm_fulfillment!` only for shipping and then have a specialized `confirm_pickup!` that does similar things. Maybe we can have a single method that can handle both fulfillment types and admin use cases.

```ruby
def self.confirm_fulfillment!(order, user_id, fulfillment_type: nil, fulfilled_by_admin: false)
  # implementation
end
```

app/admin/orders.rb
```diff
@@ -82,6 +83,18 @@
...
+    else
+      raise Errors::ValidationError, :wrong_fulfillment_type
+    end
+    redirect_to edit_admin_order_path(resource[:id]), notice: 'Offline sale created. you can now edit the price'
```
starsirius, line 94: **suggestion ** Edit details and confirm fulfillment at the same time

The order of operations seems to be reversed here:

1. we fulfilled the order first, and then
1. edit the details (e.g. price, fees, etc.) on the next page.

Given the separate steps, it's possible to fulfill an order without important information filled in.

It's probably better to
- Simply render the edit page when clicking "Confirm Offline Sale", and
- After submitting the form, we update the order and move the state to be fulfilled together.

app/admin/orders.rb
starsirius: If we move the confirm fulfillment here together with `update`, we might not need to update them?

app/admin/orders.rb
sepans: 💯 Great idea

app/admin/orders.rb
starsirius: Oh realized we are referring to `offline_sale_date` here. I think it makes sense to _not_ update the state history timestamps to `offline_sale_date` because it's technically transitioned to the fulfilled state when admin confirms it here.

If we need record `offline_sale_date`, maybe we can do it in a first class way, or append it in the [reason](https://github.com/artsy/exchange/blob/5119e7d173701a9b01d908fc9fa978b73e3a221f/db/schema.rb#L201).

app/admin/orders.rb
starsirius: **Question**: When an order is submitted (or offer is accepted), we [deduct the inventory](https://github.com/artsy/exchange/blob/2d1e48f2b00a6164db0797ed1879030aa3b6951b/app/services/order_service.rb#L53) and [mark the work as sold if the work becomes not acquireable](https://github.com/artsy/gravity/blob/1f90d7dc62b91036583b046a81d30c0f92b61c0e/app/services/inventory_service.rb#L36). Do we need to worry about that for offline sale?

app/admin/orders.rb
sepans: Isable mentioned updating the date is nice to have so I can drop this line. Also, I couldn't figure out how the sale date is propagated to other systems for accounting. But. the `state_histories` data is updated automatically [here](https://github.com/artsy/exchange/blob/32795b3d14088e5fee81d16dc5461bcb4eb09533/app/models/order.rb#L241) immediately after each state transition and the only way to update that, as far as I could figure out, is updating `created_at` and `updated_at` after creating a new history object.

app/admin/orders.rb
sepans: I don't think so because the partner has already manually done that in CMS. I can verify it.

app/admin/orders.rb
sepans: in https://github.com/artsy/exchange/pull/636/commits/3b8e94093b6c7c34a74c59c7630aff276d00735f

app/admin/orders.rb
sepans: > suggestion (refactor): Consolidate OrderService.confirm_pickup! and OrderService.confirm_fulfillment!

I am a little scared of doing that 😄 but will dig deeper to see how those methods are being used

starsirius: Happy to pair on the refactoring! 💪 

> I don't think so because the partner has already manually done that in CMS. I can verify it.
Partners can manually update the availability to sold, but unfortunately they can't update inventory in CMS for non-editioned works. It might be sufficient with the availability change (so buyers can't buy it again on artsy), but inventory count may not be accurate in this case. However, I agree that it's probably better to let partners handle the artwork update.

sepans:
> Re: To verify: not sending emails to the collector

haha I accidentally came across https://github.com/artsy/pulse/pull/557 and in fact, I was the person who made the change not to send emails to collector when the order is fulfilled by admin. Just need to verify it is also the case for pickup orders

app/admin/orders.rb
sepans: > refactor

I looked deeper into the code. It is doable but I am kind of hesitant about it. The logic of `confirm_pickup` and `confirm_fullfilment` are very similar at the moment, the only difference in business logic is sending emails to the collector which pulse takes care of. It is possible that the logic diverges in the future and in that case we would better have separate methods so 🤷‍♂️

app/graphql/mutations/confirm_fulfillment.rb
```diff
@@ -8,6 +8,9 @@ class Mutations::ConfirmFulfillment < Mutations::BaseMutation
   def resolve(id:)
     order = Order.find(id)
     authorize_seller_request!(order)
+
+    raise Errors::ValidationError, :wrong_fulfillment_type unless order.fulfillment_type == Order::SHIP
```
sepans, line 12: Added this validation (also in confirm pickup mutation) to verify the fulfillment type for this specific mutation since both are calling `confirm_fulfillment!` now and it accepts both fulfillment types.

starsirius: Makes sense to me! Realized the mutation name is also confusing, and it's only for shipping fulfillment type. Let's not explode the scope of this PR and can address it separately.

starsirius: Ohhh I just learned that we actually have 3 fulfillment mutations and service methods:
- [`FulfillAtOnce`](https://github.com/artsy/exchange/blob/5119e7d173701a9b01d908fc9fa978b73e3a221f/app/graphql/mutations/fulfill_at_once.rb) (for `SHIP` fulfillment type): I think this is what Volt uses for fulfilling shipping orders.
- [`ConfirmFulfillment`](https://github.com/artsy/exchange/blob/5119e7d173701a9b01d908fc9fa978b73e3a221f/app/graphql/mutations/confirm_fulfillment.rb) (for `SHIP` fulfillment type): Curious how/if this is used. 🤔 
- [`ConfirmPickup`](https://github.com/artsy/exchange/blob/5119e7d173701a9b01d908fc9fa978b73e3a221f/app/graphql/mutations/confirm_pickup.rb) (for `PICKUP` fulfillment type): I think this is what Volt uses for fulfilling pickup orders.

One refactor at a time I guess. 😝

app/admin/orders.rb
starsirius: Ah realized we use [`default_scope`](https://github.com/artsy/exchange/blob/5119e7d173701a9b01d908fc9fa978b73e3a221f/app/models/state_history.rb#L4) to sort state histories. Shouldn't be in this PR, but I'll be interested in replacing that with explicit sorting given it usually [creates more problems](https://stackoverflow.com/questions/25087336/why-is-using-the-rails-default-scope-often-recommend-against) in the future.

starsirius (approved):
Thanks for improving our admin workflow!! I also learn so much about active admin! @sepans 💯

app/admin/orders.rb
sepans: 👍 in https://github.com/artsy/exchange/pull/636/commits/d41b863265652ae23a08b3b6017ce5844f5497f4

PR #637, opened 2020-09-11
Remove order edit button from active admin order page

sepans:
<details><summary>Screenshots</summary>

Before:

Screen Shot 2020-09-11 at 1 11 39 PM

After:

Screen Shot 2020-09-11 at 1 11 49 PM

</details>

Updating `actions` to `actions :all, except: %i[create destroy new]` in https://github.com/artsy/exchange/pull/636 added the `edit_admin_order_path` path (used for confirm sale button) as well as adding the actual button to page. 
That new `edit` button is not needed and also dangerous. This piece of code removes that. [source](https://stackoverflow.com/questions/44723359/how-do-you-remove-the-edit-button-in-an-active-admin-show)

PR #638, opened 2020-09-11
Enable Rubocop for active admin files

starsirius:
Noticed in https://github.com/artsy/exchange/pull/637#discussion_r487205381 Rubocop wasn't running for Active Admin files. Instead of disabling Rubocop entirely for Active Admin files, let's only disable `Metrics/BlockLength` rule for them (given the Active Admin files tend to be long).

I ran `bundle exec rubocop -a` and manually fixed a few (commented inline).

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -79,7 +79,7 @@ ActiveAdmin.register Order do
   end
 
   member_action :toggle_assisted, method: :post do
-    resource.toggle!(:assisted)
+    resource.update!(assisted: !resource.assisted)
     redirect_to resource_path, notice: 'toggled assisted flag!'
   end
 
@@ -210,8 +210,8 @@ ActiveAdmin.register Order do
         partner_info[:partner_location] = partner_location
         attributes_table_for partner_info do
           row :name
-          row :partner_location do |partner_info|
-            partner_location = partner_info[:partner_location]
+          row :partner_location do |partner|
+            partner_location = partner[:partner_location]
             div partner_location.street_line1
             div partner_location.street_line2
             div "#{partner_location.city}, #{partner_location.region} #{partner_location.postal_code}"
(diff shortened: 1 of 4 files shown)


app/admin/orders.rb
```diff
@@ -79,7 +79,7 @@
   end
 
   member_action :toggle_assisted, method: :post do
-    resource.toggle!(:assisted)
+    resource.update!(assisted: !resource.assisted)
```
starsirius, line 82:
```
app/admin/orders.rb:82:14: C: Rails/SkipsModelValidations: Avoid using toggle! because it skips validations.
```

app/admin/orders.rb
```diff
@@ -210,8 +210,8 @@
         partner_info[:partner_location] = partner_location
         attributes_table_for partner_info do
           row :name
-          row :partner_location do |partner_info|
-            partner_location = partner_info[:partner_location]
+          row :partner_location do |partner|
+            partner_location = partner[:partner_location]
```
starsirius, line 214:
```
app/admin/orders.rb:213:37: W: Lint/ShadowingOuterLocalVariable: Shadowing outer local variable - partner_info.
```

app/admin/orders.rb
```diff
@@ -398,8 +398,8 @@
 
   form do |f|
     f.inputs do
-      f.input :offline_sale_date, as: :date_picker, input_html: { value: Date.today }, label: 'Offline sale date' 
-      f.input :admin_note_description, as: :string, input_html: { value: '' }, label: 'Admin note' 
+      f.input :offline_sale_date, as: :date_picker, input_html: { value: Time.zone.today }, label: 'Offline sale date'
```
starsirius, line 401:
```
app/admin/orders.rb:401:79: C: Rails/Date: Do not use Date.today without zone. Use Time.zone.today instead.
```

sepans:
🏅 Thank you for adding it.

PR #639, opened 2020-09-16
[PURCHASE-2113] Make price/offer editable

sepans:
Fixes https://artsyproduct.atlassian.net/browse/PURCHASE-2113

Follow up to https://github.com/artsy/exchange/pull/636

When creating the offline sale information form forgot to include the artwork price and offer price. This PR adds that. Offer price is straight forward since it is an attribute on `Order` model but artwork price was a bit more tricky because it is derived from sum of `LineItem` prices times their quantity. We currently only have a single `LineItem` on orders and the quantity is always 1 so the new function only handles that case and throws an error if that's not the case. We have a few old orders on production that were manually modified and have multiple line items but no need to worry about that since this 'Offline Sale' is not applicable to them.

Screen Shot 2020-09-16 at 4 38 13 PM

### DIFF
--- a/app/admin/orders.rb
+++ b/app/admin/orders.rb
@@ -1,5 +1,5 @@
 ActiveAdmin.register Order do
-  permit_params :shipping_total_cents, :tax_total_cents, :buyer_total_cents, :transaction_fee_cents, :commission_fee_cents, :seller_total_cents
+  permit_params :shipping_total_cents, :tax_total_cents, :buyer_total_cents, :transaction_fee_cents, :commission_fee_cents, :seller_total_cents, :items_total_cents
   actions :all, except: %i[create destroy new]
   # TODO: change sort order
   config.sort_order = 'state_updated_at_desc'
@@ -402,6 +402,8 @@ ActiveAdmin.register Order do
     f.inputs do
       f.input :offline_sale_date, as: :date_picker, input_html: { value: Time.zone.today }, label: 'Offline sale date'
       f.input :admin_note_description, as: :string, input_html: { value: '' }, label: 'Admin note'
+      f.input :total_list_price_cents, as: :number, input_html: { value: order.total_list_price_cents }, label: "Artwork list price (#{order.currency_code} cents)"
+      f.input :items_total_cents, as: :number, label: "Accepted offer (#{order.currency_code} cents)" if order.mode == Order::OFFER
       f.input :shipping_total_cents, as: :number, label: "Shipping (#{order.currency_code} cents)"
       f.input :tax_total_cents, as: :number, label: "Sales Tax (#{order.currency_code} cents)"
       f.input :buyer_total_cents, as: :number, label: "Buyer Paid (#{order.currency_code} cents)"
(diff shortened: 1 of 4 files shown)


spec/models/order_spec.rb
```diff
@@ -427,4 +427,32 @@
       end
     end
   end
+
+  describe 'update_total_list_price_cents' do
+    context 'when order has only one line item' do
+      it 'updates price on first line item' do
+        order = Fabricate(:order, line_items: [Fabricate(:line_item, artwork_id: 'id-0')])
+        order.update_total_list_price_cents(256)
+        expect(order.line_items.first.total_list_price_cents).to be(256)
```
starsirius, line 436: Maybe we want to assert `order.total_list_price_cents` here?

app/controllers/errors/error_types.rb
```diff
@@ -35,12 +35,12 @@ module Errors
       missing_required_info
       missing_required_param
       missing_phone_number
+      more_than_one_line_item
       no_taxable_addresses
       not_acquireable
       not_found
       not_last_offer
       not_offerable
-      offer_more_than_one_line_item
```
starsirius, line 43: Thanks for cleaning it up!

app/models/order.rb
```diff
@@ -188,6 +188,12 @@ def total_list_price_cents
     line_items.map(&:total_list_price_cents).sum
   end
 
+  def update_total_list_price_cents(price)
+    raise Errors::ValidationError, :more_than_one_line_item unless line_items.count == 1 && line_items.first.quantity == 1
```
starsirius, line 192: 👍

app/admin/orders.rb
```diff
@@ -402,6 +402,8 @@
     f.inputs do
       f.input :offline_sale_date, as: :date_picker, input_html: { value: Time.zone.today }, label: 'Offline sale date'
       f.input :admin_note_description, as: :string, input_html: { value: '' }, label: 'Admin note'
+      f.input :total_list_price_cents, as: :number, input_html: { value: order.total_list_price_cents }, label: "Artwork list price (#{order.currency_code} cents)"
+      f.input :items_total_cents, as: :number, label: "Accepted offer (#{order.currency_code} cents)" if order.mode == Order::OFFER
```
starsirius, line 406: Some data might become inconsistent with these ad-hoc changes, e.g. the artwork price doesn't match with the price of the particular artwork version, `items_total_cents` doesn't match with associated offer records. Is [`fulfilled_by_admin_id`](https://github.com/artsy/exchange/blob/2aa2dc87a9320f362174dbb57b7ff2e30f4e3f3d/app/admin/orders.rb#L420) a good way to denote that it's manually updated?

starsirius (approved):
Have a few minor comments and feel free to merge when ready! 💯

app/admin/orders.rb
sepans: That's a good point. I didn't see any places the gravity artwork price being used except for [creating the line item ](https://github.com/artsy/exchange/blob/518c7edd80ca88e1128cc931273c15667420330c/lib/order_creator.rb#L125) so seems to be fine. There will definitely be inconsistencies but this is not going to happen [very often](https://artsyproduct.atlassian.net/browse/PURCHASE-2091?jql=project%20%3D%20PURCHASE%20AND%20status%20in%20(%22Define%22%2C%20Ready)%20AND%20text%20~%20%22fulfilled%20offline%20sale%22%20AND%20Sprint%20%3D%20255) and the process is going to be very manual and if anything goes wrong we can fix the data using the console.

spec/models/order_spec.rb
sepans: 👍

sepans:
Thanks @starsirius merged!

PR #641, opened 2020-10-07
Change `myOrders` query to accept multiple states

sepans:
#### Problem:

On the purchase history page, we don't want to show the `abandoned` orders but if we do the filtering on the client-side the pagination wouldn't work as expected so we need to do the filtering on the server-side,

#### Solution:

This will change the `state` params to `states` to be able to pass an array of states (everything except 'abandoned') from the client.

#### Breaking change 🚨 

This is a schema breaking change but since we haven't launched purchase history yet it is safe to synchronously deploy it

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -22,7 +22,7 @@ class Types::QueryType < Types::BaseObject
   field :my_orders, Types::OrderConnectionWithTotalCountType, null: true, connection: true do
     description 'Return my orders'
     argument :seller_id, String, required: false
-    argument :state, Types::OrderStateEnum, required: false
+    argument :states, [Types::OrderStateEnum], required: false
     argument :sort, Types::OrderConnectionSortEnum, required: false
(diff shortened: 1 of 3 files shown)


app/graphql/types/query_type.rb
```diff
@@ -56,6 +56,8 @@ def orders(params = {})
   def my_orders(params = {})
     raise ActiveRecord::RecordNotFound unless context[:current_user][:id]
 
+    states = params.delete(:states)
+    params = params.merge(state: states) if states.present?
```
starsirius, line 60: I think the new schema
```graphql
states: [OrderStateEnum!]
```
accepts `[]` as the states, which will get ignored here because it's not present. From the API perspective it makes sense to return nothing in this case, instead of all. Adding a test for the edge case helps here too!

sepans: Great catch. Fixed in https://github.com/artsy/exchange/pull/641/commits/0924be25c3c2beba618f1f23b4fb3fb43d859c9b

PR #644, opened 2020-10-22
 fix: prefer tzinfo-data gem over tzdata package

anandaroop:
Fixes a silent failure regarding time zone parsing, resulting from updates to the IANA time zone database.

This follows the pattern in:

- https://github.com/artsy/vortex/pull/279
- https://github.com/artsy/diffusion/pull/347
- https://github.com/artsy/volt/pull/4697

Discussion here:
https://artsy.slack.com/archives/C9ZJYFDNF/p1603137517274800

and here:
https://artsy.slack.com/archives/C02BC3HEJ/p1603289217394000

### DIFF
--- a/spec/lib/time_zone_spec.rb
+++ b/spec/lib/time_zone_spec.rb
@@ -1,0 +1,9 @@
+require 'rails_helper'
+
+describe 'time zone parsing' do
+  it 'handles daylight saving time properly' do
+    utc = '2020-12-01T00:00:00+00:00'
+    est = '2020-11-30T19:00:00-05:00'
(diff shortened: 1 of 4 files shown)


spec/lib/time_zone_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+# require 'support/gravity_helper'
```
starsirius, line 2: 🙈

anandaroop: Lol, I'll delete it 😭

starsirius (approved):
Thanks for pairing! Feel free to merge after removing the extra comment!

PR #657, opened 2020-12-22
[no-merge] chore: use rails-erd to visualize models

starsirius:
This uses the [rails-erd](https://github.com/voormedia/rails-erd) gem to visualize Exchange modeling.

This PR is mainly for demonstration purposes but we can also consider merging if folks feel it's good documentation. It supports [auto generation](https://github.com/voormedia/rails-erd/tree/v1.6.0#auto-generation) when running `db:migrate`, so I assume the diagram will be re-generated when we update the modeling.

It does require `graphviz` to generate the diagrams.

---

erd


erikdstock:
is `graphviz` a separate requirement on the local machine or managed through the bundle? If the latter I'm :+1: on this.

erikdstock (approved):

starsirius:
Thanks for reminding of this PR. With the current setup `graphviz` is an external dependency. This PR is mainly for demo purposes so I'm going to close it for now! Thanks for reviewing!

PR #665, opened 2021-01-13
chore: Adds impulse conversation id column

ashleyjelks:
Adds conversation id to order table to associate an order with a conversation.

cc @artsy/purchase-devs 

https://artsyproduct.atlassian.net/browse/PURCHASE-2345

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -2,15 +2,15 @@
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
 #
-# This file is the source Rails uses to define your schema when running `rails
-# db:schema:load`. When creating a new database, `rails db:schema:load` tends to
+# This file is the source Rails uses to define your schema when running `bin/rails
+# db:schema:load`. When creating a new database, `bin/rails db:schema:load` tends to
 # be faster and is potentially less error prone than running all of your
 # migrations from scratch. Old migrations may fail to apply correctly if those
 # migrations use external dependencies or application code.
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_03_06_192634) do
+ActiveRecord::Schema.define(version: 2021_01_15_162614) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -231,6 +232,16 @@
     t.index ["order_id"], name: "index_transactions_on_order_id"
   end
 
+  create_table "versions", force: :cascade do |t|
```
ashleyjelks, line 235: Is this an expected change, since a new model was not created?? cc @starsirius

zephraph:
Excited to see this! I'll defer to @starsirius for the final approval, but it looks good to me.

db/migrate/20210113224423_add_impulse_conversation_id_to_order.rb
```diff
@@ -0,0 +1,6 @@
+class AddImpulseConversationIdToOrder < ActiveRecord::Migration[6.0]
+  def change
+    add_column :orders, :impulse_conversation_id, :string
```
starsirius, line 3: It's likely we'll be querying orders by `impulse_conversation_id` so let's also add an index on it. I'm not sure how to _just_ create the index with a rails migration, but potentially you can start over and run
```
bundle exec rails generate migration add_impulse_conversation_id_index_to_orders impulse_conversation_id:string:index
```
to add the column _and_ the index.

Or you can add ☝️  migration on top of this and delete the `add_column` line.

db/schema.rb
starsirius: Interesting. We [renamed](https://github.com/artsy/exchange/blob/master/db/migrate/20181204180921_rename_versions_to_order_versions.rb) `versions` table to `order_versions` in a previous migration, and I'm not sure why this gets generated again.

If it's ok to drop your local database, maybe we can try resetting the database?
```
bundle exec rails db:migrate:reset
```
I think this will drop, create, and migrate the database.

starsirius:
I'm expecting to see addition of the `impulse_conversation_id` column and index in db/schema.rb. Maybe you missed running `bundle exec rails db:migrate` again?

db/migrate/20210113224423_add_impulse_conversation_id_to_order.rb
ashleyjelks: > I'm expecting to see addition of the `impulse_conversation_id` column and index in db/schema.rb. Maybe you missed running `bundle exec rails db:migrate` again?

Woops, I dropped the last commit and I think the db changes were lost. 😅

db/schema.rb
```diff
@@ -23,11 +23,11 @@
     t.uuid "resource_id"
     t.string "author_type"
     t.uuid "author_id"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-    t.index ["author_type", "author_id"], name: "index_active_admin_comments_on_author_type_and_author_id"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.index ["author_type", "author_id"], name: "index_active_admin_comments_on_author"
```
sepans, line 28: The precision [used to be there](https://github.com/artsy/exchange/commit/79f65421a3d645e475caef045b1310a1d6affef3#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3L27) so maybe removed by accident. This  index name changes shouldn't be a big deal either 🤷🏽‍♂️

sepans (approved):
🚀

db/schema.rb
starsirius: Yeah looks like in Rails 6 would now [generates timestamps with precision](https://github.com/rails/rails/pull/34970), but I'm not sure why the previous PR that's with Rails 6.0 migration removed it. I don't think that's a blocker tho.

starsirius:
Merging! Thanks @ashleyjelks !

PR #666, opened 2021-01-15
fix: remove unsupported daemonization in Puma 5

starsirius:
`daemonize` has been removed in Puma 5. Removing it should behave the same since we want to run it on the foreground.

> tcp mode and daemonization have been removed without replacement. For
daemonization, please use a modern process management solution, such as
systemd or monit.

https://github.com/puma/puma/blob/master/5.0-Upgrade.md#upgrade

The error can be reproduced when running `BUNDLE_GITHUB__COM=... hokusai dev start`:

Screen Shot 2021-01-15 at 3 27 49 PM

### DIFF
--- a/config/puma.config
+++ b/config/puma.config
@@ -2,7 +2,6 @@
 
 directory "/app"
 environment ENV.fetch("RAILS_ENV") { "development" }
-daemonize false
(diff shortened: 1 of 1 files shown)


sepans:
Thanks @starsirius ❤️

PR #668, opened 2021-01-15
fix: allow postgres connection without password in hokusai dev

starsirius:
When running `hokusai dev start`, postgres gives an error of

```
Database is uninitialized and superuser password is not specified.`
```

This follows the prompt and existing pattern in [Gravity](https://github.com/artsy/gravity/blob/6d5ea778793cf9ce095494a3cc55a0485b207efd/hokusai/development.yml#L113) to set

```
POSTGRES_HOST_AUTH_METHOD=trust
```

and allow connection without password in hokusai dev.

Screen Shot 2021-01-15 at 4 26 19 PM

### DIFF
--- a/hokusai/development.yml
+++ b/hokusai/development.yml
@@ -19,6 +19,8 @@ services:
       - exchange-db
   exchange-db:
     image: postgres:9.6-alpine
+    environment:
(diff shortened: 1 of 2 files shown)


sepans:
🙏🏽

PR #672, opened 2021-01-26
[PURCH-2346] Feat: Add impulseConversationId input field to createOrderWithArtwork

ashleyjelks:
Introduces an optional input param `impulseConversationId` on the `createOrderWithArtwork` mutation.

https://artsyproduct.atlassian.net/browse/PURCHASE-2346

cc @artsy/purchase-devs 

Screen Shot 2021-01-26 at 3 17 16 PM


sepans (commented):
Nice 💯 

Seems some tests are not happy with the new parameter (without default) added to `create_with_artwork` and/or `initialize` methods other than that it looks good.

We can also add [tests](https://github.com/artsy/exchange/blob/master/spec/controllers/api/requests/offers/create_offer_order_with_artwork_mutation_request_spec.rb) for this.

My preference was to add a new mutation like `createInquiryOfferOrder` or something, just to make it very explicit that this mutation is different for future debugging but don't feel strongly about it.

starsirius:
@sepans had a good point about adding a new mutation instead (such as `createInquiryOfferOrderWithArtwork`) so it's more aligned with our GraphQL patterns (i.e. can be more specialized for individual use cases) and helpful for tracing purposes. We can still reuse the underlying services/libs so it's DRY.

@ashleyjelks let me know if I can pair on converting it to a new mutation!

spec/controllers/api/requests/offers/create_inquiry_offer_order_with_artwork_mutation_request_spec.rb
```diff
@@ -0,0 +1,367 @@
...
+          end.to change(Order, :count).by(0).and change(LineItem, :count).by(0)
+        end
+      end
+
+      context 'with successful artwork fetch' do
+        let(:artwork) { gravity_v1_artwork }
+        before do
+          expect(Gravity).to receive(:get_artwork).once.with(artwork_id).and_return(artwork)
+        end
+        context 'artwork with one edition set' do
```
sepans, line 99: suggestion:
If most of these tests are the same as `createOrderWithArtwork` mutation tests it would be good to move them to [shared examples](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples) (or any other rspec magic) and reuse the tests.

spec/controllers/api/requests/offers/create_inquiry_offer_order_with_artwork_mutation_request_spec.rb
```diff
@@ -0,0 +1,367 @@
...
+        end
+      end
+
+      context 'without impulseConversationId' do
+        it 'requires impulse conversation id' do
+          expect do
+            client.execute(mutation, input: { artworkId: '198274821', impulseConversationId: nil })
+          end.to raise_error do |error|
+            expect(error).to be_a(Graphlient::Errors::GraphQLError)
+            expect(error.message).to match(/impulseConversationId \(Expected value to not be null\)/)
```
sepans, line 74: 💯

spec/controllers/api/requests/offers/create_inquiry_offer_order_with_artwork_mutation_request_spec.rb
```diff
@@ -0,0 +1,367 @@
...
+                expect(order.seller_id).to eq seller_id
+                expect(order.line_items.count).to eq 1
+                expect(order.line_items.first.list_price_cents).to eq 5400_12
+                expect(order.line_items.first.artwork_id).to eq 'artwork-id'
+                expect(order.line_items.first.edition_set_id).to be_nil
+                expect(order.line_items.first.quantity).to eq 1
+              end.to change(Order, :count).by(1).and change(LineItem, :count).by(1)
+            end
+          end
+          context 'impulseConversationId' do
```
sepans, line 267: I don't know if we have a separate ticket for that, but maybe this is a good place to add `impulse_conversation_id` to [OfferOrderType](https://github.com/artsy/exchange/blob/c8b07db3918f00d85bc6a4830b54c5898e19ace4/app/graphql/types/offer_order_type.rb) to be able to test the mutation is returning it along the order object. Do you have thoughts on this @starsirius ?

sepans (approved):
This looks great. 🚀 

A couple of suggestions if you agree with it.

spec/controllers/api/requests/offers/create_inquiry_offer_order_with_artwork_mutation_request_spec.rb
ashleyjelks: I like this suggestion. I just looked at the docs and for me I don't think it would be a quick refactor. Since this work is blocking I wonder if it can be a follow up task rather than blocking this work from being merged?

starsirius:
Merging and let's iterate! Thanks, @ashleyjelks!

PR #676, opened 2021-02-03
Allow offerable from inquiry artworks to pass order validation

ashleyjelks:
This PR enables artworks with Gravity's `offerable_from_inquiry` property to be valid for orders.

cc @artsy/purchase-devs 

https://artsyproduct.atlassian.net/browse/PURCHASE-2412

### DIFF
--- a/lib/order_creator.rb
+++ b/lib/order_creator.rb
@@ -58,8 +58,9 @@ class OrderCreator
   end
 
   def valid_action?
+    offerable = artwork[:offerable] || artwork[:offerable_from_inquiry]
     action_error = if @mode == Order::BUY && !artwork[:acquireable] then :not_acquireable
-    elsif @mode == Order::OFFER && !artwork[:offerable] then :not_offerable
+    elsif @mode == Order::OFFER && !offerable then :not_offerable
     end
     @errors << action_error if action_error.present?
     action_error.nil?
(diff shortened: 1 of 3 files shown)


lib/order_creator.rb
```diff
@@ -58,8 +58,9 @@ def valid_edition_set?
   end
 
   def valid_action?
+    offerable_or_offerable_from_inquiry = !artwork[:offerable] || !artwork[:offerable_from_inquiry]
```
starsirius, line 61: The logic might be incorrect here. I think when one is `true` and the other is `false`, the artwork should be offerable, but we will return `:not_offerable`. We can add a test case in the tests below to verify.

Maybe we can simplify this as
```ruby
offerable = artwork[:offerable] || artwork[:offerable_from_inquiry]
# ...
elsif @mode == Order::OFFER && !offerable then :not_offerable
# ...
```

spec/lib/order_creator_spec.rb
```diff
@@ -113,6 +123,28 @@
...
+        before do
+          expect(order_creator.valid?).to eq true
+        end
+        it 'sets correct error' do
+          expect(order_creator.errors).to eq []
+        end
+      end
+      context 'offer order' do
+        let(:order_mode) { Order::OFFER }
+        let(:gravity_artwork) { gravity_v1_artwork(offerable_from_inquiry: false, offerable: true) }
```
starsirius, line 139: Perfect! Thank you adding more unit tests!

PR #677, opened 2021-02-04
Fix: persist impulse_conversation_id

xtina-starr:
Addresses [PURCHASE-2418](https://artsyproduct.atlassian.net/browse/PURCHASE-2418)

This PR expands the `OfferOrderType` to include `impulse_conversation_id`. The `impulse_conversation_id` wasn't getting persisted to the database so this also updates creation transaction inside the OrderCreator file.

[Corresponding metaphysics PR](https://github.com/artsy/metaphysics/pull/2977)

cc: @artsy/purchase-devs

### DIFF
--- a/lib/order_creator.rb
+++ b/lib/order_creator.rb
@@ -117,7 +117,8 @@ class OrderCreator
         state_expires_at: Order::STATE_EXPIRATIONS[Order::PENDING].from_now,
         original_user_agent: @user_agent,
         original_user_ip: @user_ip,
-        payment_method: Order::CREDIT_CARD # Default to credit card payment method
+        payment_method: Order::CREDIT_CARD, # Default to credit card payment method
+        impulse_conversation_id: @impulse_conversation_id
       )
 
       line_item = order.line_items.create!(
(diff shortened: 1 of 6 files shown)


lib/order_creator.rb
```diff
@@ -117,7 +117,8 @@ def create_order
         state_expires_at: Order::STATE_EXPIRATIONS[Order::PENDING].from_now,
         original_user_agent: @user_agent,
         original_user_ip: @user_ip,
-        payment_method: Order::CREDIT_CARD # Default to credit card payment method
+        payment_method: Order::CREDIT_CARD, # Default to credit card payment method
+        impulse_conversation_id: @impulse_conversation_id
```
xtina-starr, line 121: This was the missing piece to persist the `impulse_conversation_id` to the database. This gets nested pretty far down from where we initially call the create method in the service so it was easy oversight (and made for a fun scavenger hunt).

sepans: Nice catch 💯

sepans (approved):
Nice 🚀 

Great idea adding that mutation test which caught that persisting problem.

It looks ready to merge unless @starsirius has a comment.

starsirius (approved):

PR #683, opened 2021-03-04
feat: order events

sepans:
Part of [PURCHASE-2449](https://artsyproduct.atlassian.net/browse/PURCHASE-2449)

Adds `OrderHistory` type to `Order`

~Sample query~ 🛑  check https://github.com/artsy/exchange/pull/683#discussion_r590554505
```graphql
order(id: $id) {
  orderHistory {
    __typename
    ... on OfferSubmittedEvent {
      createdAt
      offer {
        amountCents
        fromParticipant
      }
    }
    ... on OrderStateChangedEvent {
      createdAt
        type
        stateReason
    }
  }
}
```

response:
```json
{
  "data": {
    "order": {
      "code": "575746786",
      "__typename": "OfferOrder",
      "orderHistory": [
        {
          "__typename": "OrderStateChangedEvent",
          "createdAt": "2021-02-26T15:45:34Z",
          "type": "PENDING",
          "stateReason": null
        },
        {
          "__typename": "OrderStateChangedEvent",
          "createdAt": "2021-02-26T21:34:46Z",
          "type": "SUBMITTED",
          "stateReason": null
        },
        {
          "__typename": "OfferSubmittedEvent",
          "createdAt": "2021-02-26T21:34:46Z",
          "offer": {
            "amountCents": 250000,
            "fromParticipant": "BUYER"
          }
        },
        {
          "__typename": "OrderStateChangedEvent",
          "createdAt": "2021-03-01T21:34:46Z",
          "type": "CANCELED",
          "stateReason": "seller_lapsed"
        }
      ]
    }
  }
}
```

Screen Shot 2021-03-08 at 5 25 39 PM

</details>

Co-authored-by: Erik Stockmeier <erik.d.stock@gmail.com>


app/graphql/types/order_event_union.rb
```diff
@@ -0,0 +1,28 @@
+module Types::EventInterface
+  include Types::BaseInterface
+  field :created_at, Types::DateTimeType, null: false
+end
+
+class Types::OfferEvent < Types::BaseObject
+  implements Types::EventInterface
+  field :offer, Types::OfferType, null: false
+end
```
starsirius, line 9: Given we are close to the source, I wonder if we have the opportunity to model events in a more generalized way and event-like. For example,

```graphql
order {
  events {
    ... on OfferEvent {
      type    # only SUBMITTED now but could include PENDING or other types in the future
      offer   # ideally a snapshot of the offer
      created_at
    }
  }
}
```

We are inferring immutable events from the mutable offers for now, assuming in the current flow offers are kind of immutable, but in the future we can generate real offer events, if needed.

app/graphql/types/order_event_union.rb
```diff
@@ -0,0 +1,28 @@
...
+
+class Types::StateEvent < Types::BaseObject
+  implements Types::EventInterface
+  field :state, Types::OrderStateEnum, null: false
+  field :reason, String, null: true
+end
```
starsirius, line 15: Similarly, I wonder if we can make the event less bound to a single attribute. Perhaps this can be an `OrderEvent` in parallel with the `OfferEvent`. For example,

```graphql
order {
  events {
    ... on OrderEvent {
      type          # SUBMITTED, APPROVED, CANCELED, etc.
      state_reason  # seller_lapsed, buyer_lapsed, etc.
      created_at
    }
  }
}
```

Curious how you think!

app/services/order_event_service.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def self.events_for(order_id:)
+    order = Order.find(order_id)
+    offer_events = offer_events(order)
+    state_events = state_events(order)
+    sorted_events(offer_events.concat(state_events))
+  end
+
+  def self.offer_events(order)
+    order.offers.submitted.map { |offer| OfferEvent.new(created_at: offer.created_at, offer: offer) }
```
starsirius, line 13: Offers are created as pending so we'll probably need the [submitted_at](https://github.com/artsy/exchange/blob/1cd7df5d875e6a04b5a22a3c6585df71af25b304/db/schema.rb#L130) timestamp instead.

sepans: 🙌🏽 we were wondering what that `submitted_at` is when we added this `order.offers.submitted.map`.
But got confused because [`submit_pending_offer `](https://github.com/artsy/exchange/blob/0f95f3931dd2f746452fa6f3214a65612e30c790/app/services/offer_service.rb#L31) sets `submitted_at` 🤔

erikdstock:
Aside from getting tests into place I see a few lingering questions:
- Initially I think we were thinking of this in terms of feeding the conversation view more than trying to backfill a list of events that could reconstruct the entire history of an offer.
  - For the first case the implementation is much simpler: `[Order pending, Order submitted, Buyer Offered x, Seller offered y, Order accepted by buyer, Order fulfilled]` etc. This avoids needing to think about associations and aggregates too much and or offer submission time- we assume they are immutable so it is no problem to have an `OfferEvent` (offer created) just map to its offer object, and for changes in state history we can just pull the relevant state change, so we have (pseudo-types)
```gql
# an offer is created on the order
interface OfferEvent {
  createdAt: Date # probably really the submitted_at property
  offer: Offer  # we assume this is immutable so we can always refer back to it as historical data
}

# order state changes
interface StateEvent {
  createdAt: Date
  state: Order::States
  reason: Order::Reasons
}
```
We could generalize this interface more for an event-sourced feel, but I'm not sure where the right balance point is for this. It might look something like: 
```gql
# Any event on an order
interface OrderEvent {
  createdAt: Date!
  type: EventType   # could just be __typename?
  # we don't really need these
  # id: ID
  # aggregateType: String
}

type OfferCreated implements OrderEvent {
  
  payload: Offer! # Or just a selection of offer values
}

type OrderStateChanged implements OrderEvent {
  payload: { state: Order::States!, reason: String }
}
type OrderSubmitted implements OrderEvent { payload: {reason: String }}
type OrderAccepted implements OrderEvent { payload: {reason: String }}
type OfferDeclined implements OrderEvent { payload: {reason: String }}
# ... etc
```
I'm not sure the latter case gets us much closer to the feature, and it feels more like we are backfilling a list of events that attempt to describe the history of the order.

I'm trying to think from the client's perspective here - we want to use this list of events to render a component, ideally without having to work around or against the generated types, so maybe it looks like this.
```ts
const query = gql` { me
order(id: $id) {
  events {
    type: __typename
    createdAt
    payload {
      ...on OfferCreated {
        offer { ... } // or the payload is the offer
      }
      ...on OrderStateChanged {
        state
        reason
      }
    }
  }
}}
`
const eventMessage = order.events.map((e) => {
  switch (e.type) {
  case "OfferCreated":
    return <OfferCreated at={e.createdAt} offer={e.payload} me={me} />
  case "OrderStateChanged":
    if (e.payload.state === 'accepted') {
      return <OfferAccepted at={e.createdAt} />
    } else if (e.payload.state === 'declined') {
      return <OfferDeclined at={e.createdAt} />
    }
  default:
    return null
  }
})
```

sepans:
👆🏽 graphql snippet and jsx code reads really well 🤩 

To be honest I don't have a clear big picture of what our future usages of these types are going to be (not even sure about the current ones 😅). I don't see a problem if we end up deprecating these types and create more generalized ones in the future, since they will surface in a very small and limited part of client code. Is that correct?

If so I am leaning toward keeping these types as simple as possible to make the iteration easier and faster.

PS: I came across this `OrderEvent` [class](https://github.com/artsy/exchange/blob/master/app/events/order_event.rb) (events being sent to pulse) which is more like the immutable 'snapshots' of order history we are envisioning for what an 'event' is. But those events are emitted when the change actually happens but here we are trying to retrospectively reconstruct those from the traces of data, which makes it more complicated.

app/graphql/types/order_interface.rb
```diff
@@ -23,6 +23,7 @@ module Types::OrderInterface
   field :last_transaction_failed, Boolean, null: true, method: :last_transaction_failed?
   field :line_items, Types::LineItemType.connection_type, null: true
   field :mode, Types::OrderModeEnum, null: true
+  field :order_history, [Types::OrderEventUnion], null: false
```
starsirius, line 26: Wondering if it makes sense to only allow `OrderStateChangedEvent` for BUY orders.

spec/controllers/api/requests/order_query_request_spec.rb
```diff
@@ -75,6 +75,21 @@
...
+                createdAt
+                offer {
+                  amountCents
+                  fromParticipant
+                }
+              }
+              ... on OrderStateChangedEvent {
+                createdAt
+                  type
+                  stateReason
```
starsirius, line 90: #minor. Extra whitespaces (human linting 😎 ).

spec/controllers/api/requests/order_query_request_spec.rb
```diff
@@ -75,6 +75,21 @@
             createdAt
             lastTransactionFailed
             displayCommissionRate
+            orderHistory {
```
starsirius, line 78: Is the `orderHistory` actually being asserted in this spec?

spec/services/order_history_service_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+        before do
+          Timecop.travel(1.minute)
+          order.approve!
+          Timecop.travel(1.minute)
+          order.fulfill!
+        end
+        it 'generates events' do
+          events = OrderHistoryService.events_for(order_id: order.id)
+          expect(events.length).to be 4
+          expect(events.map { |event| event.class.name.demodulize }).to eq %w[OrderStateChanged OrderStateChanged OrderStateChanged OrderStateChanged]
```
starsirius, line 40: #minor. Potentially this can use the [type matcher](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/type-matchers).

starsirius (approved):
Looks great to me! Have a few minor/non-blocking comments, and feel free to merge when ready! @erikdstock @sepans 💯

spec/services/order_history_service_spec.rb
sepans: Cool. Couldn't find out what is the best way of using type-matchers with an array of heterogeneous types. If it requires doing the comparison one by one like 👇🏽, I slightly prefer the current style that is more compact.
```ruby
expect(events[0]).to be_kind_of(OrderStateChanged)
...
```

app/graphql/types/order_interface.rb
sepans: Interesting. I am going to give it a try. From the client's perspective, this type-enforcement doesn't bring a lot of value (the client already knows how to handle the difference). But it makes the schema semantically valid 👍🏽

sepans: @starsirius @erikdstock I made the change in https://github.com/artsy/exchange/pull/683/commits/c9eeb5ba85a007b5ea241d526e22aef075ea6ac3

It makes the query a little bit verbose because it now needs an extra fragment over the `BuyOrder/OfferOrder` like this:
```graphql
query($id: ID) {
  order(id: $id) {
    ... on OfferOrder {
      orderHistory {
        __typename
        ... on OfferSubmittedEvent {
          createdAt
          offer {
            amountCents
            fromParticipant
          }
        }
        ... on OrderStateChangedEvent {
          createdAt
          type
          stateReason
        }
      }
    }
    ... on BuyOrder {
      orderHistory {
        __typename
        createdAt
        type
        stateReason
      }
    }
  }
}
```
but some of it can go inside a shared fragment so it is not that bad. I like that now the schema is semantically valid. What do you think?

(Also split  and moved the event types in a separate folder and added graphql test for buy orders)

erikdstock (commented):
forgot to submit this 😬

app/graphql/types/order_interface.rb
sepans: Yes, that's correct this change doesn't really help the client. It makes the schema tighter: Buy orders only have `OrderStateChanged` in their order history and no `OfferSubmitted`. The extra pieces in the query are the side effect of that change

PR #687, opened 2021-03-15
feat: do not require inventory for inquiry orders (PURCHASE-2471)

starsirius:
https://artsyproduct.atlassian.net/browse/PURCHASE-2471

(I did some refactoring and it might be easier to review by commits.)

Given that inquiry artworks don't have accurate inventory information (and we don't want partners to provide them upfront in Inquiry Checkout), we are going to relax the requirements on inventory for inquiry orders. That is,

- Do not check inventory
- Do not deduct/un-deduct inventory during the order lifecycle

I did an audit of how we currently handle inventory:

**Buy Now**
- Deduct inventory when a BN order is [submitted](https://github.com/artsy/exchange/blob/b271a36d5bd44d506ecd4bde5d1799fce79a1e99/app/services/order_service.rb#L54)
- Un-deduct inventory when [BN order submission failed](https://github.com/artsy/exchange/blob/b271a36d5bd44d506ecd4bde5d1799fce79a1e99/app/services/order_service.rb#L55) and inventory has been deducted

**Make Offer**
- Check inventory when [submitting an offer](https://github.com/artsy/exchange/blob/b271a36d5bd44d506ecd4bde5d1799fce79a1e99/app/services/offer_service.rb#L30) and [submitting an offer order](https://github.com/artsy/exchange/blob/b271a36d5bd44d506ecd4bde5d1799fce79a1e99/app/services/offer_service.rb#L41).
- Deduct inventory when [accepting an offer](https://github.com/artsy/exchange/blob/b271a36d5bd44d506ecd4bde5d1799fce79a1e99/app/services/offer_service.rb#L58)
- Un-deduct inventory when [accepting an offer failed](https://github.com/artsy/exchange/blob/b271a36d5bd44d506ecd4bde5d1799fce79a1e99/app/services/offer_service.rb#L75) and inventory has been deducted

Certain actions or events can also trigger inventory un-deduction, such as

- `OrderService.seller_lapse!`
- `OrderService.reject!`
- `OrderService.refund!`
- `StripeWebhookService.process_refund_event`

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -155,6 +155,14 @@ class Order < ApplicationRecord
     seller_type == AUCTION_SELLER_TYPE
   end
 
+  def inquiry_order?
+    impulse_conversation_id.present?
+  end
+
+  def require_inventory?
+    !inquiry_order?
+  end
+
   def to_s
     "Order #{id}"
   end
--- a/app/services/inventory_service.rb
+++ b/app/services/inventory_service.rb
@@ -1,0 +1,24 @@
+class InventoryService
+  def initialize(order)
+    @order = order
+    @deducted_items = []
+  end
+
+  def check_inventory!
+    raise Errors::InsufficientInventoryError unless @order.inventory?
+  end
+
+  def deduct_inventory!
+    @order.line_items.each do |item|
+      Gravity.deduct_inventory(item)
+      @deducted_items << item
+    end
+  end
+
+  def undeduct_inventory!
+    until @deducted_items.empty?
+      Gravity.undeduct_inventory(@deducted_items.first)
+      @deducted_items.shift
+    end
+  end
+end
--- a/app/controllers/errors/application_error.rb
+++ b/app/controllers/errors/application_error.rb
@@ -9,7 +9,7 @@ module Errors
       @code = code
       @data = data
       post_error_event if post_event
-      super(type)
+      super(code)
     end
 
     private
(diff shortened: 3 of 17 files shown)


app/services/inventory_service.rb
```diff
@@ -0,0 +1,24 @@
+class InventoryService
```
sweir27, line 1: 😍

app/models/order.rb
```diff
@@ -155,6 +155,14 @@ def auction_seller?
     seller_type == AUCTION_SELLER_TYPE
   end
 
+  def inquiry_order?
+    impulse_conversation_id.present?
+  end
+
+  def require_inventory?
```
sepans, line 162: #minor suggestion:

Have you considered moving this function into `InventoryService`?

It will encapsulate the logic there and makes these lines simpler:
instead of:
```ruby
@inventory_service.check_inventory! if @order.require_inventory?
```
having:
```ruby
@inventory_service.check_inventory!  # check_inventory! will verify require_inventory? 
```

but for these parts:
```ruby
order_cancelation_processor.queue_undeduct_inventory_jobs if order.mode == Order::BUY && order.require_inventory?
```
it makes the code a bit more verbose because an `InventoryService` needs to be initialized there (or inside `OrderCancelationProcessor`) to check `require_inveontory?` 

Now writing it down not sure if it is a good idea 😄 🤷🏽‍♂️

sepans (approved):
🤩 🚀 

Just a minor thought otherwise good to merge.

app/models/order.rb
starsirius: Yeah, I went back and forth in terms of where the logic should live. I ultimately decided to let the `Processor` handles the logic and the `Service` is more like a doer. We can definitely iterate on this tho! Thank you for raising that!

PR #689, opened 2021-03-22
feat: do not un-deduct inventory for inquiry orders when receiving Stripe refund webhook (PURCHASE-2471)

starsirius:
https://artsyproduct.atlassian.net/browse/PURCHASE-2471

Following https://github.com/artsy/exchange/pull/687, I missed one place where we un-deduct inventory when receiving a refund webhook from Stripe. This skips it for inquiry orders.

I also fixed the line item helper for determining if a line item has inventory when it's `nil` (currently it raises an error).


sepans (approved):
🚀

PR #690, opened 2021-03-22
feat: allow sales_observers to execute order queries

mdole:
Following up on https://github.com/artsy/volt/pull/4982. I originally missed that we were gatekeeping access to orders both in the Volt UI and in Exchange. This PR allows users with the `sales_observer` role to execute `order` and `orders` queries, but does _not_ allow them access to Exchange admin UI or to any mutations.

There are tests in place for the query permissions, and I confirmed the "no access to Exchange UI" via the following steps:
1. Started local Exchange server at localhost:3000
2. Opened Gravity staging console
3. `user = User.find_by_email 'matt.dole@artsymail.com'`
4. `user.update!(roles: ["admin", "customer_support", "billing_admin", "team", "sales_observer"])`
5. Opened incognito window, logged in and confirmed I was unable to access
6. `user.update!(roles: ["admin", "customer_support", "billing_admin", "team", "sales_admin", "partner_support"])`
7. Closed & reopened window for a fresh session, logged in, confirmed I was able to access (due to now having `sales_admin` and `partner_support` roles)

Original Jira ticket: https://artsyproduct.atlassian.net/browse/PX-3691

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -95,13 +95,17 @@ class Types::QueryType < Types::BaseObject
     context[:current_user][:roles].include?('trusted')
   end
 
+  def sales_observer?
+    context[:current_user][:roles].include?('sales_observer')
+  end
+
   def trusted_admin?
     user_roles = context[:current_user][:roles].split(',')
     ArtsyAuthToken.trusted_admin?(user_roles)
   end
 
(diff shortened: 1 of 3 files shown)


ansor4 (approved):
Sweet, thanks for sleuthing and fixing!

PR #691, opened 2021-03-25
WIP: expand orders to accept a new states argument

xtina-starr:
Addresses: [PURCHASE-2500](https://artsyproduct.atlassian.net/browse/PURCHASE-2500)

This is a work in progress for expanding the orders query to support filtering by multiple states.

**Filter by states argument:**
Screen Shot 2021-03-25 at 6 20 50 PM

**Error when both `state` and `states` arguments are passed:**
Screen Shot 2021-03-25 at 6 20 22 PM

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -15,6 +15,7 @@ class Types::QueryType < Types::BaseObject
     argument :seller_type, String, required: false
     argument :buyer_id, String, required: false
     argument :buyer_type, String, required: false
+    # state will soon be deprecated use states instead
     argument :state, Types::OrderStateEnum, required: false
     argument :states, [Types::OrderStateEnum], required: false
     argument :sort, Types::OrderConnectionSortEnum, required: false
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -16,6 +16,7 @@ class Types::QueryType < Types::BaseObject
     argument :buyer_id, String, required: false
     argument :buyer_type, String, required: false
     argument :state, Types::OrderStateEnum, required: false
+    argument :states, [Types::OrderStateEnum], required: false
     argument :sort, Types::OrderConnectionSortEnum, required: false
     argument :mode, Types::OrderModeEnum, required: false
   end
@@ -49,6 +50,13 @@ class Types::QueryType < Types::BaseObject
 
   def orders(params = {})
     validate_orders_request!(params)
+    states = params.delete(:states)
+    state = params[:state]
+
+    invalid_states_param = Errors::ValidationError.new(:invalid_states_params, message: 'params state and states cannot be passed together.')
+    raise invalid_states_param if state.present? && states.present?
+
+    params = params.merge(state: states) unless states.nil?
     sort = params.delete(:sort)
     order_clause = sort_to_order[sort] || { state_updated_at: :desc }
     Order.where(params).order(order_clause)
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -52,7 +52,7 @@ class Types::QueryType < Types::BaseObject
   def orders(params = {})
     validate_orders_request!(params)
     invalid_states_param = Errors::ValidationError.new(:invalid_states_params, message: 'params state and states cannot be passed together.')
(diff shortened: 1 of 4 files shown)


xtina-starr:
I'm leaning toward depreciating the `state` argument in a separate PR. I'm seeing it used in a few places:
- [Volt](https://github.com/artsy/volt/blob/master/app/operations/metaphysics/orders.graphql)
- [Metaphysics](https://github.com/artsy/metaphysics/blob/master/src/lib/stitching/exchange/v2/stitching.ts)
- [Force actually uses states](https://github.com/artsy/force/blob/master/src/v2/Apps/Purchase/Components/PurchaseHistory.tsx#L159)

app/graphql/types/query_type.rb
```diff
@@ -16,6 +16,7 @@ class Types::QueryType < Types::BaseObject
     argument :buyer_id, String, required: false
     argument :buyer_type, String, required: false
     argument :state, Types::OrderStateEnum, required: false
```
starsirius, line 18: Potentially we can add a [`deprecation_reason`](https://github.com/artsy/exchange/blob/0f95f3931dd2f746452fa6f3214a65612e30c790/app/graphql/types/line_item_type.rb#L7) to the `state` argument and make it clear in the schema.

app/graphql/types/query_type.rb
```diff
@@ -49,6 +50,13 @@ def order(args)
 
   def orders(params = {})
     validate_orders_request!(params)
+    states = params.delete(:states)
+    state = params[:state]
+
+    invalid_states_param = Errors::ValidationError.new(:invalid_states_params, message: 'params state and states cannot be passed together.')
+    raise invalid_states_param if state.present? && states.present?
```
starsirius, line 57: This may be passed incorrectly when passing `{ state: "submitted", states: [] }` because `[].present?` is `false`. It would be safer to check the presence of the _key_ instead of _value_. For example,
```ruby
raise invalid_states_param if params.has_key?(:state) && params.has_key?(:states)
```

xtina-starr: That means we might need to change from this `states = params.delete(:states)` to `states = params[:states]`?

starsirius: No necessarily. I think we can do this params validation early. For example,
```ruby
def orders(params = {})
  validate_orders_request!(params)
  raise Errors::ValidationError.new(...) if params.has_key?(:state) && params.has_key?(:states)

  states = params.delete(:states)
  params = params.merge(state: states) unless states.nil?
  # ...
end
```

app/graphql/types/query_type.rb
xtina-starr: I had this previously but it was causing the tests to fail with `Field 'orders' doesn't accept argument 'state'` which is one of the reason I was thinking do the actual deprecation in a follow up PR when we made adjustments to to places that consume this.

xtina-starr:
Also the test failure looks similar to the one I'm seeing in force when I've created the PR from my fork instead of upstream but this PR was created from an upstream branch. I'm not sure why I'm getting this build failure here.

app/graphql/types/query_type.rb
starsirius: That's interesting! I wouldn't expect adding the deprecation reason has real impact on the functionality. 🤔  Maybe we can try that again and look at the errors together?

xtina-starr: While looking into this we discovered that adding the `deprecation_reason` label actually removes the argument from the schema which is why I was seeing the `Field 'orders' doesn't accept argument 'state'` error. 

In order to be able to pass the deprecated reason and not see this error we would need also pass [isDeprecated: true](https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/introspection/type_type_spec.rb#L129). This requires that we update all the instances where we are using the state argument and is out of scope for this PR. I created a follow up [ticket](https://artsyproduct.atlassian.net/browse/PURCHASE-2539) that addresses the deprecation.

starsirius:
The failure was due to the mimemagic issue which should be fixed by https://github.com/artsy/exchange/pull/694. You can now rebase on top of latest master.

xtina-starr:
I rebase this branch with master and resolved some rubocop offenses. I'm going to go ahead and merge this PR.

PR #693, opened 2021-04-01
WIP spike: inquiry checkout p2: work without artwork location and shipping

sepans:
1st pass only collector side. the partner side is mocked

cc: @starsirius


app/models/offer.rb
```diff
@@ -18,9 +18,9 @@ def submitted?
   end
 
   def buyer_total_cents
-    return unless shipping_total_cents.present? && tax_total_cents.present?
-
-    amount_cents + shipping_total_cents + tax_total_cents
+    # does removing this break anything?
+    # return unless shipping_total_cents.present? && tax_total_cents.present?
+    amount_cents + shipping_total_cents.to_i + tax_total_cents.to_i
   end
```
sepans, line 24: I added this to avoid the error `TransactionFeeCalculator` throws but maybe it is a good change because we get the total amounts without tax and shipping.

not sure how consequential this is. 

Alternatively we can add change `TransactionFeeCalculator.calculate` function  and check
 ```ruby
return 0 if total_charge_amount.nil? || !total_charge_amount&.positive?
```
to guard against the error.

sepans: relax only for inquiry orders?

PR #699, opened 2021-04-20
WIP: spike p2 accept model as flag

sepans:
wip


app/models/offer.rb
```diff
@@ -18,9 +18,15 @@ def submitted?
...
   def buyer_total_cents
-    return unless shipping_total_cents.present? && tax_total_cents.present?
+    return if skip_buyer_total_calculation?
 
-    amount_cents + shipping_total_cents + tax_total_cents
+    amount_cents.to_i + shipping_total_cents.to_i + tax_total_cents.to_i
+  end
+
+  def skip_buyer_total_calculation?
+    return false if order.inquiry_order?
```
oxaudo, line 27: So that I understand - we will require additional approval from collector even if the inquiry artwork will have all the data initially, ie location + shipping info?

app/models/order.rb
```diff
@@ -227,6 +227,12 @@ def last_transaction_failed?
     last_transaction.failed? || last_transaction.requires_action?
   end
 
+  def initially_with_missing_artwork_metadata
+    first_artwork = line_items.first.artwork
+    # TODO: check for domestic/international simillar to ShippingHelper
+    inquiry_order? && (first_artwork[:location].blank? || first_artwork[:domestic_shipping_fee_cents].blank? || first_artwork[:international_shipping_fee_cents].blank?)
```
oxaudo, line 233: For this condition as we've discussed on Slack I believe that we could have domestic shipping but not international shipping and for inquiry works we can even have a sate where international is present but domestic is not (not sure what it's supposed to mean). So I believe checking `domestic_shipping_fee_cents || international_shipping_fee_cents` would give us negative results in more cases than we want.

db/schema.rb
```diff
@@ -187,6 +187,7 @@
     t.boolean "assisted"
     t.string "fulfilled_by_admin_id"
     t.string "impulse_conversation_id"
+    t.boolean "needs_both_party_approval", default: false
```
oxaudo, line 190: `two_factor_approval` ?  -:)

lib/offer_totals.rb
```diff
@@ -49,4 +49,8 @@ def tax_data
     # If there are no taxable addresses then we set the sales tax to 0.
     OpenStruct.new(tax_total_cents: 0, should_remit_sales_tax: false)
   end
+
+  def skip_calculation?
+    !@order.shipping_info? || (artwork[:location].blank? && @order.inquiry_order?)
```
oxaudo, line 54: Wondering if we implementing something like that - we should expand this `2factorapproval` possibility to regular BNMO works. I think those will never be possible now because of validations client side but don't see why we could not assume that we'll handle BNMO orders that end up missing metadata the same way in exchange. What you think?

oxaudo (commented):
I'm sure missing tons of context here but still leaving some of my initial reactions if they might be useful or at least a point of discussions.

spec/integration/inquiry_checkout_p2_happy_path.rb
```diff
@@ -0,0 +1,353 @@
...
+    
+    expect(events[2].offer.from_participant).to eq 'buyer'
+    expect(events[2].offer.amount_cents).to be 500_00
+    expect(events[2].offer.tax_total_cents).to be 25_00
+    expect(events[2].offer.shipping_total_cents).to be 30_00
+
+    expect(events[3].state).to eq 'approved'
+
+    # FIXME: cc card should not be captured at this point!!!
+    expect(order.transactions.length).to be(1)
```
sepans, line 316: we cannot use `OfferService.accept_offer(offer, seller_user_id)` method (line 284)

PR #700, opened 2021-04-20
Spike p2 accept model as counter

sepans:
wip


spec/integration/inquiry_checkout_p2_happy_path.rb
```diff
@@ -0,0 +1,339 @@
...
+      seller_total_cents: 483_05,
+      shipping_country: 'US',
+      credit_card_id: 'credit_card_1'
+    )
+
+    # checking events
+    # TODO: include in query
+    events = OrderHistoryService.events_for(order_id: order.id)
+    # byebug
+    # FIXME: Fix. it should not generate event for seller counter
```
sepans, line 296: - we should not get new events there for new counter offer
- the event sent to pulse needs to have needed flag(s) to be able to send a different email instead of partner counter

sepans: we also need to mark the offer without metadata and the seller counter (approve) after that with a flag, either a persisted flag or just when passing the events for the UI to create correct CTA

PR #702, opened 2021-04-20
feat: return 0 for stripe fees on staging

ashleyjelks:
Update Exchange to not calculate Stripe fees on staging.

This PR introduces a new environment variable:  `NEW_TRANSACTION_FEE_ENABLED` that enables us to test not returning Stripe transaction fees in Staging only.

cc @artsy/purchase-devs 

https://artsyproduct.atlassian.net/browse/PURCHASE-2567

### DIFF
--- a/lib/transaction_fee_calculator.rb
+++ b/lib/transaction_fee_calculator.rb
@@ -1,5 +1,7 @@
 module TransactionFeeCalculator
   def self.calculate(total_charge_amount, currency)
+    return 0 if ENV['NEW_TRANSACTION_FEE_ENABLED'] == 'true'
+
     0 unless total_charge_amount&.positive?
 
     # This is based on Stripe fee, we decided to charge unified 3.9% + 30 cents across all countries
--- a/spec/lib/transaction_fee_calculator.spec.rb
+++ b/spec/lib/transaction_fee_calculator.spec.rb
@@ -1,0 +1,23 @@
+require 'rails_helper'
+
+describe TransactionFeeCalculator do
+  let(:currency) { 'USD' }
+  let(:total_charge_amount) { 15000 }
+
+  context 'with transaction fee flag set as true' do
+    before do
+      allow(ENV).to receive(:[]).and_call_original
+      allow(ENV).to receive(:[]).with('NEW_TRANSACTION_FEE_ENABLED').and_return(true)
+    end
+    it 'calculates transaction fee' do
+      transaction_fee = TransactionFeeCalculator.calculate(total_charge_amount, currency)
+      expect(transaction_fee).to eq 0
+    end
+  end
+  context 'with transaction fee flag set as false' do
+    it 'calculates transaction fee' do
+      transaction_fee = TransactionFeeCalculator.calculate(total_charge_amount, currency)
+      expect(transaction_fee).to eq 615
+    end
+  end
+end


spec/lib/transaction_fee_calculator.spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      allow(ENV).to receive(:[]).and_call_original
+      allow(ENV).to receive(:[]).with('NEW_TRANSACTION_FEE_ENABLED').and_return(true)
+    end
+    it 'returns 0 for the transaction fee' do
+      transaction_fee = TransactionFeeCalculator.calculate(total_charge_amount, currency)
+      expect(transaction_fee).to eq 0
+    end
+  end
+  context 'outside staging environment' do
+    it 'returns 0 for the transaction fee' do
```
sepans, line 18: it 'calculates transaction fee...'

spec/lib/transaction_fee_calculator.spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      transaction_fee = TransactionFeeCalculator.calculate(total_charge_amount, currency)
```
sepans, line 19: I _think_ this handles the case where `NEW_TRANSACTION_FEE_ENABLED` is not set at all (and env returns `nil`). right?

sepans (approved):
Looks great 🚀 A couple of spec message updates and a question to confirm.

spec/lib/transaction_fee_calculator.spec.rb
ashleyjelks: Oof, good question. Not sure, I'll write a spec to see what happens when we explicitly set it as `nil`

spec/lib/transaction_fee_calculator.spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+describe TransactionFeeCalculator do
+  let(:currency) { 'USD' }
+  let(:total_charge_amount) { 15000 }
+
+  context 'inside staging environment' do
```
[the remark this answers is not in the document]
xtina-starr: I actually think we could rename this context description to something like "New transaction fee enabled" because this should actually be environment agnostic. We just care if the flag is set to true or not.

xtina-starr:
One thing that I think is worth mentioning is that we where going to make the assumption that if the transaction fee is zero we wouldn't want to display the Artsy fee as a line item in CMS, but it look like the fee can also be zero when the `total_charge_amount` is not a positive value. That seems like an edge case (I'm not sure) but might we want to re-evaluate our assumption on the CMS side and/or is okay to not display the Artsy fee as a line item if the `total_charge_amount` happens to be negative?

I bring this up here because maybe we might want to return `nil` here instead if this potential edge case warrant different treatment.

cc: @ashleyjelks @sepans @starsirius

lib/transaction_fee_calculator.rb
```diff
@@ -1,5 +1,7 @@
 module TransactionFeeCalculator
   def self.calculate(total_charge_amount, currency)
+    return 0 if ENV['NEW_TRANSACTION_FEE_ENABLED'] == true
```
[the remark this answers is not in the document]
starsirius: We could, but given it's a temporary flag we'll remove after release, I'm fine using it directly.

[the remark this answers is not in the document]
starsirius: I think environment variables are loaded as strings, so we might want to do this instead:
```ruby
return 0 if ENV['NEW_TRANSACTION_FEE_ENABLED'] == 'true'
```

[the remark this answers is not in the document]
sepans: good catch 👆🏽

lib/transaction_fee_calculator.rb
```diff
@@ -1,5 +1,7 @@
...
+
     0 unless total_charge_amount&.positive?
```
starsirius, line 5: @xtina-starr Good call out on the non-positive case! Looking at the data I don't think this is happening today where we charge buyer <= 0 dollars, so hopefully we can safely ignore and in CMS we can still rely on 0 to display the line item or not.

spec/lib/transaction_fee_calculator.spec.rb
ashleyjelks: When it's `nil` it has the default behavior/its the same as the flag being `false` which is what we want!

lib/transaction_fee_calculator.rb
oxaudo: Sorry late to this party. One note here. Given the requirement to not show this row in clients and the fact that sometimes for the old orders it might be waved (0 legitimately) maybe we need to return `nil` vs `0` and rely on that distinction?

Or we can always hide it if it's 0 on the clients? @starsirius, @ashleyjelks, @sepans?

oxaudo: @xtina-starr - wrote this and then noticed your comment above as well. Jinx! :)

PR #703, opened 2021-04-21
Update README.md

oxaudo:
And I think it's valid to say that we 'ARE' handling various types of transactions now:)

cc @ashkan18 - your baby have grown! And gained some weight:)


sepans:
Thanks, merging

PR #706, opened 2021-04-22
Allow inquiry orders without artwork location/shipping info to be submitted

sepans:
[PURCHASE-2562]

This PR allows inquiry orders with missing artwork location and shipping info to be submitted.

First, it skips tax and shipping calculation in `OfferTotals` for such orders. Otherwise `ShippingHelper` [here](https://github.com/artsy/exchange/blob/ebef2888ebafd6f44a9795d5ffeb61d5e744f04f/lib/shipping_helper.rb#L9) would throw an error.

Second, it forces `buyer_total_cents` calculation in `Offer` for inquiry orders so we still get buyer total without tax and shipping for the purpose of displaying the estimate in the front end. Those `to_i`s avoid error when value is `nil`

Also added the integration spec for such offers with missing metadata up to the submission work. Following PRs will add the rest of the steps.


spec/lib/offer_totals_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  let(:fulfillment_type) { Order::PICKUP }
+  let(:shipping_info) { {} }
+  let(:impulse_conversation_id) { nil }
+  let(:gravity_artwork) { gravity_v1_artwork(_id: 'a-1', price_listed: 1000.00, edition_sets: [], domestic_shipping_fee_cents: 200_00, international_shipping_fee_cents: 300_00) }
+  let(:order) { Fabricate(:order, seller_id: 'partner-1', seller_type: 'gallery', buyer_id: 'buyer1', **shipping_info, buyer_type: Order::USER, fulfillment_type: fulfillment_type, impulse_conversation_id: impulse_conversation_id) }
+  let(:offer) { Fabricate(:offer, order: order, from_id: 'partner-1', from_type: 'gallery', amount_cents: 800_00, shipping_total_cents: 0, tax_total_cents: 300_00) }
+  let(:line_item) { Fabricate(:line_item, order: order, artwork_id: 'a-1') }
+
+  before do
+    line_item.order
```
sepans, line 14: without this line `order.line_items` is empty when `OfferTotals` calls `order.line_items.&first`. But `line_item.order` has the correct value. As if the reverse of `has_many` relation is kind of lazy loaded and this line forces that to be filled in?! 🤔

app/models/offer.rb
```diff
@@ -18,9 +18,16 @@ def submitted?
...
-    return unless shipping_total_cents.present? && tax_total_cents.present?
+    return if skip_buyer_total_calculation?
 
-    amount_cents + shipping_total_cents + tax_total_cents
+    amount_cents.to_i + shipping_total_cents.to_i + tax_total_cents.to_i
+  end
+
+  def skip_buyer_total_calculation?
+    # always calculate it for for inquiry orders because it should work with missing tax/shipping
+    return false if order.inquiry_order?
```
oxaudo, line 28: I'm really wondering if we can be generic here and always calculate `buyer_total_cents`. That would mean that regular offers/orders just be marked `untransactionable`. We would prevent creating those from UI anyways. But if they do get created - I'm wondering if there is a downside of following this inquiry offer logic in that case.

oxaudo: And if we go with this generic approach - we need to make sure to check `order.transactionable` before moving any order to the final `accepted` state and not allow that for any orders.

lib/offer_totals.rb
```diff
@@ -49,4 +49,12 @@ def tax_data
     # If there are no taxable addresses then we set the sales tax to 0.
     OpenStruct.new(tax_total_cents: 0, should_remit_sales_tax: false)
   end
+
+  def skip_calculation?
+    inquiry_offer_without_artwork_location? || !@order.shipping_info?
+  end
+
+  def inquiry_offer_without_artwork_location?
+    @order.inquiry_order? && artwork[:location].blank?
```
oxaudo, line 58: Here as well - I think we `could` skip this for any offers without `artwork[:location]`. But maybe I'm missing something?

oxaudo (commented):
I have a conceptual question about how we approaching this very differently for inquiry offers vs regular offers. I'm wondering if there is any downside to treating all offers the same here? @sepans @starsirius - sorry if you've already went over it multiple times and that is an informed decesion.

app/models/offer.rb
sepans: Yes, I agree with generalizing this. The only reason I added it was to be conservative and make sure we are not changing anything about regular BNMO. Although unlikely, there might be a place in exchange or force where the presence of those values is being used to determine something 🤷🏽

app/models/offer.rb
```diff
@@ -18,7 +18,7 @@ def submitted?
   end
 
   def buyer_total_cents
-    return unless shipping_total_cents.present? && tax_total_cents.present?
+    return unless [amount_cents, shipping_total_cents, tax_total_cents].all?(&:present?)
```
starsirius, line 21: For now we'll only calculate `buyer_total_cents` when we have _all_ amounts ready, similar to [`seller_total_cents`](https://github.com/artsy/exchange/blob/e9a20118b87d507e1002e303a6bb4ac1f598db8c/lib/offer_order_totals.rb#L35). We are checking with design and see if we need to display the temporary buyer total amount before it's finalized.

lib/line_item_totals.rb
```diff
@@ -16,15 +16,15 @@ def shipping_total_cents
     return unless @order.shipping_info? && @line_item.artwork_location.present?
 
     @shipping_total_cents ||= begin
-      per_item_shipping_cents = ShippingHelper.calculate(@line_item.artwork, @order.fulfillment_type, @order.shipping_address)
+      per_item_shipping_cents = ShippingCalculator.new(@line_item.artwork, @order).calculate
```
starsirius, line 19: Refactored and used a `ShippingCalculator` instance, instead of class methods.

lib/offer_totals.rb
```diff
@@ -11,23 +11,21 @@ def initialize(order, offer_amount = nil)
...
   end
 
   def artwork_location
-    @artwork_location ||= Address.new(artwork[:location])
+    @artwork_location ||= Address.new(artwork[:location]) if artwork[:location]
   end
 
   def tax_data
-    return OpenStruct.new(tax_total_cents: nil, should_remit_sales_tax: nil) unless @order.shipping_info?
+    return OpenStruct.new(tax_total_cents: nil, should_remit_sales_tax: nil) unless @order.shipping_info? && artwork_location
```
starsirius, line 28: Same here for offers.

lib/shipping_calculator.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def initialize(artwork, order)
+    @artwork = artwork
+    @order = order
+  end
+
+  def calculate
+    return 0 if pickup? || consignment?
+
+    raise Errors::ValidationError.new(:missing_artwork_location, artwork_id: artwork[:_id]) if artwork[:location].blank? && !order.inquiry_order?
```
starsirius, line 12: Kept the errors for BNMO to maintain the current behavior that buyers get an earlier error message in the checkout flow (e.g. the set shipping step). We can follow up separately and move it to other layer (e.g. api) if we want, and keep this calculator more generic.

lib/shipping_calculator.rb
```diff
@@ -0,0 +1,59 @@
+class ShippingCalculator
```
starsirius, line 1: Pretty much the same as the existing `ShippingHelper`. I refactored it so we can initialize instances.

spec/integration/inquiry_checkout_missing_metadata_happy_path_spec.rb
```diff
@@ -0,0 +1,224 @@
...
+      buyer_total_cents: nil,
+      seller_total_cents: nil
+    )
+    offer = Offer.last
+    expect(offer).to have_attributes(
+      amount_cents: 500_00,
+      order_id: order.id,
+      shipping_total_cents: nil,
+      tax_total_cents: nil,
+      buyer_total_cents: nil
```
starsirius, line 137: Buyer total cents would stay `nil` until we have all necessary amounts.

spec/integration/inquiry_checkout_missing_metadata_happy_path_spec.rb
```diff
@@ -0,0 +1,224 @@
...
+      state: Order::SUBMITTED,
+      fulfillment_type: Order::SHIP,
+      shipping_country: 'US',
+      credit_card_id: 'credit_card_1',
+      # below are calculated without shipping/tax
+      shipping_total_cents: nil,
+      tax_total_cents: nil,
+      items_total_cents: 50000,
+      buyer_total_cents: nil,
+      seller_total_cents: nil,
```
starsirius, line 211: Seller total cents also stays `nil` until we have all necessary amounts.

starsirius:
@sepans Thanks for pairing on Friday and I pushed the changes and did some refactoring. See if this makes sense now! 🙏

PR #707, opened 2021-04-26
Persist shipment quote information

jpotts244:
PR to start persisting some shipment option data in exchange database. 
This work is only developer facing to unblock discovery and future modeling work. 

Included in this PR
- First pass potential modeling for shipment quotes
- API authentication and connection setup
- Potential pattern to make requests to third party in the future
- Base API error handling
- Base API timeout handling
- DB fields were driven by our mocks and designs for the user experience

Thinking this will be an iterative experience and some of this work will be subject to change. Let us know your thoughts!


app/models/quote.rb
```diff
@@ -0,0 +1,3 @@
+class Quote < ApplicationRecord
```
jpotts244, line 1: Wondering if it would be helpful to make this model more specific. So far the fields are pretty tied to the data that comes back from our third party.

```ruby
class ExternalQuote
```
?

app/services/shipping_service.rb
```diff
@@ -0,0 +1,39 @@
...
+          price_currency: quote[:total_currency],
+        )
+      end
+
+      quote_collection.save!
+    end
+
+    private
+
+    def parse_name(quote)
```
jpotts244, line 34: Parsing will need to be more sophisticated and probably handled in its own class - up to us if we should carry that out now or iteratively

db/migrate/20210420175055_create_quotes.rb
```diff
@@ -0,0 +1,14 @@
+class CreateQuotes < ActiveRecord::Migration[6.1]
+  def change
+    create_table :quotes, id: :uuid do |t|
+      t.references :quote_collection, foreign_key: true, type: :uuid
+      t.string :tier, null: false
+      t.string :name
+      t.integer :external_id, null: false
+      t.string :price, null: false
```
jpotts244, line 8: Right now price comes back as a string and we are just storing it as such.

Something we may want to design now is is this OK?
We will need the price to calculate tax/UI costs so we will need to perform some math on this value. (to note: our tax calculator expects price to be in cents)

SO maybe before we merge this, we update this col to be `Bigint` and in price_cents..?

lib/arta/quote.rb
```diff
@@ -0,0 +1,97 @@
...
+      def physical_artwork_info(artwork)
+        {
+          subtype: format_artwork_type(artwork[:category], artwork[:framed]),
+          unit_of_measurement: "cm",
+          width: artwork[:width_cm],
+          height: artwork[:height_cm],
+          depth: artwork[:depth_cm],
+          value: artwork[:price_listed],
+          value_currency: artwork[:price_currency],
+        }.compact
```
jpotts244, line 32: `.compact` here, we are getting errors + requests failures if we pass in keys will null values here.

So we will need to handle nil checks on our side when building the request hash, `compact` will do that for us

app/models/quote.rb
jpotts244: Peril left us a great suggestion: https://github.com/artsy/exchange/blob/master/docs/papertrail_audit_logging.md

A TODO for the new model classes

